PR #42, opened 2021-03-03
inventory: accept host with missing field

goneri:
When a column is not available for a given host, we use a default empty string.


murakobardenas (approved):

murakobardenas (approved):

cidrblock:
this was fixed in a different PR

Issue #62, opened 2021-03-18
[RFE] Increment logs and .json


gamuniz:
logging files and artifacts currently overwrite

PR #78, opened 2021-03-22
fixed it so logger doesn't overwrite with each action and added times…

gamuniz:
issue link https://github.com/ansible/ansible-navigator/issues/62


cidrblock:
PR #280 added log-append functionality, and by default the artifacts have an iso 8601 timestamp covered in
PR #269 

thx for this one because it was a great idea!

closing because of the others

PR #99, opened 2021-03-26
Enabling Zuul CI on ansible-navigator

GomathiselviS:
Signed-off-by: GomathiselviS <gomathiselvi@gmail.com>

Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/832

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,4 +1,0 @@
-black
-mypy
-pytest
-pytest-cov
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,0 +1,4 @@
+black
+flake8
+pytest-xdist
+yamllint
--- a/tox.ini
+++ b/tox.ini
@@ -1,17 +1,35 @@
 [tox]
-envlist = black,type,py36,py37,py38,py39
+minversion = 1.4.2
+envlist = linters
+skipsdist = True
 skip_missing_interpreters = true
 
 [testenv]
-deps =
-    -r{toxinidir}/requirements-test.txt
-    -c{toxinidir}/tests/constraints.txt
+deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/test-requirements.txt
 commands =
     py.test -v
 
 [testenv:black]
+install_command = pip install {opts} {packages}
 commands =
-    black --check .
+  black -v {toxinidir}
+
+[testenv:linters]
+install_command = pip install {opts} {packages}
+commands =
+  black -v --diff --check {toxinidir}
+  flake8 {posargs}
+  yamllint -s .
+
+[flake8]
+# E123, E125, E203 skipped as they are invalid PEP-8.
+
+show-source = True
+ignore = E123,E125,E203,W503
+max-line-length = 160
+builtins = _
+exclude = .git,.tox,*__init__.py
 
 [testenv:type]
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -26,10 +26,10 @@ commands =
 # E123, E125, E203 skipped as they are invalid PEP-8.
 
 show-source = True
-ignore = E123,E125,E203,W503
+ignore = E123,E125,E203,W503,F401
 max-line-length = 160
 builtins = _
-exclude = .git,.tox,*__init__.py
+exclude = .git,.tox
 
 [testenv:type]
 commands =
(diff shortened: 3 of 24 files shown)


relrod (changes requested):
Until/unless the current GHA stuff is all ported over, merging this will amount to less CI than we currently have. (Which is probably why this is marked "DNM: " but I didn't realize what "DNM" meant until just now :laughing:)

test-requirements.txt
```diff
@@ -0,0 +1,7 @@
+black==19.3b0 ; python_version > '3.5'
```
pabelanger, line 1: there should be no need to cap this

test-requirements.txt
```diff
@@ -0,0 +1,7 @@
...
+coverage==4.5.4
```
pabelanger, line 2: can be removed for now likely

test-requirements.txt
```diff
@@ -0,0 +1,7 @@
...
+flake8
+mock ; python_version < '3.5'
+pexpect
```
pabelanger, line 5: not sure if needed

tox.ini
```diff
@@ -1,17 +1,35 @@
 [tox]
-envlist = black,type,py36,py37,py38,py39
+minversion = 1.4.2
+envlist = black,type,py36,py37,py38,py39,linters
```
pabelanger, line 3: likely remove black here, as linters is the entry point now

requirements-test.txt
```diff
@@ -1,4 +0,0 @@
-black
```
pabelanger, line 1: These versions should be moved into test-requirements.txt

tox.ini
```diff
@@ -1,17 +1,35 @@
...
+  yamllint -s .
+
+[flake8]
+# E123, E125 skipped as they are invalid PEP-8.
+
+show-source = True
+ignore = E123,E125
+max-line-length = 160
+builtins = _
+exclude = .git,.tox,tests/unit/compat/
```
pabelanger, line 32: can remove tests/unit/compat

cidrblock (approved):

test-requirements.txt
pabelanger: we can remove

test-requirements.txt
```diff
@@ -0,0 +1,6 @@
+black
+flake8
+mock ; python_version < '3.5'
```
pabelanger, line 3: I don't think we need mock

zuul.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- project:
+    templates:
+      - ansible-navigator-jobs
```
pabelanger, line 4: now that this is passing, lets remove this file. and add project-config / ansible-zuul-jobs changes.

test-requirements.txt
GomathiselviS: Removed.

requirements-test.txt
GomathiselviS: removed the file

tox.ini
```diff
@@ -1,17 +1,35 @@
...
+  yamllint -s .
+
+[flake8]
+# E123, E125, E203 skipped as they are invalid PEP-8.
+
+show-source = True
+ignore = E123,E125,E203,W503
+max-line-length = 160
+builtins = _
+exclude = .git,.tox,*__init__.py
```
pabelanger, line 32: what is the reason for ignoring *__init__.py

tox.ini
```diff
@@ -1,17 +1,35 @@
...
+  black -v --diff --check {toxinidir}
+  flake8 {posargs}
+  yamllint -s .
+
+[flake8]
+# E123, E125, E203 skipped as they are invalid PEP-8.
+
+show-source = True
+ignore = E123,E125,E203,W503
+max-line-length = 160
```
pabelanger, line 30: this likely wants to be 100?

tox.ini
GomathiselviS: > what is the reason for ignoring ***init**.py

Flake8 throws  F401 error, because the imported module is not used in the same file. This happens for most of the __init__ files. These files have only import statements. So I decided to exclude them from flake8 check.

pabelanger: I would ignore f401 for now, rather then skipping the whole file. Then we can work to bring it into flake8 in follow up patches

PR #112, opened 2021-03-31
Fix ansible-playbook not found issue in local env

ganeshrn:
*  When running within local env get the complete executable
   path for `ansible-playbook`
*  Update docs
*  Fix linter issues


ganeshrn:
recheck

relrod:
recheck

PR #119, opened 2021-04-02
yaml config work

relrod:
- Nuke old ini stuff
- Move `Sentinel` to utils
- Add `NavigatorConfig` class with a method to pull out config values and with an optional default and ability to fall back to a global default
- Integrate a list of config parameters with argparse
- Add some tests
- Update `:open` logic to pull editor from config
- Fix an unrelated regex deprecation warning where flags must be at the start of the string
- Move argparse defaults out to config defaults and make argparse defaults be `Sentinel` so we know if the param was passed or not.


relrod:
recheck

cidrblock:
+1

relrod:
recheck

Issue #120, opened 2021-04-06
run command does not prompt form if inventory is not provided


ganeshrn:
In case multiple invalid inventory sources are provided to the `run` command in that case a form is prompted asking for the correct inventory path. However if no inventory source is provided inventory option input is not present in the form.

example command
Mulitple invalid inventory source (working scenario)
```
ansible-navigator run gather.yaml -i test1 -i test2
```

No inventory source. Ensure inventory is not set in the config file (failing scenario)
```
ansible-navigator run gather.yaml 
```

PR #121, opened 2021-04-06
Fix single inventory source form prompt

ganeshrn:
Fixes https://github.com/ansible/ansible-navigator/issues/120

If inventory source is not provided as for input within the user
form


cidrblock:
LGTM

PR #123, opened 2021-04-06
Add support for inventory subcommand stdout mode

ganeshrn:
*  Add support to run `ansible-inventory` command in
   pass-through mode to work within local and execution
   environment.
*  Refactor inventory action plugin to use new runner API's
   for interactive and stdout mode


cidrblock:
def _show_columns(self) -> List:
 
this can just return args inventory_columns now since rick is going to make it a list in the config

cidrblock:
+1

PR #125, opened 2021-04-06
Refactor testcase for config action

ganeshrn:
*  Add testcase to test config action in interactive
   and stdout mode
*  Move integration test into it's own folder
*  Add fixtures folder


cidrblock:
recheck

tools/test-setup.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -xe
```
pabelanger, line 1: This isn't needed any more we've landed changes to zuul, to always use the latest version of podman.

ganeshrn: Tracked with issue https://github.com/ansible/ansible-navigator/issues/126

ganeshrn:
Slower integration test run issue tracked via https://github.com/ansible/ansible-navigator/issues/127

PR #129, opened 2021-04-07
config updates

cidrblock:
Set default value for inventory to a list, since the action is append
Put sample yaml in README
Remove split in inventory action since inventory is a list
Add argparse entries for editor, remove config calls in :open since editor is a first class arg now
Set no_osc4 to default to Sentinel
Log the source of "where" the parameter came from (user, default, arparse)
Start utils test
Move cli test into test_cli and add inventory and run tests
other opinionated style changes that were not necessary :)


ansible_navigator/cli_args.py
```diff
@@ -141,6 +142,23 @@ def _doc_params(parser: ArgumentParser) -> None:
...
+            "--ecmd",
+            "--editor-command",
+            help="Specify the editor command, filename and line number",
+            default=Sentinel,
+            dest="editor_command",
+        )
+        parser.add_argument(
+            "--econ",
+            "--editor-console",
+            help="Specify if the  editor is console based",
```
relrod, line 157:
```suggestion
            help="Specify if the editor is console based",
```

cidrblock: thx

ansible_navigator/config.py
```diff
@@ -13,39 +17,55 @@
...
     "ansible-navigator": {
         "container-engine": "podman",
         "doc-plugin-type": "module",
         "editor": {
-            "command": "vi +{line_number} {filename}",
+            "command": generate_editor_command(),
             "console": True,
         },
-        "execution-environment": False,
+        "execution-environment": True,
```
ganeshrn, line 39: Default EE value should be `False` right?

ansible_navigator/config.py
```diff
@@ -74,11 +94,11 @@ def get(self, keys: List[str], default: Any = Sentinel) -> Any:
             else:
                 break
         else:
-            return current
+            return "user provided config file", current
```
ganeshrn, line 97: Instead of string maybe we can use enums? For eg:
```
class ConfigSource():
    NONE = 0
    # user provided config file
    USER_CONFIG_FILE = 1
    # argparse default config
    ARGPARSE_DEFAULT = 2
    # default config
    DEFAULT_CONFIG = 3
```

ganeshrn (commented):
Applying gate for now, will discuss the review comments offline and fix as required in a separate PR

PR #131, opened 2021-04-08
Remove dep on pytest-mock, use builtin monkeypatch

cidrblock:
Not sure why I didn't think of this yesterday, old habits.


cidrblock:
pushing this through, I don't believe anyone will be concerned

Issue #134, opened 2021-04-09
Stacktrace if inventory-columns does not exist in return value from action 


ganeshrn:
```
    return self._menu(dicts, cols, indicies)
  File "/home/ansible-navigator/ansible_navigator/ui_framework/menu_builder.py", line 52, in _menu
    convert_percentage(dicts[idx], cols, self._pbar_width)
  File "/home/ansible-navigator/ansible_navigator/ui_framework/utils.py", line 23, in convert_percentage
    value = dyct[key]
KeyError: 'ansible_network_cli_ssh_type'
```

Config
```
---
ansible-navigator:
  inventory-columns:
    - ansible_network_os
    - ansible_network_cli_ssh_type
    - ansible_connection
  log:
    level: debug
  no-osc4: True
```

PR #135, opened 2021-04-09
Fix issue with menu builder

ganeshrn:
Fixes https://github.com/ansible/ansible-navigator/issues/134

Try to convert value to percentage only if column value exist
in return result from action


cidrblock:
LGTM

PR #136, opened 2021-04-12
Add configuration file path setting using env variable

ganeshrn:
*  Add support to set ansible-naviagator configuration
   file path using environment variable `ANSIBLE_NAVIGATOR_CONFIG`
*  Add support to read configuration file from `.yaml` and `.json` extension
* The first found matched directory path (based on order mentioned above) can have only one
  valid configuration file. If in case more than one configuration files (with different
  supported extensions) are found it will result in an error to avoid conflict

### DIFF
--- a/ansible_navigator/cli.py
+++ b/ansible_navigator/cli.py
@@ -1,5 +1,6 @@
 """ start here
 """
+import json
 import logging
 import os
 import sys
@@ -12,7 +13,6 @@ from curses import wrapper
 from functools import partial
 from typing import Callable
 from typing import List
-from typing import NoReturn
 from typing import Optional
 from typing import Tuple
 from yaml.scanner import ScannerError
@@ -23,9 +23,10 @@ from .config import NavigatorConfig
 from .action_runner import ActionRunner
 
 from .utils import check_for_ansible
+from .utils import error_and_exit_early
 from .utils import flatten_list
 from .utils import get_and_check_collection_doc_cache
-from .utils import get_conf_dir
+from .utils import get_conf_path
 from .utils import set_ansible_envar
 from .utils import Sentinel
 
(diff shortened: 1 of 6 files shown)


ansible_navigator/cli.py
```diff
@@ -181,17 +181,29 @@ def setup_config() -> Tuple[List[str], NavigatorConfig]:
...
+                break
+
+        if config_dir is not None:
+            # Since we give get_conf_dir our config path, it's guaranteed to exist
+            # if config_dir is not None.
+            config_path = os.path.join(config_dir, found_filename)
+            pre_logger_msgs.append("Found config file at {0}".format(config_path))
+            found_config = True
+        else:
+            pre_logger_msgs.append("Could not find config directory, using all defaults.")
```
samccann, line 206:
```suggestion
            pre_logger_msgs.append("Could not find config directory. Using all default values for configuration.")
```

docs/configuration.rst
```diff
@@ -14,13 +14,19 @@ Several options in ``ansible-navigator`` can be configured by making use of a
...
 - ``/etc/ansible-navigator/ansible-navigator.yml``
 - ``[prefix]/etc/ansible-navigator/ansible-navigator.yml`` (e.g., ``/usr/local/etc/...``)
 
-Here is an example conifguration file which can be copied into one of those paths.
+.. note::
+   The configuration file extension can be either ``.yml`` or ``.yaml``. If files with both
+   extension are present in that case ``.yml`` is given preference over ``.yaml``.
+
+
+Here is an example configuration file which can be copied into one of those paths.
```
samccann, line 29:
```suggestion
You can copy the example configuration file below into one of those paths to start your ``ansible-navigator`` config file.
```

samccann (commented):
some docs suggestions

cidrblock:
this looks great +1

PR #137, opened 2021-04-12
Add test for inventory subcommnd

ganeshrn:
Depends-On: https://github.com/ansible/ansible-navigator/pull/136

*  Update integration test common code to emulate user
   UI integration and fetch the terminal output
   for end-to-end testing
*  Add integration tests for inventory subcommnad
*  Add unit test for inventory subcommand

### DIFF
--- a/tests/defaults.py
+++ b/tests/defaults.py
@@ -1,1 +1,5 @@
+import os
+
 default_container_image = "quay.io/ansible/ansible-runner:devel"
+pexpect_use_poll = True
+pexpect_timeout = 600
--- a/tests/integration/_common.py
+++ b/tests/integration/_common.py
@@ -3,6 +3,7 @@
 import os
 import re
 import sys
+import pexpect
 import tempfile
 
 from argparse import Namespace
@@ -12,12 +13,16 @@ from typing import List
 from typing import Tuple
 from typing import Optional
 
+from distutils.spawn import find_executable
+
 from ansible_navigator.app_public import AppPublic
 from ansible_navigator.ui_framework.ui import Action as Ui_action
 from ansible_navigator.ui_framework.ui import Interaction
 from ansible_navigator.ui_framework.ui import Ui
 
 from ansible_navigator.steps import Steps
+
+from .. import defaults
 
 
 class ActionRunTest:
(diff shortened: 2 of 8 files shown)


ganeshrn:
recheck

tests/integration/_common.py
```diff
@@ -131,3 +132,12 @@ def run_action_stdout(self, cmdline: List) -> Tuple[str, str]:
         sys.stderr = __stderr__
 
         return stdout, stderr
+
+
+def get_executable_path(name):
```
cidrblock, line 137: I had better luck using sys
executable here so the same interpreter and venv running the tests would be used

tox.ini
```diff
@@ -29,7 +29,7 @@ commands = coverage erase
 install_command = pip install {opts} {packages}
 commands =
   pylint {[base]pkg_name} --ignore=tm_tokenize
-  black -v --diff --check {toxinidir}
+  black -v {toxinidir}
```
cidrblock: Don't we want check here to find errors?

ganeshrn: I remove `--diff` and `--check` in my local env so that black can autoformat. By mistake committed those changes. Reveted in latest commit.

ganeshrn:
recheck

tests/defaults.py
```diff
@@ -1 +1,5 @@
 default_container_image = "quay.io/ansible/ansible-runner:devel"
+
+# TODO: Need to figure the optimal sleep time, if sleep time is less
+# output on the terminal is not caputred correctly
+tumx_read_delay_after_user_interaction = 3
```
cidrblock, line 5: I think this can go away, missed removal of it

tests/integration/actions/test_inventory.py
```diff
@@ -0,0 +1,86 @@
...
+fixtures_dir = os.path.join(os.path.dirname(__file__), "..", "..", "fixtures")
+
+CLI = (
+    f"{get_executable_path('python')}"
+    " -m ansible_navigator inventory"
+    f" -i {os.path.join(fixtures_dir,'inventory')}"
+)
+
+testdata = [
+    (0, CLI, "ansible-navigator inventory command top window"),
```
cidrblock, line 19: Just a note, this approach is dependant on the use of : commands sine we look for :help after a command to determine the screen has been updated

ganeshrn: +1

PR #140, opened 2021-04-14
Add set_environment_variable and pass_environment_variable

### LINKED ISSUE #127: Integration tests running in stdout mode with EE are slower than expected
Integration test running in stdout mode with EE is slow (eg: )
Adding `--tty` option to container execution command https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/config/_base.py#L384  results in subprocess execution to wait till default timeout (300 sec) is triggered. 

Note the `--tty` option is added in stdout mode so that the user password prompt is not echoed in the `ansible-navigator ` terminal

### LINKED ISSUE #138: ansible-navigator.yml should allow ENV vars to be set or passed through to the docker/podman command
Similar to what tox allows:

```
setenv = 
    TERM = xterm
passenv = HOME
```
I think our keys would be `set_envvars` and `pass_envvars` the first a dict and the later a list

### LINKED ISSUE #139: Command line params do not match settings file
Specifically ee_image vs. execution-environment-image

cidrblock:
Beginning work to add set & pass ENV VAR support to the EE
Fixes #139 
Fixes #138 
Fixes #127 (adding PAGER: cat to test)

Tests and docs to follow in next PR


ganeshrn (approved):
Looks good

PR #146, opened 2021-04-16
Add tests that check from CLI out to runner

cidrblock:
These tests check from cli.main() out to the respective runner calls for:

- inventory
- run
- config

in both stdout and interactive mode

the following parameters are covered:
- execution-environment
- execution-environment-image
- set-environment-variable
- pass-environment-variable

and should cover their use at the CLI and in a configuration file

Some code changes necessary:
- artifact file fix in run
- turn execution-environment into a true/false parameter
- pass the original command line back through argparse in run to cover all params
- reorg the post processing in CLI, this could be better still

This also includes the doc changes for set and pass env var since these were left out of the original PR

**Note**:  The 4 integration tests share a common base class, this pattern can be repeated for doc, collection etc

**Note**: Test fixtures have been moved into directories to indicate and contain their scope to the tests using them


cidrblock:
recheck

PR #147, opened 2021-04-20
Add config interactive tests using tmux

cidrblock:
Add interactive tests for config using tmux
Remove old config interactive tests
move other stdout config tests into action/config

### DIFF
--- a/tests/integration/_common.py
+++ b/tests/integration/_common.py
@@ -204,9 +204,6 @@ class TmuxSession:
         if self._server.has_session(self._session_name):
             self._session.kill_session()
 
-    def send_only(self, value):
-        self._pane.send_keys(value, suppress_history=False)
-
     def interaction(self, value):
         """interact with the tmux session"""
         self._pane.send_keys(value, suppress_history=False)
--- a/tests/integration/_common.py
+++ b/tests/integration/_common.py
@@ -162,11 +162,13 @@ class TmuxSession:
         config_path=None,
         cwd=None,
         session_name="ansible-navigator-integration-test",
+        setup_commands=None,
     ) -> None:
         self._window_name = window_name
         self._config_path = config_path
         self._session_name = session_name
         self._cwd = cwd
+        self._setup_commands = setup_commands or []
 
         if self._config_path is None:
             self._config_path = os.path.join(
@@ -192,11 +194,18 @@ class TmuxSession:
         self._pane.set_width(132)
         # do this here so it goes away with the tmux shell session
         self._pane.send_keys(f"export ANSIBLE_NAVIGATOR_CONFIG={self._config_path}")
+
+        # send any other commands needed for setup
+        for command in self._setup_commands:
+            self._pane.send_keys(command)
         return self
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         if self._server.has_session(self._session_name):
             self._session.kill_session()
+
+    def send_only(self, value):
+        self._pane.send_keys(value, suppress_history=False)
 
     def interaction(self, value):
         """interact with the tmux session"""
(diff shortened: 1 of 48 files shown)


tests/integration/actions/test_inventory.py
```diff
@@ -7,7 +7,7 @@
 from ... import defaults
 from .._common import get_executable_path
 from .._common import fixture_path_from_request
-from .._common import update_fixtures
+from .._common import update_fixtures  # pylint: disable=unused-import
```
relrod, line 10: Is this change necessary with the `UPDATE_FIXTURES` changes? Can we just leave it uncommented and conditionalized on the value of the constant?

tox.ini
```diff
@@ -12,9 +12,11 @@ skip_missing_interpreters = true
 [testenv]
 allowlist_externals = cat
                       rm
+                      grep
 deps = -r{toxinidir}/requirements.txt
        -r{toxinidir}/test-requirements.txt
 commands =
+    /bin/bash -c 'grep -FInrq "UPDATE_FIXTURES = True" ./tests && exit 1 || exit 0'
```
relrod, line 19: This will miss a lot of cases, but it might not be a huge deal.

tests/integration/_common.py
```diff
@@ -233,6 +239,21 @@ def fixture_path_from_request(request, index):
     return dir_path, file_name
 
 
+def container_runtime_or_fail():
+    """find a container runtime, prefer podman
+    fail if neither available"""
+    # pylint: disable=import-outside-toplevel
+    import subprocess
+
+    for runtime in ("docker", "podman"):
```
relrod, line 248: The doc comment says "prefer podman", but this tries docker first and returns if it finds it.

relrod: Also, should this list come from somewhere else, so there's only one place to update if we add support for another container runtime in the future? We hardcode a similar list in `cli_args`.

tests/integration/actions/config/base.py
```diff
@@ -0,0 +1,47 @@
...
+        }
+        with TmuxSession(**params) as tmux_session:
+            yield tmux_session
+
+    def test(self, request, tmux_config_session, index, user_input, comment):
+        # pylint:disable=unused-argument
+        # pylint: disable=too-few-public-methods
+        """test interactive config from cli
+
+        uncomment the update_fixtures line to update the fixtures
```
relrod, line 31: This line can probably be removed, since `UPDATE_FIXTURES` is a thing now.

tests/integration/actions/config/base.py
```diff
@@ -0,0 +1,47 @@
...
+            mask = "X" * 50
+            if "13│BECOME_PLUGIN_PATH" in line:
+                received_output[idx] = mask
+            if "15│CACHE_PLUGIN_CONNECTION" in line:
+                received_output[idx] = mask
+
+        if self.UPDATE_FIXTURES:
+            update_fixtures(request, index, received_output, comment)
+        dir_path, file_name = fixture_path_from_request(request, index)
+        with open(f"{dir_path}/{file_name}") as infile:
```
relrod, line 45: Prefer `os.path.join()`

tests/integration/actions/config/test_direct_interactive_noee.py
```diff
@@ -0,0 +1,30 @@
+""" config direct from cli interactive w/o ee
+"""
+import pytest
+
+from .base import BaseClass
+
+from ..._common import get_executable_path
+
+CLI = (
+    f"{get_executable_path('python')}" " -m ansible_navigator config --execution-environment false"
```
relrod, line 10:
```suggestion
    f"{get_executable_path('python')} -m ansible_navigator config --execution-environment false"
```

tests/integration/actions/config/test_direct_interactive_ee.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+import pytest
+
+from .base import BaseClass
+
+from ..._common import get_executable_path
+from ..._common import container_runtime_or_fail
+
+CLI = (
+    f"{get_executable_path('python')}"
```
relrod, line 11: Similar to below - I'd either merge this with the f-string below, or drop the f-string syntax from this line and just leave the function call to return a string.

PR #149, opened 2021-04-20
Inventory tests following config format

cidrblock:
- Had to set the tmux window width to 200 for the long :i commands, hence the fixtures changed
- Move inventory tests into inventory folder
- Fixed an issue with :inventory params not being split before reparsed
- Clean up as suggested by rick for f strings
- clean up container engine finding and fixture, tests now require a container engine present on the test system
- Found an interesting issue #150


tests/integration/_common.py
```diff
@@ -191,7 +187,7 @@ def __enter__(self):
         self._pane.split_window(vertical=False, attach=False)
         # attached to upper left
         self._pane.set_height(20)
-        self._pane.set_width(132)
+        self._pane.set_width(200)
```
cidrblock, line 190: this is what triggered all the fixture rebuilds

tests/integration/actions/inventory/test_welcome_interactive_noee.py
```diff
@@ -0,0 +1,49 @@
...
+from ..._common import get_executable_path
+from ..._common import container_runtime_or_fail
+
+CLI = get_executable_path("python") + " -m ansible_navigator"
+
+testdata = [
+    (0, CLI, "ansible-navigator inventory command top window"),
+    (
+        1,
+        f":i -i {ANSIBLE_INVENTORY_FIXTURE_DIR} --ee false",
```
cidrblock, line 17: this is what is kind of gross and captured in issue #150

ganeshrn (approved):

PR #152, opened 2021-04-21
Add support for doc command stdout mode

ganeshrn:
*  Add support for doc command to run in stdout mode
   with EE enabled/disabled
*  Refactor doc command in interactvie mode to use runner
   API's to fetch the plugin docs


ganeshrn:
recheck

PR #153, opened 2021-04-22
[:run] work with same-directory ansible.cfg

relrod:
Change:
- Set cwd runner arg in run action

Test Plan:
- Local testing
- New integration test using run in stdout mode

Tickets:
- Subsumes #151

### DIFF
--- a/ansible_navigator/runner/api.py
+++ b/ansible_navigator/runner/api.py
@@ -173,7 +173,6 @@ class CommandBaseRunner(BaseRunner):
         """generate arguments required to be passed to ansible-runner"""
         if self._playbook:
             self._cmdline.append(self._playbook)
-            self._runner_args.update({"cwd": os.path.dirname(os.path.abspath(self._playbook))})
 
         for inv in self._inventory:
             self._cmdline.extend(["-i", inv])
--- a/tests/fixtures/integration/actions/run/ping.yml
+++ b/tests/fixtures/integration/actions/run/ping.yml
@@ -1,0 +1,5 @@
+- hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: ensure we can run a ping task
+      ping:
--- a/tests/integration/actions/run/test_stdout.py
+++ b/tests/integration/actions/run/test_stdout.py
@@ -1,0 +1,53 @@
+""" run integration tests
+"""
+import os
+
+from ..._common import ActionRunTest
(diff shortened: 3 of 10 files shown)


tests/integration/actions/run/test_stdout.py
```diff
@@ -0,0 +1,30 @@
+""" run integration tests
```
ganeshrn, line 1: Thanks for adding the initial testcases for `run` subcommand. Since we can have tescases for EE and without EE the file can be renamed to `test_stdout_noee.py` in sync with testcases for other commands.

ganeshrn: Also will have to use the libtmux send keys to test for password prompts with local and container (EE) runs

tests/fixtures/integration/actions/run/ping.yml
```diff
@@ -0,0 +1,5 @@
+- hosts: localhost
+  gather_facts: false
+  tasks:
+    - name: ensure we can run a ping task
+      ping:
```
ganeshrn, line 5: Can we add `vars_prompt` scenario. It is useful especially in EE case when the playbook is executed within a container but any stdin prompt from ansible-core has to be passed on to the local terminal for user input.

relrod: Yeah, I'm planning to add more tests over the next few days, but wanted to keep this PR self contained. I'll open new PRs with more tests around `run` and also some of the other existing functionality. :)

ganeshrn (approved):

ansible_navigator/runner/api.py
```diff
@@ -173,7 +173,6 @@ def generate_run_command_args(self) -> None:
         """generate arguments required to be passed to ansible-runner"""
         if self._playbook:
             self._cmdline.append(self._playbook)
-            self._runner_args.update({"cwd": os.path.dirname(os.path.abspath(self._playbook))})
```
relrod, line 176: Unsure if this is safe to do or not. Tests pass, and it fixes the issue in the title of this PR, but I don't know what other effects it has. /cc @cidrblock

cidrblock:
may be replaced by #383

cidrblock:
closing as #383 did this, I didn't look deep enough to see if the tests were of value to harvest but we can still do that

Issue #150, opened 2021-04-20
When using :inventory or :run, EE parameters need to be passed on the : line or they revert to default


cidrblock:
for instance

`ansible-navigator --ee false --eei image --ce docker`

If you `:inventory -i some_inventory.yaml`

The original command line params are lost

You would have to `:inventory -i some_inventory.yaml -ee false --eei image --ce docker`

which is goofy

(the inventory integration tests have made made to work, but should be cleaned up when this is fixed)

cidrblock:
PR #167 took care of this

PR #154, opened 2021-04-22
Create a config file and multilayer value setup

jborean93:
This is a very early draft of moving the config setup away from argparse argument results to a centralised config setup. This supports layered config options and I'm hoping a way to update each option rather than blow away all existing options when the argparser is invoked again interactively.

Needs a lot more work to tidy up and hook into the code but so far the `--help` options for argparse still works.

Fixes https://github.com/ansible/ansible-navigator/issues/150


jborean93:
Closing in favour of https://github.com/ansible/ansible-navigator/pull/155.

PR #155, opened 2021-04-23
Configuration subsystem refactor

cidrblock:
ansible-navigator is not yet using this, that will happen in another PR


ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+
+from ..utils import oxfordcomma
+
+
+def abs_user_path(fpath):
+    """Resolve a path"""
+    return os.path.abspath(os.path.expanduser(fpath))
+
+
+def generate_editor_command():
```
ganeshrn, line 20: Should this function have return type defined?

ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+            cli_parameters=CliParameters(short="--econ"),
+            description="Specify if the editor is console based",
+            settings_file_path_override="editor.console",
+            value=EntryValue(default=True),
+        ),
+        Entry(
+            name="execution_environment",
+            choices=[True, False],
+            cli_parameters=CliParameters(short="--ee"),
+            description="Enable the use of an execution environment",
```
ganeshrn, line 94: We can have a top section for EE parameters `execution_environment` as there can be multiple options related to EE?
`execution_environment`, `execution_environment_image`, `container_engine`, `container_mount_path`

ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+            name="cmdline",
+            description="Placeholder for argparse remainder",
+            value=EntryValue(default=[]),
+        ),
+        Entry(
+            name="container_engine",
+            choices=["podman", "docker"],
+            cli_parameters=CliParameters(short="--ce"),
+            description="Specify the container engine",
+            value=EntryValue(default="podman"),
```
ganeshrn, line 73: The default value and the description strings can be read from a separate config file `(anisble-navigator.yaml or ansible-navigator.json)` which can be stored under `ansible_navigator/configuration` file.

ganeshrn: It can help kee the docs and config option within code in sync

ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+        ),
+        Entry(
+            name="editor_command",
+            cli_parameters=CliParameters(short="--ecmd"),
+            description="Specify the editor comamnd",
+            settings_file_path_override="editor.command",
+            value=EntryValue(default=generate_editor_command()),
```
ganeshrn, line 80: To avoid callable in default this option can be changed to `editor_name` and define a fallback value to read from environment variable

ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+        Entry(
+            name="execution_environment_image",
+            cli_parameters=CliParameters(short="--eei"),
+            description="The name of the execution environment image",
+            value=EntryValue(default="image_here"),
+        ),
+        Entry(
+            name="inventory",
+            cli_parameters=CliParameters(action="append", nargs="+", short="-i"),
+            description="Specify an inventory path",
```
ganeshrn, line 106: In addition to inventory path the `ansible-inventory -i` option can take  comma-separated host list as input

ansible_navigator/configuration/application_configuration.py
```diff
@@ -0,0 +1,199 @@
...
+from .definitions import Config
+from .definitions import CliParameters
+from .definitions import SubCommand
+from .definitions import Entry
+from .definitions import EntryValue
+
+from ..utils import oxfordcomma
+
+
+def abs_user_path(fpath):
```
ganeshrn, line 15: This function can be moved to utils.py file

ansible_navigator/configuration/application_post_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        return messages, errors
+
+    return wrapper
+
+
+class ApplicationPostProcessor:
+    """application post processor"""
+
+    @staticmethod
+    def _abs_user_path(fpath):
```
ganeshrn, line 34: Any specific reason function is defined here again?

ansible_navigator/configuration/application_post_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        return os.path.abspath(os.path.expanduser(fpath))
+
+    @staticmethod
+    def _str2bool(value: Any) -> bool:
```
ganeshrn, line 38: Function can be moved to utils file for common handling of boolean value across the App's

ansible_navigator/configuration/application_post_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            if value.lower() in ("yes", "true"):
+                return True
+            if value.lower() in ("no", "false"):
+                return False
+        raise ValueError
+
+    def _flatten_list(self, lyst: Any) -> List:
```
ganeshrn, line 51: Also can be moved to utils file

ansible_navigator/configuration/application_post_processor.py
```diff
@@ -0,0 +1,181 @@
...
+        messages: List[Message] = []
+        errors: List[str] = []
+        if config.app == "inventory" and not entry.value.current:
+            msg = "An inventory is required when using the inventory subcommand"
+            errors.append(msg)
+            return messages, errors
+        if entry.value.current is Sentinel:
+            entry.value.current = []
+        else:
+            entry.value.current = self._flatten_resolve_list_of_paths(entry.value.current)
```
ganeshrn, line 105: This might will if hostname is given as input, since the path does not exist

ansible_navigator/configuration/configuration.py
```diff
@@ -0,0 +1,143 @@
...
+    """the configuration class"""
+
+    def __init__(
+        self,
+        params,
+        application_configuration,
+        apply_previous_cli=False,
+        save_as_intitial=False,
+        settings_file_path=None,
+    ):
```
ganeshrn, line 29: Doc string for parameters will be helpful here :-)

ansible_navigator/configuration/application_configuration.py
ganeshrn, line 15: Not required if function moved to utils file

ansible_navigator/configuration/application_configuration.py
cidrblock: I think I prefer it all in one place, this file/object can be loaded by python and the entries iterated to generate a doc

cidrblock: I think it might also leave the door open for problems in the future, right now an "Entry" cannot be defined without a description since it's required by the underlying obj definition, if it gets pulled if could be forgotten.....  So we would have to author some tests to check for that.....  the key's would have to stay in sync as well since each entry would need a corresponding entry in the external file, I'm not thinking it's worth it here.

ansible_navigator/configuration/application_configuration.py
cidrblock: by having this here, it will show up when a user does a `--help` if it gets moved to post processing, the user won't see that their EDITOR is the default.

ansible_navigator/configuration/application_configuration.py
cidrblock: I'm assuming you mean in the settings file, so we can add execution-environment as a parent key and nest the related params

cidrblock: `  execution-environment:
    container-engine: podman
    enabled: False
    environment-variables:
      pass:
        - ONE
        - TWO
        - THREE
      set:
        KEY1: VALUE1
        KEY2: VALUE2
        KEY3: VALUE3
    image: test_image`

cidrblock: set and pass environment variables will be moved under a the same key since they are specific to the ee

ansible_navigator/configuration/application_configuration.py
cidrblock: will add to description

ansible_navigator/configuration/application_post_processor.py
cidrblock: nope, just a miss because of the way I coded it, will draw it from utils now

ansible_navigator/configuration/application_post_processor.py
cidrblock: will look for a comma and not resolve as path

ansible_navigator/configuration/configuration.py
cidrblock: of course

PR #156, opened 2021-04-26
Add testcases for doc subcommand [part 1]

ganeshrn:
*  Add integration test and fixtures for doc
   command


cidrblock (changes requested):
Should we add a welcome_interactive_ee?  so it aligns with the tests for inventory and config?

ganeshrn:
> Should we add a welcome_interactive_ee? so it aligns with the tests for inventory and config?

I am planning to add the testcase in another PR as there is an know issue with parsing the doc option in TUI mode

PR #167, opened 2021-04-30
Configuration subsystem integration

cidrblock:
Integrate configuration subsystems into ansible-navigator:

Major changes:
- changes in the settings file structure, nesting ee params under the execution-environment ket
- actions that use class App now have an entry and exit function to normalize their ui handling and arg restoration
- many changes to adapt to NOT_SET as the value for parameters that have not been set at all
- remove all cli handling from action as the config subsystem takes care of it
- action log now uses class App as it base to take advantage of enter, exit functionality
- cli_args and config are removed as they are replaced by configuration subsystem
- Moved all application initialization code into initialize.py
- Normalize the signature of all prelogger code to return messages, errors, value
- (where messages is a LogMessage, that can specify a log level)
- introduce ability to write a default location log file with level debug if we sys.exit from the beginning
- changes to tests and test fixtures as required


ansible_navigator/utils.py
```diff
@@ -403,3 +347,70 @@ def str2bool(value: Any) -> bool:
...
+    :param template_vars: The vars used to render the template
+    :type template_vars: dict
+    """
+    # hide the jinja that may be in the template_vars
+    template_vars = escape_moustaches(template_vars)
+
+    if HAS_ANSIBLE:
+        logger.info("Ansible installed, Ansible's plugins will be available")
+
+        ansible_templar = Templar(loader=None)
```
ganeshrn, line 367: Should we support Ansible plugins? This requires direct import from `ansible/ansible`

cidrblock: No, this can be switched to just jinja for templating, I think jordan yaml PR should give us good looking doc examples......  this was really just added for the to_yaml filter plugin

PR #171, opened 2021-05-03
add pytest duration, move cli options to toml

### LINKED ISSUE #90: Timing of unit tests
Some units tests should be timed, although we might not be able to fail based on a threshold value, having the data would be good.

Example-

tokenize large yaml file
catalog collections
others?

cidrblock:
fixes #90 

it's a little noise but I would rather have this info when we need it

### DIFF
--- a/artifact.json
+++ b/artifact.json
@@ -1,0 +1,7 @@
+{
+    "version": "1.0.0",
+    "plays": [],
+    "stdout": [],
+    "status": "failed",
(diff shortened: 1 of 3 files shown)


artifact.json
```diff
@@ -0,0 +1,7 @@
+{
```
relrod, line 1: I'm guessing you didn't mean to add this file?

PR #185, opened 2021-05-04
:stream changed to :stdout

### LINKED ISSUE #168: :stream should be :stdout to match the mode option

cidrblock:
It matches the `--mode` flag now

added some basic integration tests & a new tmux kwarg for waiting on a playbook status for when playbooks are run and the tests should only continue when the playbook is completed

Fixes #168


tests/integration/_common.py
```diff
@@ -213,16 +213,20 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
+        while not all(ok_to_return):
+            ok_to_return = []
             showing = []
             while not showing:
                 time.sleep(0.1)
                 showing = self._pane.capture_pane()
-            help_not_on_screen = not any(":help help" in line for line in showing)
+            if wait_on_help:
+                ok_to_return.append(any(":help help" in line for line in showing))
+            if wait_on_playbook_status is not None:
```
relrod, line 228: Why not `if wait_on_playbook:` and make it a bool above? Otherwise, with this strict check, if someone calls with `wait_on_playbook_status=False`, we'll still wait on it right?

cidrblock: my thinking here is that you may want to wait on a specific status, for instance, you may write a playbook that is known to fail, and then you may want to wait on FAILED

cidrblock: ok, I get it :) thx, will fix

relrod: Ah I see. I read it too quickly. Alright, I'm fine with either default then. It might still be worth making the check non strict just to save someone time in the future, but not a big deal.

PR #193, opened 2021-05-05
Post processing messages

### LINKED ISSUE #186: Improve post processing error formatting to include the command that was run
```210504190755.564 ERROR 'ansible_navigator.actions.run_9fb3._update_args' Configuration errors encountered, rolling back to previous configuration.
210504190755.564 ERROR 'ansible_navigator.actions.run_9fb3._update_args' A playbook is required when using the run subcommand
```

Could be better as 
```
A playbook is required when using the run subcommand, the command provided was ':run'
```

So it is clear it was the use of :run from inside navigator that caused the error

cidrblock:
Fixes #186 

- This changes configuration subsystem errors to WARNING, since a form will be shown and the user prompted
- When a configuration subsystem error happens, prepend the command to the errors so it is in the log file and provided to the user if a sys.exit happens
- make some tests a little less narrow to allow for the additional message


ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -83,15 +83,16 @@ def configure(self) -> Tuple[List[LogMessage], List[str]]:
         restore them
         """
         self._config.original_command = self._params
-        message = f"Params provided: {self._config.original_command}"
-        self._messages.append(LogMessage(level=logging.DEBUG, message=message))
+        cmd_message = f"Command provided: '{' '.join(self._config.original_command)}'"
```
relrod, line 86: Is this enough? What will this do if the original command has an argument with spaces in it? I think that information will be lost. (i.e. args `foo --bar "baz buzz" biz` will get printed as `foo --bar baz buzz biz`)... Unless I'm wrong. :)

cidrblock: just testing this out, sys.argv handles the quotes when it generates the list
```
➜  /tmp python go.py foo --bar "baz buzz" biz
['go.py', 'foo', '--bar', 'baz buzz', 'biz']
```

but in the actions we just split the `:` command:

```
>>> foo = 'foo --bar "baz buzz" biz'
>>> foo.split()
['foo', '--bar', '"baz', 'buzz"', 'biz']
```
https://github.com/ansible/ansible-navigator/blob/c4957f0228303a6baeb07945635229c6ac18baed/ansible_navigator/actions/run.py#L308

but we _should_ be using shlex

```
>>> import shlex
>>> shlex.split(foo)
['foo', '--bar', 'baz buzz', 'biz']
```

https://github.com/ansible/ansible-navigator/issues/195

I'll change this to shlex.join in anticipation of that

cidrblock: nice catch, btw, could have been a nasty little bug later on

PR #197, opened 2021-05-05
Shlex in actions

### LINKED ISSUE #195: Use shlex to split params instead of just split
https://github.com/ansible/ansible-navigator/blob/c4957f0228303a6baeb07945635229c6ac18baed/ansible_navigator/actions/run.py#L308

```
Python 3.8.2 (default, Dec 21 2020, 15:06:04)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> foo = 'foo --bar "baz buzz" biz'
>>> foo.split()
['foo', '--bar', '"baz', 'buzz"', 'biz']
>>> import shlex
>>> shlex.split(foo)
['foo', '--bar', 'baz buzz', 'biz']
>>>
```

so we get the same output

cidrblock:
Fixes #195 

- Switch to shlex.split in actions for `:` splitting and for `cldline` entry from forms

### DIFF
--- a/ansible_navigator/actions/collections.py
+++ b/ansible_navigator/actions/collections.py
@@ -2,6 +2,7 @@
 import curses
 import json
 import os
+import shlex
 import subprocess
 
 from json.decoder import JSONDecodeError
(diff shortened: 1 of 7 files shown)


relrod (approved):
:shipit:

Issue #200, opened 2021-05-06
Rename `:load` to `:replay`, add a `--mode stdout` capability


cidrblock:
The hack in actions/load.py can also be removed which prevents exposing the :run run_stdout method

cidrblock:
@NilashishC mentioned he might add run_stdout to replay

PR #203, opened 2021-05-06
Remove modes from configuration subsystem Subcommand definition

### LINKED ISSUE #198: Replace `modes` within the Subcommand definition with `internals`
this will remove ansible-navigator specific code from the configuration subsystem

cidrblock:
We can dynamically introspect the action to see what modes it supports instead.

Fixes #198 

Creates #200 , which will clean up the load action hackery

add integration tests for :load, since it was modified
add unit test to configuration subsystem for action introspection failure cases


ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -173,10 +174,48 @@ def mode(entry: Entry, config: ApplicationConfiguration) -> PostProcessorReturn:
...
+            errors.append(error)
+            return messages, errors
+
+        subcommand_modes = []
+
+        try:
+            getattr(subcommand_action, "run_stdout")
+            subcommand_modes.append("stdout")
+        except AttributeError:
+            pass
```
relrod, line 208:
```suggestion
        try:
            subcommand_action.run_stdout
        except AttributeError:
            pass
        else:
            subcommand_modes.append("stdout")
```

Thoughts on this instead as a minor style cleanup? It keeps the `try` block limited to its actual goal, and gets rid of an explicit call to `getattr()` since we know the attribute name upfront. Similar for below.

cidrblock: think I'll leave the getattr, because it is explicit and doesn't annoy pylint.  adding the else, I always forget about that...

PR #204, opened 2021-05-06
update readme quickstart with correct filepath

thedoubl3j:
path to install location is incorrect. added 2 steps to properly start the install since the setup.py is at the top level and no longer housed in `ansible_navigator`. 

edit: rewrite install since location moved

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ python3 -m venv venv
 source venv/bin/activate
 pip install -U setuptools
 pip install -U ansible
-pip install ../ansible-navigator
+pip install ../ansible_navigator
 ```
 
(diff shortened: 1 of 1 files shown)


thedoubl3j:
@cidrblock can squash if preferred as well

cidrblock:
Doesn't the `cd..` take you to the parent of the demo directory, so you would need to `pip install ./ansible-navigator`?

cidrblock (commented):
the way it was written, the user would be left in the ansible-navigator_demo directory.....

cidrblock:
Ah! The hyphen.... too many name changes, I'll come up with something simple

PR #206, opened 2021-05-06
Remove templar calls, add HumanDumper

### LINKED ISSUE #181: Remove internal ansible call (templar)
depends on #84 so to_yaml from ansible isn't necessary for examples

if #84 does not playout, add support for ansible-navigator internally hosted filter plugins and author a from_yaml plugins and use just jinja's templar

cidrblock:
Fixes #181 
Closes #84   (ty @jborean93 )

- Move the plugin name to the header of the doc, so examples can be grabbed with. `:{{ examples }}`. [couldn't with org.name.plugin  - . delimited collection name]
- Add human dumper from #84 
- add yaml dump func to yaml.py so open, write, and ui, can have the same settings
- add `:{{ examples }}` to doc integration tests
- add remaining doc integration tests
- add fixtures, although unused, for tests that have expected values, so they can be 'seen'
- remove name from a few fixtures


tests/integration/actions/doc/base.py
```diff
@@ -49,33 +49,28 @@ def test(
...
-                assert out in received_output
-        else:
-            updated_received_output = []
-            for line in received_output:
-                mask = "X" * 50
-                if "filename" in line or "│warnings:" in line:
-                    updated_received_output.append(mask)
+        updated_received_output = []
+        for line in received_output:
+            mask = "X" * 50
```
ganeshrn, line 54: This can move outside of for loop

ganeshrn (approved):

tests/integration/actions/doc/base.py
cidrblock: right, I'll label gate and log an issue, TY

PR #211, opened 2021-05-07
Refactor collections action to use new runner API

ganeshrn:
*  Refactor collections action to use new ansible-runner API's
*  Add capability to send log messages from `catalog_collections.py`
   script to collections action
*  Add integration test for collections action
* Fix issue in `catalog_collections.py` script in case only one processor is available
* Add command to `pip install` ansible-navigator to move away from using `python -m ansible_navigator`
  in integration tests.
  TODO:
   Will replace `python -m ansible_navigator` with `ansible-navigator` in a separate after this PR is merged


cidrblock:
Recheck

ganeshrn:
recheck

ansible_navigator/actions/collections.py
```diff
@@ -350,13 +343,25 @@ def _parse(self, output) -> None:
             collection["__version"] = collection["collection_info"]["version"]
             collection["__shadowed"] = bool(collection["hidden_by"])
             if self._args.execution_environment:
+                collection_path_envvar = os.environ.get("ANSIBLE_COLLECTIONS_PATH")
                 if collection["path"].startswith(self._adjacent_collection_dir):
                     collection["__type"] = "bind_mount"
+                elif collection_path_envvar:
```
cidrblock, line 349: Wanted to make sure this is correct, is runner automatically vol mounting anything in the ANSIBLE_COLLECTION_PATH?, while reviewing the fixtures, I noticed they were showing as "contained" instead of "bind_mount"

cidrblock (approved):

ansible_navigator/actions/collections.py
ganeshrn: `ANSIBLE_COLLECTIONS_PATH` collection path is not mounted, as per our discussion only the CWD is mounted. If user wants to introspect collection on local system they can always disable EE setting. In the test case the CWD is changed to fixtures collections top-level folder for it to be identified within EE

cidrblock: got it, will remove

cidrblock: Interesting, we do pass the ANSIBLE_COLLECTIONS_PATH to the EE, so it it is set to a deeply nested directory in the CWD, the collections will get picked up. I'm going to allow the action to specify envvars to be skipped in runner/api so we don't pass the ANSIBLE_COLLECTIONS_PATH to the EE

ganeshrn: IIUC wIth EE enabled and `ANSIBLE_COLLECTIONS_PATH` path set outside of CWD, it will refers to the path within an EE

cidrblock: Removed the logic, replace with error messages for the user.

PR #223, opened 2021-05-09
Catch exception in tokenization subsystem per line, log CRITICAL

### LINKED ISSUE #180: Ensure graceful fail for tm_tokenize
ensure the tm tokenize calls are wrapped with a try / broad except so an error can be logged with full traceback information for debugging after the fact

cidrblock:
Fixes #180 

The errors in the log appear like this:

```
210509182154.651 CRITICAL 'ansible_navigator.ui_framework.colorize.render' An unexpected error occured within the tokenization subsystem.  Please log an issue with the following:
210509182154.652 CRITICAL 'ansible_navigator.ui_framework.colorize.render'   Err: '', Scope: 'text.html.markdown', Line follows....
210509182154.652 CRITICAL 'ansible_navigator.ui_framework.colorize.render'   '## Welcome'
210509182154.652 CRITICAL 'ansible_navigator.ui_framework.colorize.render'   Please try again, or disable the text-based user interface with '-m stdout' at the command line.

```
which should give us enough to try and recreate the error.  Rather than exit the app, this will result in lines missing from the output

### DIFF
--- a/ansible_navigator/ui_framework/colorize.py
+++ b/ansible_navigator/ui_framework/colorize.py
@@ -11,10 +11,13 @@ import curses
 import functools
 
 from itertools import chain
-from ..tm_tokenize.grammars import Grammars
-from ..tm_tokenize.tokenize import tokenize
+
 from .curses_defs import CursesLine
 from .curses_defs import CursesLinePart
+
+from ..tm_tokenize.grammars import Grammars
+from ..tm_tokenize.tokenize import tokenize
+
 
 CURSES_STYLES = {
     0: None,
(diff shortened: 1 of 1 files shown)


ansible_navigator/ui_framework/colorize.py
```diff
@@ -104,7 +107,27 @@ def render(self, doc, scope):
...
+                        "  Err: '%s', Scope: '%s', Line follows....", str(exc), scope
+                    )
+                    self._logger.critical("  '%s'", line)
+                    self._logger.critical(
+                        (
+                            "  Please try again, or disable the text-based user"
+                            " interface with '-m stdout' at the command line."
+                        )
+                    )
+                    continue
```
relrod, line 129: What is the actual net result of `continue` here, and skipping the line? Do we not render the line in output at all? Or is it just not colored? At a quick glance it looks like we'd just not render the line at all which is probably nonideal. But I'm not familiar enough with this subsystem to be sure.

cidrblock: I was thinking the same thing, we should return the line as is. Will revise.... Good catch

PR #229, opened 2021-05-10
[config] micro-optimization, short-circuit lookups

relrod:
Change:
- In ApplicationConfiguration methods, return as soon as a match is
  found rather than collecting all matches in a list first.

- In addition to being a micro-optimization, this also gives a more
  useful KeyError for debugging.

### DIFF
--- a/ansible_navigator/configuration_subsystem/definitions.py
+++ b/ansible_navigator/configuration_subsystem/definitions.py
@@ -140,23 +140,24 @@ class ApplicationConfiguration(SimpleNamespace):
     def __getattribute__(self, attr: str) -> Any:
         """Returns a matching entry or the default bwo super"""
         try:
-            found_entry = [
-                entry for entry in super().__getattribute__("entries") if entry.name == attr
-            ]
-            if found_entry:
-                return found_entry[0].value.current
+            for entry in super().__getattribute__("entries"):
+                if entry.name == attr:
+                    return entry.value.current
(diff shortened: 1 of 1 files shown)


cidrblock:
I had a minute and this intrigued me
```
class ApplicationConfiguration(SimpleNamespace):
    """The main object for storing an application config"""

    application_name: str = ""
    entries: List[Entry] = []
    internals: SimpleNamespace
    subcommands: List[SubCommand]
    post_processor = Callable

    initial: Any = None
    original_command: List[str]

    def _get_by_name(self, name, kind):
        try:
            return next(entry for entry in super().__getattribute__(kind) if entry.name == name)
        except StopIteration as exc:
            raise KeyError(name) from exc

    def __getattribute__(self, attr: str) -> Any:
        """Returns a matching entry or the default bwo super"""
        try:
            return self._get_by_name(attr, "entries").value.current
        except KeyError:
            return super().__getattribute__(attr)

    def entry(self, name) -> Entry:
        """Retrieve a configuration entry by name"""
        return self._get_by_name(name, "entries")

    def subcommand(self, name) -> SubCommand:
        """Retrieve a configuration subcommand by name"""
        return self._get_by_name(name, "subcommands")
```
was the best I could come up with, the only thing is I don't think we can return a KeyError for an a not found entry in __getattribute__ because we don't know specifically the caller was looking for an entry

PR #233, opened 2021-05-11
[:replay] rename load to replay, add to welcome screen

NilashishC:
Signed-off-by: NilashishC <nilashishchakraborty8@gmail.com>

will partial solve for #200 
--mode stdout will be in a separate PR

Note to reviewers:  This PR has a ton of fixture only changes due to modifications of the welcome page, as well as older changes that did not update meta data in the fixtures.  (tmux session changes).  The significant changes are in the .py files, which replace load with replay. (cidrblock)


ganeshrn (approved):

cidrblock:
recheck

ganeshrn:
recheck

cidrblock:
there are fixtures that will need to be updated for the tests to pass, fixture generation can happen by running the tests with the test's UPDATE_FIXTURES constant set to true, for example https://github.com/ansible/ansible-navigator/blob/main/tests/integration/actions/doc/test_welcome_interactive_ee.py#L59

The tests integration tests that capture the welcome screen are typically named test_welcome_, meaning they start at the welcome screen and then proceed into an action

cidrblock (changes requested):
see note about fixture updates

cidrblock:
although a more extensive change, this could be done along with #200

cidrblock:
Code changes look great @NilashishC , the only follow on changes were in the fixtures and test.  Some tests are buried deep within the configuration subsystem unit tests and their fixtures.

The fixture changes are not all a result of this changes, some had meta-data changes that were left over from a previous PR and did not get updated.  More attention can be paid to the fixture changes when the data or output changes, the meta data can change swhen the tests name are modified or parameters changed.

cidrblock (approved):

PR #254, opened 2021-05-12
[:config] added --help-config subcommand and test updates

ganeshrn:
*  Added --help-config subcommand to work in stdout mode
   and fetch ansible-config help options
*  Added integration test for --help-config command
*  Refactor integration test to simplify the tmux
   interaction method
*  Added timeout logic to return from tmux interaction
   method to get the actual values displayed in tmux
   terminal
*  Doc updates

### DIFF
--- a/ansible_navigator/actions/config.py
+++ b/ansible_navigator/actions/config.py
@@ -15,7 +15,6 @@ from . import _actions as actions
 from ..app import App
 from ..app_public import AppPublic
 from ..configuration_subsystem import ApplicationConfiguration
-from ..configuration_subsystem.definitions import Constants as C
 from ..runner.api import AnsibleCfgRunner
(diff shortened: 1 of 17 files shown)


ansible_navigator/actions/config.py
```diff
@@ -211,7 +212,15 @@ def _run_runner(self) -> None:
                     return
                 ansible_config_path = exec_path
 
-            kwargs.update({"cmdline": self._args.cmdline})
+            if self._args.cmdline is not C.NOT_SET:
```
cidrblock, line 215: just a note, importing the Constants could be avoided by checking if cmdline is instance list

GomathiselviS:
recheck

cidrblock (approved):

PR #252, opened 2021-05-12
[:ee-details] add new action for execution-environment metadata

cidrblock:
depends-on: https://github.com/ansible/ansible-navigator/pull/257

- add interactive & stdout mode for `ee-details`
- add subcommand & action for `ee-details`
- add integration tests for `:ee-details` in interactive and stdout modes
- add internal collection and playbook to collect ee-details
- include internal collection & playbook in setup.py
- small changes as needed in configuration subsystem to allow for collecting ee-details while running without ee false
- changes to `:run` to allow for custom column settings

PR #257, opened 2021-05-12
[tests] fixes for multiprocessor use and tmux refactor

cidrblock:
closes #264 

- move test classes into separate files
- refactor tmux session to capture command prompt, setup details, log for zuul etc
- changes to calling tests as needed
- enable multiproc testing
- ensure unique names for tmux sessions per test for multiproc
- changes to setup.py for #252 and pypi packaging with older setup tools
- draw screen from top down in ui.py


relrod:
recheck

relrod:
recheck

relrod:
recheck

cidrblock:
recheck

tests/integration/_common.py
```diff
@@ -215,14 +221,43 @@ def __enter__(self):
         venv = os.path.join(shlex.quote(venv_path), "bin", "activate")
 
         # send the config envvar + other set up commands
-        venv = f"source {venv}"
-        navigator_config = f"export ANSIBLE_NAVIGATOR_CONFIG={self._config_path}"
-        set_up_commands = [venv, navigator_config] + self._setup_commands
+        tmux_common = [f"source {venv}"]
```
relrod, line 224:
```suggestion
        tmux_common = [f". {venv}"]
```

cidrblock: will add

cidrblock: (just checking locally)

tests/integration/_common.py
```diff
@@ -206,23 +211,71 @@ def __enter__(self):
...
         if venv_path is None or ".tox" not in venv_path:
             raise AssertionError(
                 "VIRTUAL_ENV environment variable was not set but tox should have set it."
             )
         venv = os.path.join(shlex.quote(venv_path), "bin", "activate")
+        # get the USER before we start a clean shell
+        user = os.environ.get("USER")
+        home = os.environ.get("HOME")
+
+        # get a clean shell and predicatble prompt
```
relrod, line 226:
```suggestion
        # get a clean shell and predictable prompt
```

tests/integration/actions/collections/base.py
```diff
@@ -3,7 +3,8 @@
 import difflib
 import json
 import os
-import shutil
+
+from distutils.dir_util import copy_tree
```
relrod, line 7: It might be worth keeping `shutil.copytree()` here, because of [this](https://www.python.org/dev/peps/pep-0632/).

tests/integration/_common.py
```diff
@@ -232,44 +287,84 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
...
         """get cli prompt"""
+        # give tmux a second to start
+        # start a fresh clean shell, set TERM
+        self._pane.send_keys("env -i bash --noprofile --norc")
+        self._pane.send_keys("export TERM=xterm")
         self._pane.send_keys("clear")
+        bash_prompt_visible = False
         showing = []
-        while len(showing) != 1:
+        while True:
```
relrod, line 360: Can we add another timer here? Otherwise if we get here and have an issue, we stall CI for an hour again.

.gitignore
```diff
@@ -1,6 +1,9 @@
+
```
relrod, line 1: superfluous blank line, nuke?

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
 
-        # send the config envvar + other set up commands
-        venv = f"source {venv}"
-        navigator_config = f"export ANSIBLE_NAVIGATOR_CONFIG={self._config_path}"
-        set_up_commands = [venv, navigator_config] + self._setup_commands
+        # set envars for this session
+        tmux_common = [f". {venv}"]
+        tmux_common.append(f"export HOME={home}")
+        tmux_common.append(f"export USER={user}")
+        tmux_common.append(f"export ANSIBLE_NAVIGATOR_CONFIG={self._config_path}")
```
relrod, line 233: These paths should probably be quoted/escaped in case they ever have a space in them. But this is test code, so maybe not a huge deal.

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+        tmux_common.append("export ANSIBLE_NAVIGATOR_LOG_LEVEL=debug")
+
+        self._test_log_dir = None
+        if user == "zuul":
+            self._test_log_dir = os.path.join(
+                "/home/zuul", "zuul-output", "logs", "anible-navigator", self._test_path
+            )
```
relrod, line 240:
```suggestion
            homedir = os.path.expanduser("~")
            self._test_log_dir = os.path.join(
                homedir, "zuul-output", "logs", "anible-navigator", self._test_path
            )
```

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+        else:
+            self._test_log_dir = os.path.join("./", ".test_logs", self._test_path)
```
relrod, line 242:
```suggestion
            self._test_log_dir = os.path.join(".", ".test_logs", self._test_path)
```

or just

```suggestion
            self._test_log_dir = os.path.join(".test_logs", self._test_path)
```

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
 
-        # send the config envvar + other set up commands
-        venv = f"source {venv}"
-        navigator_config = f"export ANSIBLE_NAVIGATOR_CONFIG={self._config_path}"
-        set_up_commands = [venv, navigator_config] + self._setup_commands
+        os.makedirs(self._test_log_dir, exist_ok=True)
+
+        log_file = os.path.join(self._test_log_dir, "ansible-navigator.log")
+        tmux_common.append(f"export ANSIBLE_NAVIGATOR_LOG_FILE={log_file}")
```
relrod, line 247: Consider escaping quoting/escaping `log_file`, in case the path has spaces/anything else.

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+        playbook_artifact = os.path.join(self._test_log_dir, "playbook-artifact.log")
+        tmux_common.append(
+            f"export ANSIBLE_NAVIGATOR_PLAYBOOK_ARTIFACT_SAVE_AS={playbook_artifact}"
```
relrod, line 250: quote/escape

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+        )
+        collection_doc_cache = os.path.join(self._test_log_dir, "collection_doc_cache.db")
+        tmux_common.append(
+            f"export ANSIBLE_NAVIGATOR_COLLECTION_DOC_CACHE_PATH={collection_doc_cache}"
```
relrod, line 254: quote/escape

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+        )
+
+        set_up_commands = tmux_common + self._setup_commands
         set_up_command = " && ".join(set_up_commands)
+
+        # send setup, wait for the prompt in last line
```
relrod, line 260: Can we return early and fail out somehow if the setup commands fail for any reason? They're `&&`'d so it'll give us back a useful return code, but that's not really useful since we're just screen-scraping. If they fail, we'll still sit around waiting. :thinking:

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
         self._pane.send_keys(set_up_command)
+        while True:
```
relrod, line 262: Can we add a timer here?

tests/integration/_common.py
```diff
@@ -206,23 +212,72 @@ def __enter__(self):
...
+            if prompt_showing:
+                break
+            time.sleep(0.1)
+
+        # capture the setup screen
+        self._setup_capture = self._pane.capture_pane()
+
+        # clear the screen, wait for prompt in line 0
+        self._pane.send_keys("clear")
+        while True:
```
relrod, line 273: And here? (timer)

tests/integration/conftest.py
```diff
@@ -43,8 +43,5 @@ def os_indendent_tmp():
         an_tmp = os.path.join(tmp_real, "an")
     else:
         an_tmp = os.path.join("/tmp", "private", "an")
-    os.makedirs(an_tmp)
-    try:
-        yield an_tmp
-    finally:
-        shutil.rmtree(an_tmp)
```
relrod, line 50: It might still be good to clean up after ourselves, otherwise when iterating locally, we can potentially end up accumulating stuff. For now, it's probably not a huge issue, but in the future if we start using the directory for more, it could be.

I'm guessing this is solving a race condition? It might be better to use per-test tmpdirs instead in that case, and still keep the removal logic. That way tests don't step on each other, but we can still nuke the directories after.

tox.ini
```diff
@@ -21,10 +21,10 @@ deps = -r{toxinidir}/requirements.txt
 commands =
     /bin/bash -c 'grep -FInrq "UPDATE_FIXTURES = True" ./tests && exit 1 || exit 0'
     /bin/bash -c 'podman pull {[base]default_ee} || docker pull {[base]default_ee}'
+    py.test -vvvv -n 4 --dist=loadfile --maxfail=15 --durations=100 --cov ansible_navigator --cov share --cov-report term-missing --cov-branch --showlocals
```
relrod, line 24: :tada:

tests/integration/actions/collections/base.py
cidrblock: I did this because copytree doesn't have a exists_true until 3.8 :(

tests/integration/_common.py
cidrblock: got it

.gitignore
cidrblock: doh! done

tests/integration/_common.py
cidrblock: quoted

tests/integration/_common.py
cidrblock: added a timer for the prompt after, and we'll log the setup screen

tests/integration/_common.py
cidrblock: good idea

tests/integration/_common.py
cidrblock: time_everthing, ty

tests/integration/conftest.py
cidrblock: we needed a directory of predictable length between dev boxes and CI so the column widths don't vary, I'll add a rm to tox, yes because with xdist it was getting deleted during other test runs

cidrblock:
recheck

cidrblock:
recheck

tests/integration/actions/collections/base.py
cidrblock: added one to common for tests, thx!

cidrblock:
recheck

cidrblock:
recheck

ganeshrn (approved):

cidrblock:
recheck

cidrblock:
recheck

PR #264, opened 2021-05-13
[tests] use . instead of source, to source

relrod:


cidrblock:
recheck

cidrblock:
should be covered in #257

PR #287, opened 2021-05-17
[tests] Fix log append to account for container engine, promote fixture

### LINKED ISSUE #286: [tests] `tests/unit/test_log_append.py` does not account for the use of docker
the cli args should set the container engine to the one available

cidrblock:
Fixes #286 

- promote container runtime fixture out of integration directory
- use container runtime fixture for `--ce` setting in log append unit tests


cidrblock (commented):
approved

PR #289, opened 2021-05-18
[stdout] return exit code for stdout mode subcommands

### LINKED ISSUE #259: expose ansible-playbook exit code with --mode=stdout
For CI testing, we are running ansible-navigator using --mode=stdout in an ansible shell task.

We need the ability to use the exit codes from ansible-navigator command, to identify if the run command playbooks passed or failed.

Today we are using ansible-test, to run our integration playbooks but hoping to replace that with ansible-navigator.

I would prefer not to inspect the json output ourself, looking for failures.

cidrblock:
Fixes #259 

- When runner reports "failed", `sys.exit(1)`
- tests to support

### DIFF
--- a/ansible_navigator/actions/__init__.py
+++ b/ansible_navigator/actions/__init__.py
@@ -15,7 +15,7 @@ names = actions.names_factory(__package__)
 
 kegexes: Callable = actions.kegexes_factory(__package__)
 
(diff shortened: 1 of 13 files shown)


relrod (approved):

cidrblock:
Related: #291

Issue #277, opened 2021-05-16
[docs] Repo should contain a FAQ to capture questions that arise


cidrblock:
#276 should be included

PR #296, opened 2021-05-18
[:config] Add `-c` and `--config` to `config` subcommand

### LINKED ISSUE #118: Add subparse for config action
Add argument subparser for config action to parse `ansible-config` command-line option to accept path of the configuration file as input in Text user interface (TUI) mode.

cidrblock:
Fixes #118 

- these get passed in either stdout mode or interactive
- remove setting the ANSIBLE_CONFIG variable
- updated docs & tests
- environment_variable_override now requires the full envvar, to we can leverage the ANSIBLE_CONFIG envvar (#298)

If ansible-navigator is being used without an EE, ansible is running on the local file system as the local user, so the ansible.cfg file will be found normally
if ansible-navigator is being used with an EE, an ansible.cfg file in the CWD will be used, otherwise the user would need to have it set and volume mount it's destination (mention in #277)


ansible_navigator/actions/config.py
```diff
@@ -223,6 +226,8 @@ def _run_runner(self) -> None:
 
             if self._args.help_config is True:
                 pass_through_arg.append("--help")
+            elif isinstance(self._args.config, str):
```
ganeshrn, line 229: this can be an `if` condition instead of `elif`

ansible_navigator/utils.py
```diff
@@ -115,6 +115,34 @@ def environment_variable_is_file_path(
...
+def find_ansible_config_file() -> Tuple[List[LogMessage], List[str], Optional[str]]:
+    """Load config file(first found is used): CWD, HOME, /etc/..."""
+
+    messages: List[LogMessage] = []
+    errors: List[str] = []
+    path = None
+    filename = "ansible.cfg"
+
+    # Current working directory
+    potential_paths = [abs_user_path(os.path.join(os.getcwd(), filename))]
```
ganeshrn, line 127: The Ansible config file can be set using the ANSIBLE_CONFIG env variable also. Another option to get the config file is to execute the `which ansible` command using runner `run_command` API and fetch the config_file path value.

ansible_navigator/actions/config.py
cidrblock: got it, thx

ansible_navigator/utils.py
cidrblock: removed this logic entirely

ganeshrn (approved):

PR #305, opened 2021-05-20
Remove .tox check in VIRTUAL_ENV for tmux

ganeshrn:
*  While running testcase locally the user enabled
   virtual env set the enviornment variable `VIRTUAL_ENV`
   required for intial tmux session and the check for `.tox`
   in not required in that case.
*  In CI run since virtual enviornment is enabled by `tox`
   the `.tox` will always be presnet in the `VIRTUAL_ENV` path
   and hence the check is not required


cidrblock:
recheck

PR #306, opened 2021-05-20
[cli] Provide `hints` when errors are encountered

### LINKED ISSUE #201: Improve error messages in post processing
for instance, if we error because podman is not available we should say try `--ce docker` or `--ee false`

same for other subcommands

cidrblock:
image

Fixes #201 

- replace all `errors` with ExitMessages
- add `hints` after each error in post processing
- add tests
- fix collection_doc_cache_path so it has a cli entry
- fix missing choices for log append
- 2 missing fixtures added (although no used b/c text is being found for the assert)

Note to reviewer: This PR has a lot of lines changed, which is a result of changing err to exit_msg to be consistent through all files and the tests


s-hertel (approved):
LGTM

PR #312, opened 2021-05-20
[ui] Correct footer previous and next `+` and `-`

### LINKED ISSUE #294: on bar section("footer" with options) - the previous symbol(-) and the next(+) are switched
## Steps to reproduce
1 - run the command ansible-navigator to use the GUI
2 - go to collections
3 - choice the `2` option
3 - on the footer, we could see the information
image (1)

## Result
 - if you click on `-` option, you will go to the previous and if you try use the `+` option you go to the next, therefore, they are switched.

## Suggestion:
 - switch the text, replacing for:
```
`+` go to the next
`-` go to the previous
```

cidrblock:
Fixes #294 

- fixtures updates as needed
- small change in fixture generator to not put additional_information in fixture if None (was causing unnecessary fixture updates as not all tests use this)


relrod (approved):
hah

PR #317, opened 2021-05-21
[cli] Add a pull policy parameter and pull the image during application initialization

### LINKED ISSUE #175: Download EE ahead of time (prior to any runner call)?  - or document how to podman/docker pull
1) We could have a post processing step to get the initial EE, which would show in stdout before curses is enabled.....

2) Subsequent images could be pulled in the runner/api code, at least we could log there..

3) Or maybe just check to see if the image is installed, and log a message saying there maybe a delay and let runner grab it

I think #3 is the quickest fix and could stay in place even if #1 or #2 is done

add pull policy logic

https://kubernetes.io/docs/concepts/configuration/overview/#container-images

cidrblock:
image

Fixes #175 

unit tests included


ansible_navigator/image_manager/image_puller.py
```diff
@@ -0,0 +1,158 @@
...
+        """pull the image, print to stdout
+
+        podman writes to stderr
+        docker writes to stdout
+        """
+        try:
+            if self._container_engine == "podman":
+                subprocess.run([self._container_engine, "pull", self._image], check=True)
+            elif self._container_engine == "docker":
+                subprocess.run(
```
ganeshrn, line 139: Can we pipe both `stdout` ans `stderr` for both docker and podam and display stdout to stderr on console?

cidrblock: I wanted to grab real steer where I could to format in our formet

cidrblock: For docker, the progress bars are returned as stdout, and errors as stderr, so we can let stdout be printed on the screen, and capture stderr, render it with our [ERROR] format and log it.

For podman, the progress bars and errors are returned as stderr, so we have no choice but to let it print on the screen because there is no way to know what is just a message or error.

So this was the best case scenario, I suspect most users will be using docker, which in this scenario "looks" the best because we can color the errors, separate from the progress bars.

ganeshrn (approved):

PR #337, opened 2021-05-25
[ui] non blocking notification

cidrblock:
This is a non blocking notification, the single field immediately returns and the while loop exited.

(no buttons)


cidrblock:
not in use yet, staging for later use

Issue #238, opened 2021-05-11
:run with --mode stdout produces empty playbook artifact


cidrblock:
```
{
    "version": "1.0.0",
    "plays": [],
    "stdout": [],
    "status": "successful",
    "status_color": 10
}
```

PR #340, opened 2021-05-26
[:run] enable artifact creation with mode stdout

cidrblock:
when artifact creation is enabled, collect events and print stdout.

this suggests if mode=stdout and playbook_artifact_enable=True, playbook prompts will not work, this has been noted in the help.

Fixes: #238 

tested locally

### DIFF
--- a/ansible_navigator/actions/run.py
+++ b/ansible_navigator/actions/run.py
@@ -230,6 +230,14 @@ class Action(App):
         self._task_list_columns = task_list_columns
         self._content_key_filter = content_key_filter
 
+    @property
+    def mode(self):
+        """if mode == stdout and playbook artifact creation is enabled
+        run in interactive mode, but print stdout"""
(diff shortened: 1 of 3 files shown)


ansible_navigator/actions/run.py
```diff
@@ -530,12 +538,18 @@ def _take_step(self) -> None:
     def _run_runner(self) -> None:
         """spin up runner"""
         executable_cmd: Optional[str]
+
+        if self.mode == "stdout_w_artifact":
```
ganeshrn, line 542: The if block can be moved within `def mode` property. All the mode related setting can be handled in one place by doing so.

cidrblock: The reason it's here, is that def mode needs to return 1 of 3 modes:

1 = interactive
2 = stdout_w_artificat
3 = stdout

The second is needed within handle message to determine if the stdout mesaages should be printed

PR #348, opened 2021-05-26
[:images] Add images subcommand

cidrblock:
updated docs too


cidrblock:
approving, so tests can be started.  docs & subcommand entry only

PR #349, opened 2021-05-26
[:images] Switch to a nonblocking notification vs content

cidrblock:
Use the nonblocking notification to show images are being found


cidrblock:
getting ready for tests

Issue #350, opened 2021-05-26
[ui] nonblocking_notification


cidrblock:
when the messages list length is only 1, pad it before and after

remove the logic from :images

PR #351, opened 2021-05-26
[:collections] Switch to nonblocking notification for working on it message

cidrblock:
re #350 

use the popup instead of a content view


cidrblock:
approving

PR #355, opened 2021-05-26
[pkg] Remove extra, uneeded doc updater

### LINKED ISSUE #347: [docs] Remove duplicate doc updater from inside code base
I believe only the utilities version is required and the above was a mistake one late night

cidrblock:
Fixes #347 

tox uses the one in utilities, so do I


cidrblock:
approved, trash collection

PR #368, opened 2021-05-27
[doc] support --list and --list_files with --mode stdout

s-hertel:
@cidrblock It looks like the interactive code path behaves pretty differently (no doc validation for example), so I made plugin name optional for stdout mode only. Thoughts?


cidrblock:
seems to make sense to me, I had forgotten about the non-plugin doc use cases :)

cidrblock (approved):

PR #370, opened 2021-05-27
[configuration] Set container-engine default to `auto`, uses `podman` then `docker` or errors

### LINKED ISSUE #367: [configuration]  container engine auto discovery
we can add a new 3rd option of --ce auto, as the default which will look for and use podman/docker in that order

re #362

cidrblock:
Fixes #367


cidrblock:
let me know if you disagee

PR #372, opened 2021-05-27
[settings] Moves ansible related keys under a parent ansible key in settings file

### LINKED ISSUE #290: [configuration] Consider moving `playbook` `inventories` future `config` and other related keys under an `ansible` key in the settings file
^ this

cidrblock:
Fixes #290


s-hertel (approved):

tests/fixtures/unit/configuration_subsystem/ansible-navigator.yml
```diff
@@ -2,9 +2,12 @@
 ansible-navigator:
   ansible:
     config: /tmp/ansible.cfg
+    cmdline: "--forks 15"
```
s-hertel, line 5: Oh, I just spotted something. This isn't a change specific to this PR since you just moved lines, but I noticed the value of "{{ ansible_forks }}" isn't correct unless this is actually a list: `cmdline: ["--forks", "15"]`

cidrblock: I'll add an issue to check this, I'm 99.6% sure we shlex split them on the way to runner

PR #377, opened 2021-05-27
[:help] Remove the :`ee-details` entry from the `:help`, sort

### LINKED ISSUE #373: delete the :ee-details option on the :help, because we changed it to :images
On the `:help`  there is an option called `:ee-details -eei <image> `, but we changed the name for `:images`

image (2)

we should to remove it, in the image can already see the `:images` option

cidrblock:
Fixes #373


cidrblock:
quick fix, just help

PR #379, opened 2021-05-27
`NO_COLOR` support for accessibility

### LINKED ISSUE #174: NO_COLOR interactive mode

### LINKED ISSUE #394: [forms] With `--display-color false` form field focus is broken because of no decorations
Might need to make the string all caps or put it in [brackets] since we lose all curses decorations with no_color

relrod:
This is an initial go at `NO_COLOR` support using the techniques that @cidrblock and I discussed in a call yesterday.

Fixes #174 
Fixes #394

### DIFF
--- a/ansible_navigator/ui_framework/ui.py
+++ b/ansible_navigator/ui_framework/ui.py
@@ -141,9 +141,6 @@ class UserInterface(CursesWindow):
         self._xform = self._default_obj_serialization
         self._status = ""
         self._status_color = 0
-
-        # curses.curs_set(0)
-        # self._set_colors()
         self._screen: Window = curses.initscr()
         self._screen.timeout(refresh)
         self._one_line_input = FormHandlerText(screen=self._screen, ui_config=self._ui_config)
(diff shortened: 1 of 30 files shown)


ansible_navigator/ui_framework/ui.py
```diff
@@ -477,16 +475,21 @@ def _serialize_color(self, obj: Any) -> CursesLines:
...
         else:
             string = obj
-        colorized = self._colorizer.render(doc=string, scope=self.xform())
-        lines = self._color_lines_for_term(colorized)
-        return lines
+        if self._ui_config:
+            rendered = self._colorizer.render(doc=string, scope=self.xform())
+        else:
+            # TODO: dedupe this from colorize.py
+            rendered = [[{"column": 0, "chars": doc_line, "color": None}] for doc_line in string.splitlines()]
```
cidrblock: we could create a new scope called "no color" or something and move this over there

cidrblock:
recheck

cidrblock (approved):

PR #380, opened 2021-05-27
[:inventory] Catch inventory errors

### LINKED ISSUE #365: [:inventory] Not catching ERRORS!
just like #357

cidrblock:
Fixes #365

tested locally with corrupted ansible.cfg file


ansible_navigator/actions/inventory.py
```diff
@@ -455,9 +456,13 @@ def _collect_inventory_details(self) -> None:
                     inventory_output = "{" + parts[1]
                 else:
                     inventory_output = ""
-            if inventory_err:
-                msg = f"Error occurred while fetching ansible inventory: '{inventory_err}'"
-                self._logger.error(msg)
+            warn_msg = ["Errors were encountered while gathering the inventory:"]
+            warn_msg += inventory_err.splitlines()
+            self._logger.error(" ".join(warn_msg))
+            if "Error" in inventory_err:
```
s-hertel, line 462:
```suggestion
            if "ERROR!" in inventory_err:
```
Here's what I was testing with:
```
$ cat invalid.yml
plugin: notinstalled
```

```
$ ANSIBLE_INVENTORY_ANY_UNPARSED_IS_FAILED=True ansible-navigator inventory -i invalid.yml --graph --ee false
```
(no output)

cidrblock: I didn't see ERROR! when I recreated the error with an invalid ansible.cfg, I think it might be being removed back in runner

cidrblock: image

cidrblock: (btw, the dupe is a known issue & logged)

cidrblock:
if/when the signature from runner changes WRT errors, we can fix here

PR #383, opened 2021-05-28
Update container workdir if EE enabled

ganeshrn:
*  The host CWD passed to the runner API
   will be volume mounted with EE if enabled
   and will be used as workdir within the 
   container
*  Pass CWD for run action so as to a similar
   experience while running ansible on local
   sytem VS EE, this enables the use of `ansible.cfg`
   in CWD of host with EE.


cidrblock (approved):

cidrblock:
recheck

PR #384, opened 2021-05-28
[:replay] support replay with mode stdout

NilashishC:
Signed-off-by: NilashishC <nilashishchakraborty8@gmail.com>

- Fixes https://github.com/ansible/ansible-navigator/issues/200

### DIFF
--- a/ansible_navigator/actions/replay.py
+++ b/ansible_navigator/actions/replay.py
@@ -5,24 +5,22 @@ from . import _actions as actions
 from .run import Action as BaseAction
 
 
-class MissingAttribute:
-    # pylint: disable=too-few-public-methods
-    """Raise an attribute error for any get"""
-
-    def __get__(self, instance, owner):
-        raise AttributeError()
-
-
 @actions.register
-class Action(MissingAttribute, BaseAction):
+class Action(BaseAction):
 
     # pylint: disable=too-many-instance-attributes
     """:replay"""
+
+    @property
+    def mode(self):
+        return self._args.mode
+
+    def run_stdout(self) -> int:
+        """Replay artifact file with mode stdout"""
+        return int(self._init_replay())
 
     KEGEX = r"""(?x)
             ^
             (?P<replay>rep(?:lay)?
             (\s(?P<params_replay>\S+))?)
             $"""
-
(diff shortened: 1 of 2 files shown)


NilashishC:
recheck

tests/unit/configuration_subsystem/test_invalid_params.py
```diff
@@ -141,15 +141,3 @@ def test(subcommand, param, look_for):
     else:
         # argparse choice error
         test(subcommand, "", "choose from")
-
```
cidrblock, line 144: this test can stay, but now just for collection, in fact, images needs to be added here

ansible_navigator/actions/replay.py
```diff
@@ -5,24 +5,22 @@
...
-
 @actions.register
-class Action(MissingAttribute, BaseAction):
+class Action(BaseAction):
 
     # pylint: disable=too-many-instance-attributes
     """:replay"""
 
+    @property
+    def mode(self):
```
cidrblock, line 15: does this need to be here?  :run has a mode property, we should be able to use that I think

ansible_navigator/actions/replay.py
```diff
@@ -5,24 +5,22 @@
...
+        return self._args.mode
+
+    def run_stdout(self) -> int:
```
cidrblock, line 18: since run for replay uses the run func in :run, can we use the run_stdout func in run too?  (just so interactive and stdout have the same pattern)?

ansible_navigator/actions/replay.py
NilashishC: The `mode` property in run becomes "stdout_w_artifact". That is _fine_ from _init_replay()'s workings since the if-conditions are branched with == "interactive" but for logging with the correct mode, we'd have to replace "stdout_w_artifact" with "stdout". But I'm okay with removing `mode` from replay.

tests/unit/configuration_subsystem/test_invalid_params.py
NilashishC: +1 I should not have removed the entire test case. :-)

cidrblock (approved):

PR #385, opened 2021-05-28
[:run]  Clean up progress bars

cidrblock:
image
small improvement for the progress bars, 
- I didn't like how crowded it felt, so going back to u2587
- turning off the inverse
- reducing the progress bar size to 8, to match "COMPLETE" 
- moved the status box 1 char to the right to it right aligns with the column above
- the status will be width 12 to nicely pad SUCCESSFUL.

update fixtures as needed


cidrblock:
small style change

PR #389, opened 2021-05-28
[:doc] Use playbook dir or CWD for playbook dir

### LINKED ISSUE #208: :doc for plugin in cwd isn't found
If a playbook value is not provided, the cwd should be included in the search directory for docs

cidrblock:
Fixes #208


cidrblock:
recheck

PR #395, opened 2021-05-30
[docs] Settings doc updates

### LINKED ISSUE #237: Ansible-Navigator Setting - some grammatical corrections
Note
The settings file can either be in JSON or YAML format.
For settings in JSON format (add comma) the file name should be ansible-navigator.json  (add period)
For settings in YAML format (add comma) the file name can be either ansible-navigator.yml or ansible-navigator.yaml.
The first found matched directory path (based on order mentioned above) can have only one valid settings file. If in case  (remove in case?) more than one settings files (file) (with different supported extensions) are (is) found (add comma) it will result in an error to avoid conflict.

cidrblock:
Fixes #237

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -10,8 +10,12 @@ ansible-navigator settings
 The ansible-navigator settings file
 ========================================
 
-Several options in ``ansible-navigator`` can be configured by making use of a
-settings file. The settings file can live in one of two places.
+Settings for ``ansible-navigator`` can provided at the command line,
+set using an environment variable or specified in a settings file.
+
+The settings file name and path can be specified with an environment variable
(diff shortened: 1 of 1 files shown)


docs/settings.rst
```diff
@@ -10,8 +10,12 @@ ansible-navigator settings
 The ansible-navigator settings file
 ========================================
 
-Several options in ``ansible-navigator`` can be configured by making use of a
-settings file. The settings file can live in one of two places.
+Settings for ``ansible-navigator`` can provided at the command line,
```
relrod, line 13:
```suggestion
Settings for ``ansible-navigator`` can be provided at the command line,
```

cidrblock: TY

PR #426, opened 2021-06-02
[docs] What's an extention? (typo)

cidrblock:


cidrblock:
Simple, self approved

Issue #328, opened 2021-05-25
[runner] Refactor runner API's to use the changed parameter name for CWD


ganeshrn:
Rename the `cwd` parameter in runner API's to `host_cwd` due to the change added in PR https://github.com/ansible/ansible-runner/pull/689

PR #428, opened 2021-06-03
Update runner API cwd parameter name to host_cwd

ganeshrn:
Fixes https://github.com/ansible/ansible-navigator/issues/328

As par to PR https://github.com/ansible/ansible-runner/pull/689
the cwd parameter was renamed to host_cwd for sake of clarity
when running with execution environment.

### DIFF
--- a/ansible_navigator/actions/collections.py
+++ b/ansible_navigator/actions/collections.py
@@ -282,7 +282,7 @@ class Action(App):
                 {"container_volume_mounts": self._args.execution_environment_volume_mounts}
             )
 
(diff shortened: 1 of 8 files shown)


ganeshrn:
The test will fail for now as `ansible-runner` latest alpha release does not have the dependent PR https://github.com/ansible/ansible-runner/pull/689.
After ansible-runner is released with this PR updated the minimum version requirement https://github.com/ansible/ansible-navigator/blob/main/setup.cfg#L36

PR #431, opened 2021-06-03
Add issue template

ganeshrn:
* Shamelessly copied from https://github.com/ansible/ansible/tree/devel/.github/ISSUE_TEMPLATE
  and customised for ansible-navigator


cidrblock:
A few ansible specific things we should talk about but a great start

PR #456, opened 2021-06-07
update runner alpha version

ganeshrn:
update runner alpha version to `2.0.0a5`


ganeshrn:
Added as part of PR https://github.com/ansible/ansible-navigator/pull/428 to avid circular dependency.

Issue #457, opened 2021-06-07
Comma separated lists don't work 


shanemcd:
### Summary

This works:

```
$ ansible-navigator -m stdout run ~/ansible-runner/demo/project/test.yml -i localhost, -c local
```

This does not:
![image (2)](https://user-images.githubusercontent.com/773186/121055549-5aa3d500-c78b-11eb-8f43-7d1359357a61.png)

### Issue Type

Bug Report

### Ansible-navigator Version

```console
54b92f9a80ced72a135dd2f708806b3eca119cdc
```

### Ansible-runner Version

```console
2.0.0.0a5
```
```


### Settings

```console
N/A
```

### Command

```console
See above
```

### Ansible Version

```console
N/A
```

### Ansible Configuration

```console
$ ansible-navigator config dump --only-changed -m stdout
```

### OS / Environment

Fedora 34

### Steps to Reproduce

See issue summary

### Expected Results

See issue summary

### Actual Results

```console
See issue summary
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

s-hertel:
FYI (since I don't have access to assign myself), I'm working on a fix for this.

PR #460, opened 2021-06-07
Handle different types of inventory sources

s-hertel:
Fixes https://github.com/ansible/ansible-navigator/issues/457

Inventory can be a path, file, comma-separated list of hosts, or any string that meets a plugin's requirements.

This also fixes:

  * Hosts without variables
  * Empty inventory (shouldn't error by default if there are perhaps-unrelated warnings in stderr - INVENTORY_ANY_UNPARSED_IS_FAILED and INVENTORY_UNPARSED_IS_FAILED config options do that)
  * Displaying fatal `ansible-inventory` errors


ganeshrn (commented):
Code changes look good to me. Can you please add test cases to cover the scenario's fixed by this PR?

PR #464, opened 2021-06-09
[tests] add mypy type stub deps

relrod:
Change:
- This is now required with newer mypy, it seems.

Test Plan:
- `tox -re type`


relrod:
recheck

PR #466, opened 2021-06-10
[dnm] Add demo mode which shows single key presses as an overlay on the screen

cidrblock:
image

[asciicast](https://asciinema.org/a/evr1UXCKS361FJfYlb2GI6b3u)

- intended for use during presentations or training sessions
- add `--demo` for demo mode
- Demo mode shows single key presses on the screen
- Add post process to detect pyfiglet
- Pass args.demonstration to UI


relrod:
I feel like this might be better left to an external thing (the screen capture tool in the case of recordings, or another tool running (for presentations) that shows current keypresses). But maybe others have other opinions.

cidrblock:
RE pyfiglet, good call. Although it wasn't _really_ a runtime dep, there's no reason to pull it in...

cidrblock:
The more I think about this, I would like to have this included for a while. Not a permanent thing, but during early adoption, sharing, demoing and showing others, I think this will be handy.  I would have loved it for some early demos I gave.

My only concern about leaving this to a 3rd party recording tools is that it won't work live, and will not be able to differentiate between colon commands and single key presses.

relrod:
> My only concern about leaving this to a 3rd party recording tools is that it won't work live

There are [external tools](https://www.thregr.org/~wavexx/software/screenkey/) that [can](https://boinx.com/mousepose/) [show](https://github.com/keycastr/keycastr) keypresses [without](https://github.com/Code52/carnac) [recording](https://github.com/KaustubhPatange/Kling).

In the interest of avoiding scope creep, I'm still -1 on this, *especially* for 1.0. But even beyond that, I still think this is better left to a dedicated tool.

cidrblock:
+100 for the embedded links :)  I had tried a couple of those and realized they would never work with asciinema.

Nor does it appear asciinema will support an overlay.

I will mark this as dnm, and we can leave it out of 1.0.0.  I would like it left around until we get a better asciinema in the README.

cidrblock:
This can either be removed or merged after 1.0.0

PR #470, opened 2021-06-13
Reformat example settings file

ptoal:
The way the example settings file was formatted gave the appearance that 'mode' and 'playbook-artifact' were children of 'logging'.

Comment out all of the lines in the example, and fix indentation.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -36,60 +36,60 @@ You can copy the example settings file below into one of those paths to start yo
 .. code-block:: yaml
 
     ---
-    ansible-navigator:
-    #   ansible:
-    #     config: /tmp/ansible.cfg
-    #     cmdline: "--forks 15"
-    #     inventories:
-    #     - /tmp/test_inventory.yml
-    #     playbook: /tmp/test_playbook.yml
-    #   app: run
-    #   collection-doc-cache-path: /tmp/cache.db
-    #   color:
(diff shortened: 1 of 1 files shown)


cidrblock:
hey @ptoal thanks for bringing attention to this, the intention was to make that sample cut-and-paste usable, but looking at it now, I can see how it does look goofy.  That doc is maintained and checked by code with this utility:

https://github.com/ansible/ansible-navigator/blob/main/utilities/doc_updater.py#L234

based on this test fixture:

https://github.com/ansible/ansible-navigator/blob/main/tests/fixtures/unit/configuration_subsystem/ansible-navigator.yml

So, if you think everything should be commented out, the script would need to be modified and rerun as it is used to ensure the docs are correct.

let me know, i can take care of this too if you want

ptoal:
I don't have strong feelings about how the example is presented, as long as it is intuitive and the indentation doesn't cause confusion (as it did for me).

Perhaps adding some newlines between major sections might help?

cidrblock:
yeah, that might help...  It would need to be done in the fixture tho still and the utility run, want me to take care of that, I can have you review when the PR is up

ptoal:
If you don't mind, that would be great. You can probably do it in 1/10th the time it would take me. :)

cidrblock:
np, brb

PR #473, opened 2021-06-14
[ui] vscode trash fix, the appearance of clear

cidrblock:
Print a screen of blank lines for vscode (and others as they arise) to prevent bleed through of pre curses stdout/stderr in the TUI

image

### DIFF
--- a/ansible_navigator/utils.py
+++ b/ansible_navigator/utils.py
@@ -133,6 +133,21 @@ def check_for_ansible() -> Tuple[List[LogMessage], List[ExitMessage]]:
     return messages, exit_messages
 
 
+def clear_screen() -> None:
+    """print blank lines on the screen, preserving scrollback
+
+    Note: In certain cases, xterm.js based terminals show stdout
+    under the initial curses output, this was found with vscode
+
+    Rather than issueing "clear" print blank lines to preserve
+    the users' scrollback buffer
+    """
+    affected_terminals = ["vscode"]
+    if os.environ.get("TERM_PROGRAM") in affected_terminals:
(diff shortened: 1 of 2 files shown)


cidrblock:
recheck

ansible_navigator/utils.py
```diff
@@ -133,6 +133,21 @@ def check_for_ansible() -> Tuple[List[LogMessage], List[ExitMessage]]:
     return messages, exit_messages
 
 
+def clear_screen() -> None:
+    """print blank lines on the screen, preserving scrollback
+
+    Note: In certain cases, xterm.js based terminals show stdout
+    under the initial curses output, this was found with vscode
+
+    Rather than issueing "clear" print blank lines to preserve
```
relrod, line 142:
```suggestion
    Rather than issuing "clear" print blank lines to preserve
```

ansible_navigator/utils.py
```diff
@@ -133,6 +133,21 @@ def check_for_ansible() -> Tuple[List[LogMessage], List[ExitMessage]]:
...
+    the users' scrollback buffer
```
relrod, line 143:
```suggestion
    the user's scrollback buffer
```

relrod:
So this likely works (I didn't test it), but I'm still unclear on why it's needed. Other curses apps (htop, irssi) work fine without doing this. They correctly don't display old terminal text while they are running, and when they close, the old terminal "stuff" is replaced like normal without having to scroll.

What is different about our use of curses? Is this a bug in the Python curses bindings?

cidrblock:
@relrod I really really really wish I knew....  I tried everything I could think of.

I'll add an issue for research in the future

Issue #401, opened 2021-05-31
[runner] Get stdout return code from runner API


ganeshrn:
Get stdout return code from runner API.
Refactor the code in the navigator to get the return code from runner after PR https://github.com/ansible/ansible-runner/pull/696 is merged.

Related to #291

PR #477, opened 2021-06-15
Add runner return code in stdout mode

ganeshrn:
Fixes https://github.com/ansible/ansible-navigator/issues/401

*  Add support to accept return code from runner API's
*  Modify config, doc, inventory, collection and images
   action to get return code from runner API's
*  Update ansible-runner version to 2.0.0b1 which has
   the corresponding runner changes


cidrblock:
recheck

cidrblock (approved):

Issue #453, opened 2021-06-04
Set lower/upper major version constraint for all python dependencies in 1.0.0


cidrblock:
### Summary

Add major version constraint for all python packages in 1.0.0

main can stay unconstrained to detect future issues

### Issue Type

Bug Report

### Ansible-navigator Version

```console
$ ansible-navigator --version

N/A
```

### Ansible-runner Version

```console
$ ansible-runner --version

N/A
```

### Settings

```console
$ cat ansible-navigator.yaml

N/A
```

### Command

```console
$ ansible-navigator run site.yaml

N/A
```

### Ansible Version

```console
$ ansible --version

N/A
```

### Ansible navigator logs

```console
$ cat ./ansible-navigator.log

N/A
```

### Execution environment details

```console
$ ansible-navigator images quay.io/ansible/ansible-runner:devel

N/A
```

### Container enginer details

```console
$ podman info

N/A
```

### Ansible Configuration

```console
$ ansible-navigator config dump --only-changed -m stdout

N/A
```

### OS / Environment

N/A

### Steps to Reproduce

```yaml (paste below)

```

### Expected Results

constraint in place

### Actual Results

```console
constraint not in place
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

cidrblock:
@ansible/devtools Thoughts?

ganeshrn:
Need to integrate piptools to create dependency tree for different environments.

PR #488, opened 2021-06-21
Relax versions

cidrblock:
Leaving runner with constraints, as well as `onigurumacffi`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,11 +34,11 @@ classifiers =
 packages = find:
 install_requires =
     ansible-runner >=2rc1,<3
-    jinja2 >=3,<4
+    jinja2
     onigurumacffi >=1.1.0,<2
-    pyyaml >=5.4.1,<6
-    importlib-resources >=5.1.4,<6; python_version < "3.9.0"
-    dataclasses >=0.8,<1; python_version < "3.7.0"
+    pyyaml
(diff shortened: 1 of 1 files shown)


jimi-c (approved):

cidrblock:
recheck

setup.cfg
```diff
@@ -34,11 +34,11 @@ classifiers =
 packages = find:
 install_requires =
     ansible-runner >=2rc1,<3
-    jinja2 >=3,<4
+    jinja2
```
relrod, line 37: Can I propose we cap at <4 instead of not having any upper bound at all? It would be really nice to avoid all the hardcoded checks that core has to do. At least if we cap to only support 2.y and 3.y we narrow down how many of those internal checks we need to add. Not having versioned dependencies has bitten us several times in core (in the time I've been there), and particularly on the `ansible-test` side of things, a lot of effort recently has gone into ensuring that everything is capped for what we actually want to support.

I think having no bounds is asking for trouble. We're effectively saying we support things that we have no way of knowing what they will look like

cidrblock:
recheck

setup.cfg
cidrblock: can we revisit after 1.0.0?  I think we should have nightly testing to vet any new version of deps that come out, I've reopened https://github.com/ansible/ansible-navigator/issues/453

PR #529, opened 2021-08-05
[WIP]: Add venv tox entrypoint

pabelanger:
It is possible users want to have tox bootstrap a sandbox environment to
run ansible-navigator from. In this case, zuul so we can do better
integration testing.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -65,6 +65,8 @@ allowlist_externals = ansible-playbook
 commands =
     mypy --txt-report ./{[base]mypy_tmp_dir} ./{[base]pkg_name} ./tests ./share
 
+[testenv:venv]
+commands = {posargs}
 
 [flake8]
 # E123, E125, E203 skipped as they are invalid PEP-8.


tox.ini
```diff
@@ -65,6 +65,8 @@ allowlist_externals = ansible-playbook
 commands =
     mypy --txt-report ./{[base]mypy_tmp_dir} ./{[base]pkg_name} ./tests ./share
 
+[testenv:venv]
```
webknjaz, line 68: @pabelanger why not just use `tox --devenv` for this? No need for a separate toxenv.

pabelanger: Mostly because this has been the only way to do it until https://github.com/tox-dev/tox/pull/1644.  I'll have to see how the command works. But for now, we can close this PR, I've made the changes in another repo.

pabelanger: sorry: https://github.com/tox-dev/tox/pull/1326

PR #534, opened 2021-08-18
Use `extend-` versions of the flake8 options

webknjaz:


ssbarnea (approved):
Good changes. Maybe after that maybe you also move it to pyproject.toml?

webknjaz:
> Good changes. Maybe after that maybe you also move it to pyproject.toml?

I don't think so: the current config is relatively small but when it becomes bigger it kinda makes sense to have it in a properly separated file. But I understand where it's coming from, I got lost in the project configs several times when I tried to find where this is configured.

webknjaz:
recheck

PR #536, opened 2021-08-19
Add initial docs infra

webknjaz:
View this PR build in your browser: https://ansible-navigator--536.org.readthedocs.build/en/536/. [demo]


pabelanger:
Another option is to use the ansible-tox-docs job, ansible-runner is an example: https://github.com/ansible/ansible-runner/blob/devel/.zuul.d/project.yaml#L11

Which we pull back to the log server and render via a Docs preview site artifacts: https://dashboard.zuul.ansible.com/t/ansible/build/fbd85a1615b241389d24a6b36a217058/artifacts / https://d3188a306cbed441caec-b1bff2579d40a74e7b81412336b63eed.ssl.cf1.rackcdn.com/798/ba330275830c6976c629958f5fd9e8866bb51937/check/ansible-tox-docs/fbd85a1/docs/

One less 3rd-party service to depend on

webknjaz:
@pabelanger that could be added later but currently it's out of the scope in this PR. Also, AFAIK Zuul cannot add a nice status with a direct link to the rendered site, right?

pabelanger:
Where would you like the direct link? In the 'Details' link of the status, right?

Yes, we have the ability to update it, infact that is how the docs job worked initially but people wanted the logs first, so we default to the dashboard now.

This can be changed per project

pabelanger:
github comments is an option too, but we stopped doing that as people thought it was too spammy

webknjaz:
> github comments is an option too, but we stopped doing that as people thought it was too spammy

I'm specifically talking about Checks API, they are non-invasive.

webknjaz:
> Where would you like the direct link? In the 'Details' link of the status, right?

Yes, this.

> Yes, we have the ability to update it, infact that is how the docs job worked initially but people wanted the logs first, so we default to the dashboard now.

The way RTD works is that it puts the logs link while the build is in progress and then it changes it to the rendered site. But RTD has a nice widget allowing you to navigate to the builds page from the site anyway.

pabelanger:
Yes, should be possible. I'll push up example job change in a bit, so you can see.

webknjaz:
Well, I still prefer RTD because it's a well-known ecosystem that's easy to manage w/o having to master Zuul first. And you get a nice short URL too.

pabelanger:
> Well, I still prefer RTD because it's a well-known ecosystem that's easy to manage w/o having to master Zuul first. And you get a nice short URL too.

I wouldn't say it is specific to zuul, it is more using the existing shared jobs at platform level. Given each project has limited resources for CI, using common jobs help ease the burden.

webknjaz:
I don't see how this changes the level of burden because it's already configured and works. If anything, Zuul can add more burden because it doesn't have the best UX which makes it harder to navigate to places. That is not taking into account the maintenance part.

pabelanger:
> I don't see how this changes the level of burden because it's already configured and works. If anything, Zuul can add more burden because it doesn't have the best UX which makes it harder to navigate to places. That is not taking into account the maintenance part.

Right, when it works there is no burden I agree. When it doesn't, then people have to get involved, which is where my comments were coming from. I hear that you are not a fan of zuul, and I understand that. I am mostly sharing, there are other projects in the ecosystem that do have doc publishing jobs, which seem to do almost everything (expect from some UX changes). _If_ we wanted to work towards a shared CI system, that could be one way of getting closer to it. Versus adding a new process to test documentation changes for the project. I admit I haven't see any other projects using this service.

ssbarnea (changes requested):
Lets fix zuul jobs first because there is zero pleasure on reviewing code that does not pass our jobs. In general it looks ok and I am happy to use RTD, as we already do with the other repos.

pabelanger:
what is wrong with zuul jobs?

webknjaz:
> I hear that you are not a fan of zuul, and I understand that.

I actually like the idea behind Zuul, it's the UI/UX that is sometimes "too much" for me. And if I had time, I would've gotten involved in improving that.

> _If_ we wanted to work towards a shared CI system, that could be one way of getting closer to it. Versus adding a new process to test documentation changes for the project.

There should be other docs tests that must run in CI but RTD integration is something that a reasonable Python project has by default, in my experience. I saw next to zero projects using something else.

As for the "shared CI", isn't Zuul able to respect required checks that GH exposes? We should rely on that because there are some things that are simply unavailable within Zuul, like Windows and macOS VMs, for example.

> I admit I haven't see any other projects using this service.

Almost every project outside of OpenStack/Zuul ecosystem that uses GitHub and has Sphinx docs uses RTD.

> what is wrong with zuul jobs?

Nothing. It's the internal testing and linting setup of this project that is broken (likely because of the deps update).

webknjaz:
> Lets fix zuul jobs first because there is zero pleasure on reviewing code that does not pass our jobs.

Yep, that was the next thing on my list. But what you're interested in is already reviewable. So don't let Zuul stop you.

cidrblock (approved):

webknjaz:
recheck

PR #538, opened 2021-08-19
Keep pylint under v2.9.0

webknjaz:
This is a temporary measure and pylint should be upgraded once the
linting rule is no longer violated

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,7 +4,7 @@ flake8
 libtmux
 lxml
 mypy
-pylint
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

cidrblock:
Can an issue be opened to fix the linting error and remove the pin, with an example of what needs to be fixed.

Issue linked here so it's not lost, and is connected to the PR

cidrblock (approved):

webknjaz:
The test failure will be addressed by https://github.com/ansible/ansible-navigator/pull/533. Together with this PR, they dead-lock Zuul. So they need to be merged simultaneously.

PR #549, opened 2021-08-29
Set use_develop in tox.ini

pabelanger:
We are using ansible to install navigator into tox, we do not need to do
this. Lets set use_develop and have tox to it for free.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,6 @@ minversion = 1.4.2
 skipsdist = True
 skip_missing_interpreters = true
 
-
 [testenv]
 description = Run pytest under {basepython} ({envpython})
 # the pytest command line is not in the project.toml because of issues 
@@ -34,11 +33,11 @@ setenv =
 passenv = HOME
           USER
           XDIST_CPU
+usedevelop = True
 
 [testenv:clean]
 description = Erase coverage data
 deps = coverage
-skip_install = true
 commands = coverage erase
            rm -rf ./{[base]mypy_tmp_dir}
 
@@ -54,7 +53,6 @@ commands =
 [testenv:report]
 description = Produce coverage report
 deps = coverage
-skip_install = true
 commands =
     coverage report
     cat ./{[base]mypy_tmp_dir}/index.txt
@@ -64,7 +62,6 @@ commands =
 [testenv:smoke]
 description = Run smoke tests under {basepython} ({envpython})
 commands = ansible-playbook tests/smoketests/run.yml
-allowlist_externals = ansible-playbook
 
 [testenv:type]
 description = Verify static typing with MyPy under {basepython} ({envpython})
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -34,11 +33,11 @@ setenv =
 passenv = HOME
           USER
           XDIST_CPU
+usedevelop = True
 
 [testenv:clean]
 description = Erase coverage data
 deps = coverage
-skip_install = true
```
webknjaz, line 41: Don't do this. This env doesn't need navigator at all.

tox.ini
```diff
@@ -10,7 +10,6 @@ minversion = 1.4.2
 skipsdist = True
 skip_missing_interpreters = true
```
webknjaz, line 12:
```suggestion


```

tox.ini
```diff
@@ -61,13 +63,14 @@ commands =
 
 # Note included in the default envlist above since it's destructive
 # (requires tmux, git, runs a bunch of commands, etc.)
+
 [testenv:smoke]
 description = Run smoke tests under {basepython} ({envpython})
 commands = ansible-playbook tests/smoketests/run.yml
-allowlist_externals = ansible-playbook
```
webknjaz, line 67: This needs to be recovered.

tox.ini
```diff
@@ -113,10 +116,8 @@ commands =
 isolated_build = true
 passenv =
   SSH_AUTH_SOCK
-skip_install = false
 usedevelop = true
```
webknjaz, line 120:
```suggestion


```

tox.ini
```diff
@@ -61,13 +63,14 @@ commands =
...
 # (requires tmux, git, runs a bunch of commands, etc.)
+
 [testenv:smoke]
 description = Run smoke tests under {basepython} ({envpython})
 commands = ansible-playbook tests/smoketests/run.yml
-allowlist_externals = ansible-playbook
 
 [testenv:type]
 description = Verify static typing with MyPy under {basepython} ({envpython})
+skip_install = false
```
ssbarnea, line 73: is this really needed? seem unrelated change

tox.ini
```diff
@@ -113,7 +116,6 @@ commands =
 isolated_build = true
 passenv =
   SSH_AUTH_SOCK
-skip_install = false
```
ssbarnea, line 116: also seems unrelated.

ssbarnea (changes requested):

cidrblock:
@ssbarnea @webknjaz Givent he changes in the repo recently is this still needed?

ssbarnea:
There are several problems with it, one of them being that it combines at least 3 different changes. My guess is that the standalone use of pip outside tox with smoke was intended and not an accident as the intent was to see if installation with pip works. What tox does is not always the same thing as pip, especially when using develop. Also testing install outside a virtualenv is desirable as if we use only tox venvs we may fail to identify other bugs, like conflicts with system python packages.

So, I think that PR is no longer needed and I will close it but we can reopen it at any time if needed.

Issue #558, opened 2021-08-31
[maintenance] Implement changelog management


webknjaz:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

I haven't found any changelog in the project. So we should probably start having one. Of course, manually editing it is no fun so it'd be nice to integrate some software for automatic log generation out of per-PR collected change notes.
I've got quite a lot of experience with Towncrier which many projects in the ecosystem use. I also have a Sphinx extension for showing a pre-release changelog. We could go with that or discuss other tools if anybody has strong opinions.

###### Why the changelog should be written by humans for humans
* https://keepachangelog.com
* https://changelog.com/podcast/127

###### Git history, although important, (squashed or not) is not the answer
* https://blog.jaraco.com/dont-use-squash-and-merge/

relrod:
My preference here is to use `antsibull-changelog` since ansible-core already uses it and people are already familiar with that workflow (adding in a fragment with each change).

https://github.com/ansible-community/antsibull-changelog/pull/60 allows antsibull-changelog to be used with other projects, so integrating it here should be relatively easy.

webknjaz:
@relrod do you have a Sphinx extension for displaying the unreleased fragments and a GitHub App for checking their presence? My solution with Towncrier has all of this. And I think it's important for the DX. Also, I'm not a huge fan of the whole RST-in-YAML approach.

cidrblock:
I believe this to be complete. Future enhancements to change log generation can be added as new issues as encountered.

PR #561, opened 2021-09-01
Enable use of release drafter

ssbarnea:
Make use of release-drafter for building release notes and computing
the next version using semver meaning of merged PRs.

This config is based on the one already used by molecule and
ansible-lint. Results can be seen at https://github.com/ansible-community/molecule/releases

The release manager still has the freedom to change the release
notes manually.

Fixes: #558

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,51 @@
+# Format and labels used aim to match those used by Ansible project
+name-template: '$RESOLVED_VERSION'
+tag-template: '$RESOLVED_VERSION'
+categories:
+  - title: 'Major Changes'
+    labels:
+      - 'major'  # c6476b
+  - title: 'Minor Changes'
+    labels:
+      - 'feature'  # 006b75
+      - 'enhancement'  # ededed
+      - 'refactoring'
+  - title: 'Bugfixes'
+    labels:
+      - 'bug'  # fbca04
+  - title: 'Deprecations'
+    labels:
+      - 'deprecated'  # fef2c0
+exclude-labels:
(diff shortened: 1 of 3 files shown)


webknjaz (changes requested):
This needs a discussion. I don't like the idea of untracked change fragments.

ssbarnea:
> This needs a discussion. I don't like the idea of untracked change fragments.

They are not really untracked, we could easily add a rule to update tracked release notes before a release. Keep in mind that this change brings us a decent level of information regarding what was included in a release, including ability to open a discussion thread on each release. It is low maintenance approach, and much better the the current zero-release notes.

We can keep the original ticket open if you want. IMHO, it is an ok tradeoff which proved to work fine for a number of years.

webknjaz:
The fragments are not tracked in Git. It's a vendor-lock on GitHub without any ability to build the change log offline.

webknjaz:
Point is that I cannot generate a changelog with a simple tox command having GitHub unavailable.

ssbarnea:
recheck

ssbarnea:
@webknjaz Please reconsider this. Lets put it from another point of view: I do not expect this to change the project process for producing changelog. I only want to get something meaningful as comment on https://github.com/ansible/ansible-navigator/releases page. Currently we have zero content there.

Argumentation that this does not make everything perfect does not help improving the situation. I propose to activate this now, until we find a replacement that fits all requirements. I prefer a progressive improvement today, than the perfect implementation that never happens to be implemented.

cidrblock:
@ssbarnea Given the flurry of work in the repo recently is this still needed?

Issue #551, opened 2021-08-31
Migrate to `src/`-layout


webknjaz:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

The current repository is structured to have `ansible_navigator/` in the project root. This is considered problematic because it may make testing scenarios run checks against the git checkout contents rather than the package "as installed" on the end-user systems (because, depending on the Python invocation, the current working dir may be prepended to `sys.path` shadowing the imports). It may potentially also mask problems with parts of the content not getting into the distribution packages.

Refs:
- https://blog.ganssle.io/articles/2019/08/test-as-installed.html
- https://hynek.me/articles/testing-packaging/

Issue #557, opened 2021-08-31
[maintenance][packaging] Integrate Packit with an upstream RPM spec


webknjaz:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

I think we should have an upstream RPM spec to better understand the packaging issues that may arise in envs where the users rely on dnf to install the navigator. Especially, in the light of #503.
There's a Red Hat lead project called Packit that simplifies building RPMs for different OSs and they also have a GitHub App that essentially implements a CI for running this in the cloud. Let's consider using it!
I'm currently getting experience integrating it into several projects, so over time, I'll be able to do that here too.

PR #543, opened 2021-08-21
Revert "Drop `{tty}` and `{temp_dir}` substitutions"

webknjaz:
This reverts commit 48601e8df49c5240aad5ee96cf9b9bce7e8a50cb.

Depends-On: https://github.com/ansible/ansible-navigator/pull/571

PR #562, opened 2021-09-01
Adopt src code layout

ssbarnea:
Fixes: #551


ganeshrn (approved):

tox.ini
```diff
@@ -94,7 +94,7 @@ commands =
     -a \
     -n \
     -W --keep-going \
-    -d "{toxinidir}/.tmp/.doctrees" \
+    -d "{toxinidir}/.tox/.doctrees" \
```
webknjaz, line 97: This change is wrong, it's missing `.tmp`. Besides, it's unrelated, so submit it separately or wait for #543 to fix it.

webknjaz (changes requested):

webknjaz (approved):

webknjaz:
FTR we probably need to notify the current RPM spec maintainers about the structural changes. OTOH with #557, it should be easier for them to pick up a well-tested upstream spec.

PR #566, opened 2021-09-02
Fix tox config

ssbarnea:
- include all supported python versions in default invocation of tox (when is called without any arguments)
- remove leading space at end of line
- add a `minversion` pointing to current newest release of tox in order to ensure we do not accidentally run with old versions that may be affected by bugs that can affect the final outcome.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,15 +5,16 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
-minversion = 3.16.1
+envlist = linters, type, py{36,37,38,39,310}, report, clean
+# we do not want debug old bugs
+minversion = 3.24.3
 skipsdist = True
 skip_missing_interpreters = true
 
 
 [testenv]
 description = Run pytest under {basepython} ({envpython})
-# the pytest command line is not in the project.toml because of issues 
+# the pytest command line is not in the project.toml because of issues
 # with catching breakpoints while debugging unit tests
 # the number of CPUs is set to 50% b/c CI timeout failures were more likely
 # the small test image is used an alternative to the default, but should be it


tox.ini
```diff
@@ -115,7 +113,6 @@ passenv =
 skip_install = false
 usedevelop = true
 
-
```
webknjaz, line 118: Let's keep double line separators between different kinds of sections (flake vs tox global vs tox testenvs). They signal the difference clearly.

tox.ini
```diff
@@ -115,7 +113,6 @@ passenv =
 skip_install = false
 usedevelop = true
```
webknjaz, line 115:
```suggestion


```

tox.ini
```diff
@@ -5,15 +5,13 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
+envlist = linters, type, py, report, clean
 minversion = 1.4.2
 skipsdist = True
-skip_missing_interpreters = true
-
```
webknjaz, line 11: Same here, don't break this separator.
```suggestion


```

tox.ini
```diff
@@ -5,15 +5,13 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
+envlist = linters, type, py, report, clean
 minversion = 1.4.2
 skipsdist = True
-skip_missing_interpreters = true
```
webknjaz, line 11: This is good to have in the base config and it should be explicitly disabled in CI, not here.

tox.ini
```diff
@@ -5,15 +5,13 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
+envlist = linters, type, py, report, clean
 minversion = 1.4.2
 skipsdist = True
```
webknjaz, line 10:
```suggestion
skipsdist = True
skip_missing_interpreters = true
```

tox.ini
```diff
@@ -5,15 +5,13 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
+envlist = linters, type, py, report, clean
```
webknjaz, line 8: I'm not sure if this is needed. It was clearly selected to signal what interpreter is currently expected. If we want more, we need to start with the CI first.

webknjaz (changes requested):

tox.ini
ssbarnea: Previous value was incorrect, especially for a project that officially supports multiple versions of python. Listing specific version on default envlist is a mistake. As others said the version agnostic "py" should be used here.

ssbarnea:
recheck

tox.ini
webknjaz: I still don't like this.

webknjaz (changes requested):
I don't really see how this fixes the CI, it seems more like a band-aid. The proper fix is to actually fix the CI, not to make tox pretend that all's good. Plus some of the earlier comments ended up ignored for some reason 🤷‍♂️

tox.ini
ssbarnea: That configuration settings is permanent source of false-positive testing on CI as it will make tox use a different python version than the expected one. Is one of the most common mistakes people make while writing tox.ini. 

This repository was affected by such a bug, where we run py36 for almost an year while telling everyone we test with py38. Similar bugs happened with OpenStack projects and that this why they even went so far to be explicit about disabling this anti-feature. See https://codesearch.opendev.org/?q=skip_missing_interpreters&i=nope&literal=nope&files=&excludeFiles=&repos=

If you do not care about python version, you use `py` env. If you care, you mention it and tox should ensure that it using it, regardless if this happens locally or on CI.

There is another reason why this must be fixed here first: CI system that we can use can change, we may decide to switch from one system to another and almost for sure when doing this we will forget about small details like this. Changing in tox.ini prevents further accidents.

ganeshrn (approved):

tox.ini
webknjaz: @ssbarnea you marked this as resolved but it isn't.

cidrblock:
@ssbarnea given the flurry of activity in the repo, is this still needed?

tox.ini
ssbarnea: I do no agree with on this but I removed that change to unblock the PR.

ssbarnea:
@cidrblock Yes. In fact there are more changes needed but I let them out to ease the review. I also dropped one where I disagree with @webknjaz for the same reason.

tox.ini
```diff
@@ -5,15 +5,16 @@ small_test_ee = quay.io/ansible/python-base
...
+# we do not want debug old bugs
+minversion = 3.24.3
 skipsdist = True
 skip_missing_interpreters = true
 
 
 [testenv]
 description = Run pytest under {basepython} ({envpython})
-# the pytest command line is not in the project.toml because of issues 
+# the pytest command line is not in the project.toml because of issues
```
webknjaz, line 17: This does not seem related.

ssbarnea: I added it to comment, and I hope you do not suggest to make a standalone change to remove that extra space.

webknjaz: I would prefer it to be separate but I won't block. I was reviewing this from the GH Android app and it wasn't even visible what changed here, they don't illuminate those spaces.

tox.ini
```diff
@@ -5,15 +5,16 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
-minversion = 3.16.1
+envlist = linters, type, py{36,37,38,39,310}, report, clean
+# we do not want debug old bugs
+minversion = 3.24.3
```
webknjaz, line 10: I feel like this is a separate thing to change. I don't see why this arbitrary version is chosen but it's not a "fix" as the PR title suggests. If you want to bump it, do so in a PR that separately demonstrates that there are no problems with this bump. OTOH, I would use the first version that drops Python 3.5 support unless there are legitimate tox features (like new substitutions) that are really used in the config. If you want to add those new features, it would be expected that the min version would be set in a respective PR.

tox.ini
webknjaz: I just realized that it also has a typo in "pyproject". This is one of the reasons I prefer separate PR — small mistakes like this will block the entirety of the pull request, including perfectly good changes that could've been merged ages ago.

webknjaz:
> * use version agnostic py

I thought you were adding more envs to the default invocation and not just `tox -e py`. Could you reflect that somehow?

tox.ini
```diff
@@ -5,15 +5,16 @@ small_test_ee = quay.io/ansible/python-base
 mypy_tmp_dir = mypy_tmp_dir
 
 [tox]
-envlist = linters, type, py38, report, clean
-minversion = 3.16.1
+envlist = linters, type, py{36,37,38,39,310}, report, clean
```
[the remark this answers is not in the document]
ssbarnea: Considering that `py38` is not documented in any readme, I would find the documentation request as scope-creeping.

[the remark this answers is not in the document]
webknjaz: PR title is not documentation. The request is not to put nonsense there.

Issue #503, opened 2021-06-25
RPM installed ansible and ansible-navigator, collection cataloging fails when execution-environment=False


yagomarques:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
RPM installed ansible and ansible-navigator, collection cataloging fails when execution-environment=False, because when runner runs the catalog collection script the import of ansible fails.

##### LOG FILE
```
210624202944.120 ERROR 'ansible_navigator.actions.collections._parse' Unable to extract collection json from stdout
210624202944.120 DEBUG 'ansible_navigator.actions.collections._parse' error json loading output: 'not enough values to unpack (expected 2, got 1)'
210624202944.120 DEBUG 'ansible_navigator.actions.collections._parse' Traceback (most recent call last):
  File "/usr/share/ansible_navigator/utils/catalog_collections.py", line 22, in <module>
    from ansible.utils.plugin_docs import get_docstring  # type: ignore
ModuleNotFoundError: No module named 'ansible'
```

##### STEPS TO REPRODUCE
**On RHEL**
1. there is a RHEL machine with ansible-navigator installed with RPM
2. run the command `ansible-navigator collections --ee false --ll debug --la false`
3. check the logs, you will see the error.
##### EXPECTED RESULTS

Show all collections installed locally 

##### ACTUAL RESULTS
Costumer cant see all collections installed locally

##### ADDITIONAL INFORMATION
We should be able to reproduce the same error on Fedora, but the process and the cause is other.
to reproduce:

1. install `pip install ansible` without a virtual environment
2. python3 -m venv venv
3. source venv/bin/activate
4. pip install ansible-navigator
5. `ansible-navigator collections --ee false --ll debug --la false`
6. check the error:
```
210624204127.219 ERROR 'ansible_navigator.actions.collections._parse' Unable to extract collection json from stdout
210624204127.220 DEBUG 'ansible_navigator.actions.collections._parse' error json loading output: 'not enough values to unpack (expected 2, got 1)'
210624204127.220 DEBUG 'ansible_navigator.actions.collections._parse' Traceback (most recent call last):
  File "/home/yagosilva/projects/ansible-navigator/share/ansible_navigator/utils/catalog_collections.py", line 22, in <module>
    from ansible.utils.plugin_docs import get_docstring  # type: ignore
ModuleNotFoundError: No module named 'ansible'
```
the scenario:
```
(navigator)  yagosilva@localhost  ~/projects  which ansible
[...]

PR #567, opened 2021-09-06
Add ansible-related extras to distribution package

### LINKED ISSUE #559: [packaging] Add an extra for requiring `ansible-core`
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

In the light of #503, it could be a nice convenience feature for the pip users.

webknjaz:
Resolves #559

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ A demo of the interface can be found [on youtube][YT demo].
 Getting started with ansible-navigator is as simple as:
 
 ```
-pip install ansible-navigator
+pip install ansible-navigator[ansible-core]
 ansible-navigator --help
 ```
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ A demo of the interface can be found [on youtube][YT demo].
 Getting started with ansible-navigator is as simple as:
 
 ```
-pip install ansible-navigator[ansible-core]
+pip3 install 'ansible-navigator[ansible-core]'
 ansible-navigator --help
 ```
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,6 +43,14 @@ install_requires =
     dataclasses; python_version < "3.7.0"
 python_requires = >=3.6.1
 
+[options.extras_require]
+ansible =
+    ansible
+ansible-base =
+    ansible-core
+ansible-core =
+    ansible-core
+
 [options.entry_points]
 console_scripts =
     ansible-navigator=ansible_navigator.cli:main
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -17,7 +17,7 @@ A demo of the interface can be found [on youtube][YT demo].
 Getting started with ansible-navigator is as simple as:
 
 ```
-pip install ansible-navigator
+pip install ansible-navigator[ansible-core]
```
ssbarnea, line 20: Use `pip3` and add double quotes as use of brackets does not work in all shells, like `pip3 install "ansible-navigator[ansible-core]"` -- safer to copy/paste ;)

webknjaz: Fair enough. Although, I'd use single quotes then, to prevent interpolation if somebody accidentally copy-pastes vars inside...
```suggestion
pip3 install 'ansible-navigator[ansible-core]'
```

setup.cfg
```diff
@@ -43,6 +43,14 @@ install_requires =
     dataclasses; python_version < "3.7.0"
 python_requires = >=3.6.1
 
+[options.extras_require]
+ansible =
+    ansible
+ansible-base =
+    ansible-core
```
webknjaz, line 50: Oh, I missed this too
```suggestion
    ansible-base
```

ssbarnea:
@ganeshrn @priyamsahoo what is your take on this?

* use only "ansible" and "core" extras, same as ansible-lint
* use the currently proposed 3 extras
* don't use any extra and rely on user to know how to install (not really a big deal)

relrod:
This doesn't seem immediately useful to me; what does a user get doing `pip install ansible-navigator[ansible-core]` vs `pip install ansible-navigator ansible-core`?

If there were more (optional) packages that got brought in when doing that, I could use the usecase for it. But as it stands, it just seems to add a character to the install command and not do much else. Am I missing anything?

webknjaz:
@relrod mostly it's a convenience thing. Helps people not to get themselves into situations like #503 where ansible-core is installed but in the wrong place.
Also, think of it as an explicit expression of intent. _"I want ansible-navigator with the support of local ansible-core"_ == `pip install ansible-navigator[ansible-core]`. When somebody lists direct dependencies of their virtualenv, it's good to make sure that the intent behind adding each of those is clear. It's like a feature flag, in a way.
Besides, it'd be aligned with what other projects do.

ganeshrn (approved):

cidrblock (approved):

ganeshrn:
@ssbarnea My preference is `use the currently proposed 3 extras` as it identifies those 3 are different packages.

priyamsahoo:
> @ssbarnea My preference is `use the currently proposed 3 extras` as it identifies those 3 are different packages.

+ 1.

PR #571, opened 2021-09-10
Restore zuul jobs

ssbarnea:
This moves in-tree the tox-py38 job to make it easier for others to
maintain.

It also makes sure that the tests are *actually* running under Python 3.8 as requested.

Closes #565.


.zuul.d/jobs.yaml
```diff
@@ -0,0 +1,29 @@
+---
+- job:
+    name: ansible-navigator-tox-py38
+    parent: ansible-buildset-registry-consumer
+    dependencies:
+      - ansible-buildset-registry
+    timeout: 5400
+    vars:
+      container_command: podman
+      tox_extra_args: -vv --skip-missing-interpreters false
```
webknjaz, line 10: :heart:

ssbarnea: This will be the default in ~2 weeks anyway but why to wait. I can remove it after i merge it in zuul-jobs.

webknjaz (approved):

PR #594, opened 2021-10-11
[runner.api] Let command runner wrap with sh -c

relrod:
Change:
- In some cases, when using the runner API to call into external
  commands, we might wish to retrieve *only* stdout. With runner
  defaults (pexpect), it will force --tty, which will munge the
  containerized command's stdout and stderr all into just stdout.
  (see: https://git.io/JoAAR and https://git.io/JoAxi)
- With this change, we can wrap the containerized command in
  `/bin/sh -c` by passing an argument to the constructor of
  CommandBaseRunner and ignore stderr output.

Test Plan:
- Tested with new, not-yet-pushed :lint action
- New units

### DIFF
--- a/tests/unit/runner/test_api.py
+++ b/tests/unit/runner/test_api.py
@@ -12,22 +12,22 @@ from ansible_navigator.runner import api
         (
             "ansible-lint",
             [],
-            "ansible-lint 2>/dev/null",
+            "exec 2>/dev/null; ansible-lint",
         ),
         (
             "ansible-lint",
             ["/tmp/myroles/supercool/tasks/main.yml"],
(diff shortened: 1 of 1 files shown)


relrod:
@cidrblock for review, maybe there's a better way to do this that doesn't drop stderr entirely. It would be nice to be able to capture it still, but currently the only way I can think of is writing it out to a file and then slurping that which seems super hacky.

The root cause here is that `--tty` is given by default from `ansible-runner` to docker/podman. And when that is given, docker/podman will combine the containerized command's stdout and stderr all into stdout on the host. So in cases where we want to parse the stdout (think `ansible-lint -f codeclimate` JSON stdout), stderr messes up being able to do so.

I figure this might come up again later, so made it general instead of just shoving it in the new `lint` action.

PR #595, opened 2021-10-14
[ui_framework] Several niceties in prep for lint

relrod:
Change:
- Add error_notification (similar to warning_notification, but more red)

- In menu_builder, don't try to render IntEnums as ints (preferring to
  keep them left-aligned)

- Let filter_content_keys take any object, don't force it to be a
  Dict[Any, Any]. As long as it *returns* a dictionary, we can render it
  just fine. This lets actions use custom classes (such as dataclasses)
  to manipulate their internal data, without needing to do as many
  conversions to interact with the rest of the navigator UI system.

Test Plan:
- Most of this is being tested and used in the lint action I'm working
  on; there isn't code that makes use of these changes in the current
  codebase.


relrod:
recheck

PR #596, opened 2021-10-15
[lint] Introduce basic ansible-lint integration

### LINKED ISSUE #531: Add subcommand to run ansible-lint with local and execution-environment support
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Add subcommand to run ansible-lint with local and execution-environment support 
in stdout and interactive mode

relrod:
Fixes https://github.com/ansible/ansible-navigator/issues/531
```
    [lint] Introduce basic ansible-lint integration
    
    Change:
    - Introduces basic ansible-lint integration via `:lint` and
      `ansible-navigator lint`.
    
    - Configuration can be passed in with -c, but must be specified as a
      mount point to ansible-navigator if using execution environments.
    
    - Added to welcome.md as a preview.
    
    Test Plan:
    
    - Lots of local testing. Integration tests have yet to be generated.
    
    Tickets:
    
    - Fixes #531
    
```

```
    Fix up sh -c wrapper from #594
    
    Change:
    - Work around bug in ansible-runner. We need to break this into two
      cases; in interactive mode, arguments get passed pretty much as
      expected. But in stdout (what runner calls "subprocess") mode, we need
      to munge the arguments into one string because arguments aren't
      properly quoted.
    
    Test Plan:
    - Updated units so we can test stdout and interactive arg results.
    
```


relrod:
recheck

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -270,7 +270,7 @@ def execution_environment_volume_mounts(
                     if entry.cli_parameters:
                         exit_msg = (
                             "The value of execution-environment.volume-mounts"
-                            + "should be list of dictionaries"
+                            + " should be list of dictionaries"
```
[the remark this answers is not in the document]
webknjaz: I'd even drop that runtime concat (`+`) and let the interpreter glue these together during the parsing stage.

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,372 @@
...
+        # with just the line number.
+        if isinstance(issue["location"]["lines"]["begin"], collections.Mapping):
+            line = issue["location"]["lines"]["begin"]["line"]
+            column = issue["location"]["lines"]["begin"]["column"]
+        else:
+            line = issue["location"]["lines"]["begin"]
+
+        # The codeclimate formatter combines the rule id and message, so we
+        # parse them out.
+        msg_split = issue["check_name"].split("] ", 1)
```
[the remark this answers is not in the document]
relrod: > can we capitalize the first letter of the message so they are consistent

done

> should we keep the class of the check and stuff it in

Not sure on displaying this yet; I'm still not sure how useful it is. The initial prototype had it but I nuked it pretty early on.

ssbarnea:
@relrod I moved this PR to draft due to its huge number of conflicts. Lets keep it as draft until we have it ready for review.

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+
+        column = None
+        # This will be a dict if we have a line + column, otherwise an int
+        # with just the line number.
+        if isinstance(issue["location"]["lines"]["begin"], collections.Mapping):
+            line = issue["location"]["lines"]["begin"]["line"]
+            column = issue["location"]["lines"]["begin"]["column"]
+        else:
+            line = issue["location"]["lines"]["begin"]
```
webknjaz, line 101: I feel like this can be simplified by reducing copy-paste when accessing nested locations and labeling things:
```suggestion
        beginning_marker = issue["location"]["lines"]["begin"]

        line = beginning_marker
        column = None

        # This will be a dict if we have a line + column, otherwise an int
        # with just the line number.
        is_a_2d_coordinate = isinstance(beginning_marker, collections.Mapping)

        if is_a_2d_coordinate:
            line = beginning_marker["line"]
            column = beginning_marker["column"]
```

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,372 @@
...
+
+        return "", "", rc
+
+    def _run_runner(self) -> Tuple[str, str, int]:
+        """Spin up runner to run ansible-lint, either in an exec env or not."""
+
+        ansible_lint_path = "ansible-lint"
+        if not self._args.execution_environment:
+            ansible_lint_path_maybe = shutil.which("ansible-lint")
+            if ansible_lint_path_maybe is None:
```
webknjaz, line 193: Maybe try adjacent to `sys.executable` too?

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,372 @@
...
+            "set_environment_variable": self._args.set_environment_variable,
+            "private_data_dir": self._args.ansible_runner_artifact_dir,
+            "rotate_artifacts": self._args.ansible_runner_rotate_artifacts_count,
+            "timeout": self._args.ansible_runner_timeout,
+            "host_cwd": os.getcwd(),
+            "wrap_sh": True,
+        }
+
+        if isinstance(self._args.execution_environment_volume_mounts, list):
+            kwargs.update(
```
webknjaz, line 213: What is the motivation for using `dict.update()` for just a single item?

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,372 @@
...
+                {"container_volume_mounts": self._args.execution_environment_volume_mounts}
+            )
+
+        if isinstance(self._args.container_options, list):
+            kwargs.update({"container_options": self._args.container_options})
+
+        cmd_args = []
+
+        if self._args.mode == "interactive":
```
webknjaz, line 222: This condition is repeated in a few places. Could you move it into a `@property`?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -598,7 +609,7 @@ def pass_environment_variable(
     @_post_processor
     def playbook(entry: Entry, config: ApplicationConfiguration) -> PostProcessorReturn:
         # pylint: disable=unused-argument
-        """Post process pass_environment_variable"""
+        """Post process playbook"""
```
webknjaz, line 612:
```suggestion
        """Post-process playbook."""
```

src/ansible_navigator/steps.py
```diff
@@ -19,7 +19,7 @@ def __init__(
         self,
         name: str,
         tipe: str,
-        value: List[Dict[str, str]],
+        value: List[Any],
```
webknjaz, line 22: What is it, then?

tests/unit/runner/test_api.py
```diff
@@ -7,27 +7,36 @@
 
 
 @pytest.mark.parametrize(
-    "orig_exec, orig_args, sh_c_arg",
+    "orig_exec, orig_args, exp_stdout, exp_interactive",
```
webknjaz, line 10: While you're at it, could you list separate args separately, plz?
```suggestion
    ("orig_exec", "orig_args", "exp_stdout", "exp_interactive"),
```

relrod:
recheck

relrod:
recheck

relrod:
recheck

relrod:
recheck

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,343 @@
...
+def massage_issues(issues: List[Dict]) -> List[Dict]:
+    """Massage issues by injecting some useful keys with strings for rendering."""
+    out = []
+    for issue in issues:
+        issue["__message"] = issue["check_name"].split("] ", 1)[1].capitalize()
+        issue["__path"] = abs_user_path(issue["location"]["path"])
+        if isinstance(issue["location"]["lines"]["begin"], Mapping):
+            issue["__line"] = issue["location"]["lines"]["begin"]
+        else:
+            issue["__line"] = issue["location"]["lines"]["begin"]
```
webknjaz, line 130: Both if-branches are exactly the same. It's probably a mistake.

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,343 @@
...
+        issue["issue_path"] = f"{issue['__path']}:{issue['__line']}"
```
webknjaz, line 131: Could you move this set of transformations into a dedicated pure function? It'd be easier to understand it this way.

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,343 @@
...
+        # pylint: disable=invalid-name
+        self._logger.error(msg)
+
+        if self.is_interactive:
+            notification = error_notification(messages=[msg])
+            self._interaction.ui.show(notification)
+        else:
+            raise RuntimeError(msg)
+
+        return "", "", rc
```
webknjaz, line 160: Does this really need to return a tuple? I only saw one place where the return value is even used.
Also, `rc` is passed into this method and never used, just returned back to the caller — why can't the caller just use it?

src/ansible_navigator/actions/lint.py
```diff
@@ -0,0 +1,343 @@
+"""
+This action provides ansible-lint results through ansible-navigator.
+Internally, it works by using ansible-runner to execute ansible-lint (optionally
+in an execution environment). When running ansible-lint, it passes
+'-f codeclimate' which requests JSON output on stdout. The JSON output contains
```
webknjaz, line 5: I think highlighting this would render more nicely:
```suggestion
``-f codeclimate`` which requests JSON output on stdout. The JSON output contains
```
See https://ansible-navigator--596.org.readthedocs.build/en/596/pkg/ansible_navigator.actions.lint/#module-ansible_navigator.actions.lint

src/ansible_navigator/actions/lint.py
relrod: Heh, a former version of this code used to work a bit differently and this mistake fell out of a refactor. First branch should be `issue["location"]["lines"]["begin"]["line"]`, good catch.

cidrblock:
In the case of no errrors `out = []`, think we should show a notification saying `Congratulations no errors were found in `some_path` rather than bringing up an empty menu?

cidrblock:
Can we move the welcome screen change to it's own PR?   This is the approach `:settings` is taking.....

(just to drop some of the fixture noise)

cidrblock (approved):

PR #614, opened 2021-11-17
Replace use of yum w/ rpm

### LINKED ISSUE #613: `:images`  System package collection broken b/c `yum` not in EE
https://github.com/ansible/ansible-navigator/blob/56ea2ee934419005254e9e4320febc8ee4b07b6c/share/ansible_navigator/utils/image_introspect.py#L259

cidrblock:
Fixes #613 

EEs (downstream) no longer have yum installed, so switch to RPM for package information.

The shared parser could not easily be adapted to the way RPM formats it's output, so parse the RPM output in a dedicated parser

as always, the real changes are in the image introspection script, the majority of the fixture updates are a back log from before the default EE was changed.


share/ansible_navigator/utils/image_introspect.py
```diff
@@ -115,6 +115,11 @@ def start_workers(self, jobs):
 
 
 class CmdParser:
+    @staticmethod
+    def _strip(value: str) -> str:
```
webknjaz, line 119: This is a pure function, it shouldn't belong to a class.

share/ansible_navigator/utils/image_introspect.py
```diff
@@ -256,11 +256,44 @@ class SystemPackages(CmdParser):
     @property
     def commands(self) -> List[Command]:
         """generate the command"""
-        return [Command(id="system_packages", command="yum info installed", parse=self.parse)]
+        return [Command(id="system_packages", command="rpm -qai", parse=self.parse)]
```
webknjaz, line 259: Why not dnf?

share/ansible_navigator/utils/image_introspect.py
cidrblock: It's use is limited and specific to the class in which it is contained, and not intended to be used outside this class nor is it.

share/ansible_navigator/utils/image_introspect.py
cidrblock:
```
bash: dnf: command not found
```

cidrblock: dnf is not available in the EE

share/ansible_navigator/utils/image_introspect.py
webknjaz: It's just that it's an antipattern in Python. It comes from Java where you cannot do things differently. Plus the name `_strip()` is not semantically connected to this class.

Imagine you write a unit test for this method (which you probably should). You'll call `CmdParser._strip('   as\'d"f ')` — now this looks like this has something to do with the parser or the command.

This is why it's recommended to at least put this helper outside of the class.

PR #619, opened 2021-11-18
[POC] Redo #614 with rpm Python lib

relrod:
Change:
- This is a proof of concept to rewrite the change in #614 but
  using the Python rpm binding instead. In addition to fixing various
  issues with parsing `rpm -qi` output, this also gives us access to
  more information to display.
- This only serves as a demo - there is some work to be done before
  this could actually replace 614. See TODOs in this patch.

Test Plan:
- Tested locally a bunch.


relrod:
Closing, apparently the supported EE images don't have python3-rpm.

PR #621, opened 2021-11-18
`:images` fix for `Name` in `Description`

cidrblock:
@relrod noticed that if the Description of a package has a line starting with Name, it would break the package splitter

This specifies that the package delimiter is Name followed by spaces and a :, which is not perfect but better

Small unit test added to demonstrate the fix, which also caught an issue only one & the final package missing

Follow up on #614

### DIFF
--- a/tests/unit/test_image_introspection.py
+++ b/tests/unit/test_image_introspection.py
@@ -78,6 +78,8 @@ def test_system_packages_parse_many(imported_ii):
     assert len(command.details) == count
     for entry in command.details:
         assert entry["name"] == "net-snmp"
+        assert entry["version"] == "5.9.1"
+        assert entry["summary"] == "A collection of SNMP protocol tools and libraries"
         assert entry["description"].startswith("SNMP")
         assert entry["description"].endswith("utilities.")
         assert "summary: summary_string" in entry["description"]
--- a/tests/unit/test_image_introspection.py
+++ b/tests/unit/test_image_introspection.py
@@ -1,0 +1,75 @@
+""" unit tests for image introspection
+"""
+import importlib
+import pytest
+import ansible_navigator.utils as utils
+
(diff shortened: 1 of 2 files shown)


tests/unit/test_image_introspection.py
```diff
@@ -0,0 +1,82 @@
...
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_system_packages_parse_one(imported_ii):
+    """test parsing one package"""
+    command = imported_ii.Command(id="test", parse=lambda x: x, stdout=RPM_OUTPUT)
+    imported_ii.SystemPackages().parse(command)
+    assert len(command.details) == 1
+    assert command.details[0]["name"] == "net-snmp"
```
relrod, line 63:
```suggestion
    assert command.details[0]["name"] == "net-snmp"
    assert command.details[0]["version"] == "5.9.1"
    assert command.details[0]["summary"] == "A collection of SNMP protocol tools and libraries"
```

tests/unit/test_image_introspection.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_system_packages_parse_many(imported_ii):
+    """test parsing many packages"""
+    count = 10
+
+    command = imported_ii.Command(id="test", parse=lambda x: x, stdout=RPM_OUTPUT * count)
+    imported_ii.SystemPackages().parse(command)
+    assert len(command.details) == count
+    for entry in command.details:
+        assert entry["name"] == "net-snmp"
```
relrod, line 78:
```suggestion
        assert entry["name"] == "net-snmp"
        assert entry["version"] == "5.9.1"
        assert entry["summary"] == "A collection of SNMP protocol tools and libraries"
```

relrod (changes requested):

cidrblock:
recheck

relrod (approved):

Issue #299, opened 2021-05-19
[runner] Organize runner directory


cidrblock:
Break classes into individual file
List known and only required entry point in runner/__init__.py
Change import statements as required

webknjaz:
@cidrblock do you know if it's been solved already?

Issue #497, opened 2021-06-24
runner artifact cleanup does not always run


cidrblock:
https://github.com/ansible/ansible-navigator/blob/c2167350fb202aa80427bf352f41eed58e48b232/ansible_navigator/runner/api.py#L149

found while working on https://github.com/ansible/ansible-navigator/pull/496

eamigo:
@cidrblock I think this issue may have been accidentally closed early.  You  mentioned here that you wanted this to remain open. https://github.com/ansible/ansible-navigator/pull/1438#issuecomment-1479708160

PR #623, opened 2021-11-19
Initial cleanup of runner directory

cidrblock:
- Break the single runner file into one filer per class
- Rename classes to parallel the underlying ansible command where applicable
- Drop Runner from all classes since these are contained withing the runner api subsystem
- define public entry points in the runner init file
- update actions as needed to reflect changes within the runner directory

Related: #299 

Followup work as time permits will be to define a series of classes to contain the configuration for each runner interaction


ssbarnea (commented):
Does this count as a `bug`, `improvement` or `skip-changelog`?

By looking at the change, I think it should not have any implications for the end-user, more of a refactoring.

cidrblock:
reply to: https://github.com/ansible/ansible-navigator/pull/623#pullrequestreview-811219951

Yes, refactoring. TY for the reminder to add the skip-changelog label

ssbarnea (approved):
LGTM but ideally someone will more knowledge of navigator should also check it.

src/ansible_navigator/runner/command_async.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class CommandAsync(CommandBase):
+    # pylint: disable=too-few-public-methods
+    # pylint: disable=too-many-instance-attributes
+    """a runner async wrapper"""
+
+    def __init__(self, executable_cmd: str, queue: Queue, **kwargs):
+        """class to handle arguments of ``run_command_async`` interface for ``ansible-runner``.
+           For common arguments refer documentation of ``CommandBaseRunner`` class
```
webknjaz, line 20: This docstring is in violation of PEP 257 and it also provides invalid information. First of all, it's not a class but an initializer. Then, a docstring title must be a full sentence (with a capital letter at the beginning and a trailing period at the end). The first line should fit in 72 chars and be separated from prose with an empty line.
```suggestion
        """Initialize the arguments for the ``run_command_async`` interface of ``ansible-runner``.

        For common arguments refer to the documentation of the ``CommandBaseRunner`` class.
```

src/ansible_navigator/runner/command_base.py
```diff
@@ -0,0 +1,74 @@
...
+            self._cmdline.insert(0, self._playbook)
+
+        for inv in self._inventory:
+            self._cmdline.extend(["-i", inv])
+
+        if self._wrap_sh:
+            self._cmdline.insert(0, self._executable_cmd)
+            self._runner_args["executable_cmd"] = "/bin/sh"
+            cmd_args = " ".join(shlex.quote(s) for s in self._cmdline)
+            self._runner_args["cmdline_args"] = ["-c", "exec 2>/dev/null; {0}".format(cmd_args)]
```
webknjaz, line 63: You could use an f-string here. But if you don't, make use of named placeholders to make this `0` readable.

src/ansible_navigator/runner/command_base.py
```diff
@@ -0,0 +1,74 @@
...
+
+from .base import Base
+
+
+class CommandBase(Base):
+    # pylint: disable=too-few-public-methods
+    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-instance-attributes
+
+    """base class for runner command interaction"""
```
webknjaz, line 19:
```suggestion
    """Base class for runner command interaction."""
```

src/ansible_navigator/runner/command_base.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        self._executable_cmd = executable_cmd
+        self._cmdline: List[str] = cmdline if isinstance(cmdline, list) else []
+        self._playbook = playbook
+        self._inventory: List[str] = inventory if isinstance(inventory, list) else []
+        self._wrap_sh = wrap_sh
+        super().__init__(**kwargs)
+
+    def generate_run_command_args(self) -> None:
+        """generate arguments required to be passed to ansible-runner"""
```
webknjaz, line 52:
```suggestion
        """Generate arguments required to be passed to ansible-runner."""
```

src/ansible_navigator/runner/command_async.py
```diff
@@ -0,0 +1,40 @@
...
+
+        Args:
+            executable_cmd ([str]): The command to be invoked.
+            queue ([Queue]): The queue to post events from ``ansible-runner``
```
webknjaz, line 24: Look here https://ansible-navigator--623.org.readthedocs.build/en/623/pkg/ansible_navigator.runner.command_async/ — these descriptions are not rendered. Let's use the built-in Sphinx argument description style so that it is.
And also, the types are not really necessary because they are already present in the annotations and there's no need to duplicate/keep in sync manually this info.
```suggestion
        :param executable_cmd: The command to be invoked.
        :param queue: The queue to post events from ``ansible-runner``
```

src/ansible_navigator/runner/command.py
```diff
@@ -0,0 +1,21 @@
...
+    # pylint: disable=too-few-public-methods
+    # pylint: disable=too-many-instance-attributes
+    """a runner wrapper"""
+
+    def run(self) -> Tuple[str, str, int]:
+        """run"""
+
+        self.generate_run_command_args()
+        out, err, ret_code = run_command(**self._runner_args)
+        return out, err, ret_code
```
webknjaz, line 21:
```suggestion
        return run_command(**self._runner_args)
```

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Args:
+            runner (Runner): a runner instance
+        """
+        self.status = runner.status
+        self.finished = True
+
+    def _add_env_vars_to_args(self):
+        self._runner_args["envvars"] = {
+            k: v for k, v in os.environ.items() if k.startswith("ANSIBLE_")
```
webknjaz, line 207: A lot of things are happening on this line that affects the cognitive complexity level. You can put the condition on a separate line to make it more readable:
```suggestion
            k: v for k, v in os.environ.items()
            if k.startswith("ANSIBLE_")
```

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+        }
+        self._runner_args["envvars"].update(self._set_environment_variable)
+        for env_var in self._pass_environment_variable:
+            value = os.environ.get(env_var)
+            if value is None:
```
webknjaz, line 212: The semantics of `dict.get()` is different from what the code implies here. "`None`" doesn't mean "unset", it is merely just a default returned value.
With this code, you assume that value in that dict can never be set to `None` which is not really true.
So basically, this check represents both "a dict value is unset" and "a dict value is set to exactly `None`".

This is one of the reasons it's common to use EAFP in Python.

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+                self._logger.warning(
+                    "Pass through environment variable `%s`" " not currently set, discarded",
+                    env_var,
+                )
+            else:
+                self._runner_args["envvars"][env_var] = value
```
webknjaz, line 218: A more semantically correct version of this would be
```suggestion
            try:
                env_var_value = os.environ[env_var]
            except KeyError:
                self._logger.warning(
                    "Pass through environment variable `%s`" " not currently set, discarded",
                    env_var,
                )
            else:
                self._runner_args["envvars"][env_var] = env_var_value
```

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+        if self._navigator_mode == "stdout":
+            self._runner_args.update(
+                {"input_fd": sys.stdin, "output_fd": sys.stdout, "error_fd": sys.stderr}
+            )
+
+    def __del__(self):
+        """
+        class destructor, handle runner artifact file deletion is rotate_artifacts \
+        value is not set
+        """
```
webknjaz, line 146: The name `__del__` already implies that it's a destructor. It's an instance destructor, though, not a class destructor. Anyway, it's not worth mentioning. Plus the docstring could be more precise and PEP 257 compliant.
```suggestion
        """Drop the artifact directory when the rotation is disabled."""
```

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+            # ensure the cwd ends with a trailing slash
+            host_cwd = os.path.join(self._host_cwd, "")
+            self._runner_args.update({"host_cwd": host_cwd})
+
+        if self._navigator_mode == "stdout":
+            self._runner_args.update(
+                {"input_fd": sys.stdin, "output_fd": sys.stdout, "error_fd": sys.stderr}
+            )
+
+    def __del__(self):
```
webknjaz, line 142: Note that this method is not guaranteed to be called in a predictable manner. For example, Python implementations with different GC implementations (without a reference counter), the call may be significantly delayed in time (like under PyPy).

webknjaz: This is why it's best to have a context manager to perform this in a controlled/predictable manner.

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+            and self._runner_artifact_dir
+            and os.path.exists(self._runner_artifact_dir)
+        ):
+
+            self._logger.debug(
+                "delete ansible-runner artifact directory at path %s", self._runner_artifact_dir
+            )
+            shutil.rmtree(self._runner_artifact_dir, ignore_errors=True)
+
+    @staticmethod
```
webknjaz, line 158: Unless you expect this to be reused by subclassed, move it to the module-global level. In Python, it's an antipattern to use classes for namespacing pure functions.

src/ansible_navigator/runner/base.py
```diff
@@ -0,0 +1,218 @@
...
+                "json_mode": True,
+                "quiet": True,
+                "cancel_callback": self.runner_cancelled_callback,
+                "finished_callback": self.runner_finished_callback,
+                "artifacts_handler": self.runner_artifacts_handler,
+                "timeout": self._timeout,
+            }
+        )
+        if self._rotate_artifacts is not None:
+            self._runner_args.update({"rotate_artifacts": self._rotate_artifacts})
```
webknjaz, line 128: No need to overcomplicate a single assignment here.
```suggestion
            self._runner_args["rotate_artifacts"] = self._rotate_artifacts
```

src/ansible_navigator/runner/ansible_doc.py
```diff
@@ -0,0 +1,70 @@
...
+            module_path (Optional[str], optional): This parameter is prepend colon-separated \
+                                                   path(s) to module library \
+                                                   (default=~/.ansible/plugins/modules: \
+                                                   /usr/share/ansible/plugins/modules).
+
+        Returns:
+            Tuple[Union[Dict[Any, Any], str], Union[Dict[Any, Any], str]]: Returns a tuple of \
+                                                    response and error string. If the value of \
+                                                    ``response_format`` is ``json`` it returns \
+                                                    a python dictionary object.
```
webknjaz, line 59: This is misinterpreted by Sphinx as you can see here: https://ansible-navigator--623.org.readthedocs.build/en/623/pkg/ansible_navigator.runner.ansible_doc/#ansible_navigator.runner.ansible_doc.AnsibleDoc.fetch_plugin_doc.
Use the built-in param list feature as documented here: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists.

src/ansible_navigator/runner/ansible_doc.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+from ansible_runner import get_plugin_docs
+
+from .base import Base
+
+
+class AnsibleDoc(Base):
+    # pylint: disable=too-many-arguments
+    """abstraction for ansible-doc command-line"""
```
webknjaz, line 19: What kind of abstraction? Could you make it a full sentence?

src/ansible_navigator/runner/ansible_doc.py
```diff
@@ -0,0 +1,70 @@
...
+    def fetch_plugin_doc(
+        self,
+        plugin_names: List,
+        plugin_type: Optional[str] = None,
+        response_format: Optional[str] = "json",
+        snippet: Optional[bool] = None,
+        playbook_dir: Optional[str] = None,
+        module_path: Optional[str] = None,
+    ) -> Tuple[Union[Dict[Any, Any], str], Union[Dict[Any, Any], str]]:
+        """Run ansible-doc command and get the plugin docs related details
```
webknjaz, line 30:
```suggestion
        """Run ``ansible-doc`` command and get the plugin docs related details.
```

src/ansible_navigator/runner/__init__.py
```diff
@@ -0,0 +1,8 @@
+""" Defined public entrypoints for the runner api
+"""
```
webknjaz, line 2: Modules are static, their docstrings should not define actions but describe what the thing is. In this case, it's the whole package docstring and it should describe what is `ansible_navigator.runner`.
```suggestion
"""Public entry points for the runner API."""
```

src/ansible_navigator/runner/__init__.py
```diff
@@ -0,0 +1,8 @@
...
+
+from .ansible_config import AnsibleConfig
+from .ansible_doc import AnsibleDoc
+from .ansible_inventory import AnsibleInventory
+from .command import Command
+from .command_async import CommandAsync
```
webknjaz, line 8: It's best to avoid using `__init__.py` as a proxy importer. This may lead to import order dependencies and circular imports in some cases. My favorite linter even has a rule forbidding this.
The best thing you can do is to move these to a module called `api` and then the users would for `from ansible_navigator.runner.api import AnsibleConfig`. The reason is that when there's imports of the nested submodules, the init will also get evaluated causing extra imports when somebody does things like `from ansible_navigator.runner.ansible_doc import AnsibleDoc`, it'll case side effects and evaluate a bunch of unrelated modules making it less predictable.

src/ansible_navigator/runner/command_async.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/command_base.py
cidrblock: fixed in #630

src/ansible_navigator/runner/command_base.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/command_base.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/command_async.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/command_async.py
```diff
@@ -0,0 +1,40 @@
...
+from queue import Queue
+from ansible_runner import run_command_async
+
+from .command_base import CommandBase
+
+
+class CommandAsync(CommandBase):
+    # pylint: disable=too-few-public-methods
+    # pylint: disable=too-many-instance-attributes
+    """a runner async wrapper"""
```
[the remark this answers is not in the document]
cidrblock: Fixed in #630

src/ansible_navigator/runner/command_async.py
```diff
@@ -0,0 +1,40 @@
+""" Herewith lies the ability to invoke ansible-runner
+in an async fashion. A queue is provided and ansible-runner
+uses pexpect to parse standout and error from the command run
+and populates the queue with messages
+"""
```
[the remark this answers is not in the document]
cidrblock: Fixed in #630

src/ansible_navigator/runner/command.py
cidrblock: I prefer it how it is b/c it's easier to put in a break point and introspect the values

src/ansible_navigator/runner/base.py
cidrblock: I think it reads fine as is.

src/ansible_navigator/runner/base.py
cidrblock: This returns a string, None if it doesn't exist b/c a default was not provided:

https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/os.py#L773

src/ansible_navigator/runner/base.py
cidrblock: I agree this a viable alternative, although the current implementation is fine and this falls outside the scope of the PR.

LMK if you want to PR if it is something you would like to change,

src/ansible_navigator/runner/base.py
cidrblock: duplicate of: https://github.com/ansible/ansible-navigator/issues/497

src/ansible_navigator/runner/base.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/base.py
cidrblock: agree, fixed in #630

src/ansible_navigator/runner/ansible_doc.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/__init__.py
cidrblock: Fixed in #630

src/ansible_navigator/runner/base.py
cidrblock: The use of this is limited to the class, hence the _ name and inclusion within the class itself.

Issue #526, opened 2021-08-03
Allow for interactive development within the container


ericzolf:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Developing automation is very incremental so it's a lot of editing, trying, editing, trying, especially at the beginning. The current ansible-navigator approach stays a bit in the way of this approach, with the overhead of the TUI and each time starting a new container. A mode by which the developer could just be put inside the container as themself and could there edit their code and/or start `ansible-playbook` themselves would help a lot.

Basically, it would be the equivalent of `podman run [--rm] --tty --interactive -v /home/myuser/github/pull-test/:/home/myuser/github/pull-test/ --workdir /home/myuser/github/pull-test -v /run/user/1000/keyring/:/run/user/1000/keyring/ -e SSH_AUTH_SOCK=/run/user/1000/keyring/ssh -v /home/myuser/.ssh/:/home/runner/.ssh/ --group-add=root --userns=keep-id  --user="$(id -u):$(id -g)" --ipc=host -v /tmp/ansible-navigator_th4mlx94/artifacts/:/runner/artifacts/:Z -v /tmp/ansible-navigator_th4mlx94/:/runner/:Z --env-file /tmp/ansible-navigator_th4mlx94/artifacts/d8722364-68c8-4875-973e-8d1234008533/env.list --quiet --name ansible_runner_d8722364-68c8-4875-973e-8d1234008533 registry.blup/bla/ansible-automation-platform-20-ee-supported-rhel8:latest /bin/bash`.

Two comments:
1. notice the addition of `--userns=keep-id  --user="$(id -u):$(id -g)"` to make sure that the user is themself within the container and the access right remain correct.
2. being able to keep the container across calls i.e. without `--rm` would also help, for example with the ability to give a fix name to the container, as the developer could then customize it with aspects irrelevant to the automation itself (i.e. add their favourite editor) without having to create an own image.

Thinking about it, that also means that the temporary directory `/tmp/ansible-navigator_th4mlx94` would be persistent, the location `/tmp` isn't a perfect place for this kind of approach, so it should also be moved, e.g. to a sub-directory of the current directory.

anshulbehl:
Few good to have things in future with navigator would also be to have
- An easy way to run setup module or ad-hoc modules like ping
- Like `ansible all -m setup` or `ansible all -m ping`

[...]

PR #624, opened 2021-11-19
`exec` subcommand

cidrblock:
[Demo](https://asciinema.org/a/2Gw0quCXW1FuZUrBdRCODATBS)

- Add exec action
- Add exec subcommand
- Add exec_command and exec_shell settings entries
- Add support for actions that may be stdout only
- Update test data as needed
- Update tests to shlex split for first use of space delimited parameter
- add flake8 file to set docstring format to sphinx
- one black fix in utils, ergh

Fixes: #526


cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

relrod (commented):
Few minor comments.

src/ansible_navigator/app.py
```diff
@@ -77,6 +79,19 @@ def app(self) -> AppPublic:
             )
         raise AttributeError("app passed without args initialized")
 
+    def no_interactive_mode(self, interaction: Interaction, app: AppPublic) -> None:
+        # pylint: disable=unused-argument
+        """Warm the user interactive mode is not supported"""
```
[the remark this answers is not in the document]
cidrblock: ha, thx

src/ansible_navigator/actions/_actions.py
```diff
@@ -94,7 +97,12 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
     """Call the given action's run"""
     action_cls = get(package, action)
     app, interaction = args
-    return action_cls(app.args).run(app=app, interaction=interaction)
+    if hasattr(action_cls(app.args), "run"):
+        return action_cls(app.args).run(app=app, interaction=interaction)
+    logger.error("Subcommand '%s' does not support mode interactive", action)
+    if hasattr(action_cls(app.args), "no_interactive_mode"):
```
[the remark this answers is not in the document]
cidrblock: yeah, good catch

src/ansible_navigator/actions/exec.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self, args):
+        super().__init__(args=args, logger_name=__name__, name="exec")
+
+    @staticmethod
+    def _generate_command(exec_command: str, exec_shell: bool) -> Tuple:
+        """Generate the command and args"""
+        pass_through_args = None
+        if exec_shell and exec_command:
+            command = "/bin/bash"
```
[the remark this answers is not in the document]
cidrblock: I think I might have had it `sh` at a point, but then decided it should be the same as the default command.(`bash`), so the behavior would be the same at the shell as it would with a command passed through.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -116,6 +116,16 @@ class Internals(SimpleNamespace):
             " 'ansible-navigator doc --help-doc --mode stdout'"
         ),
     ),
+    SubCommand(
+        name="exec",
+        description="Run a command within an execution environment",
+        epilog=(
+            "Note: During development, it may become necessary to interact"
+            " directly with the execution environment to review and confirm"
+            " it's build and behavior. All navigator settings will be applied"
```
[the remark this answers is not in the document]
cidrblock: good catch, thx

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -246,6 +256,26 @@ class Internals(SimpleNamespace):
             short_description="Specify if the editor is console based",
             value=EntryValue(default=True),
         ),
+        Entry(
+            name="exec_command",
+            cli_parameters=CliParameters(short="--excmd"),
```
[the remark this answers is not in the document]
cidrblock: because exec IMO is more of a debugging tool, not a primary subcommand, I didn't want to use `command`, we can only use it once and I'm not sure if we might need it for something more important in the future

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -246,6 +256,26 @@ class Internals(SimpleNamespace):
...
+            settings_file_path_override="exec.command",
+            short_description="Specify the command to run within the execution environment",
+            subcommands=["exec"],
+            value=EntryValue(default="/bin/bash"),
+        ),
+        Entry(
+            name="exec_shell",
+            choices=[True, False],
+            cli_parameters=CliParameters(short="--exshell"),
```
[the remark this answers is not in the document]
cidrblock: The configuration subsystem doesn't really support simple flags without a value, the reason is because of the way we generate an environment variable and settings file entry.   There would be some inconsistency between the CLI, env var and settings file.  Good or bad, the way it works today is all three are a bool and require true/false.

I prefixed it with `exec_` to match `exec_command`, let me know if you feel strongly you feel about this one, I can change if you do.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -246,6 +256,26 @@ class Internals(SimpleNamespace):
...
+            settings_file_path_override="exec.shell",
+            short_description=(
+                "Specify the exec command should be run in a shell."
+                " ('/bin/bash -c \"{exec_command}\"')"
```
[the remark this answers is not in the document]
cidrblock: ok

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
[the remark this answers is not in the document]
cidrblock: @relrod thoughts here?

cidrblock:
Recheck

cidrblock:
recheck

cidrblock:
recheck

tests/integration/actions/exec/test_stdout_config_cli.py
```diff
@@ -0,0 +1,79 @@
...
+    ),
+)
+
+steps = add_indicies(stdout_tests)
+
+
+def step_id(value: ShellCommand) -> str:
+    """Return the test id from the test step object.
+
+    :param value: The data fro the test iteration
```
alisonlhart, line 69:
```suggestion
    :param value: The data from the test iteration
```

tests/integration/actions/exec/test_stdout_config_file.py
```diff
@@ -0,0 +1,65 @@
...
+    ),
+)
+
+steps = add_indicies(stdout_tests)
+
+
+def step_id(value: ShellCommand) -> str:
+    """Return the test id from the test step object.
+
+    :param value: The data fro the test iteration
```
alisonlhart, line 54:
```suggestion
    :param value: The data from the test iteration
```

alisonlhart (commented):
Only a couple changes, just edits to comments.

cidrblock:
recheck

cidrblock:
recheck

cidrblock:
recheck

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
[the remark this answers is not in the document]
cidrblock: @relrod ping!

docs/changelog-fragments.d/526.feature.md
```diff
@@ -0,0 +1,2 @@
+Add ability to run commands directly within an execution environment.
```
webknjaz, line 1: Per https://github.com/ansible/ansible-navigator/pull/641#discussion_r765392465, something like this may be better:
```suggestion
Added the ability to run arbitrary shell commands from the TUI directly
so they are invoked within an execution environment.
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -95,7 +98,10 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
     """Call the given action's run"""
     action_cls = get(package, action)
     app, interaction = args
-    return action_cls(app.args).run(app=app, interaction=interaction)
+    if hasattr(action_cls(app.args), "run"):
```
webknjaz, line 101: This looks like a lot of copy-paste and repeat initialization of an in-memory object.

src/ansible_navigator/actions/_actions.py
```diff
@@ -95,7 +98,10 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
+        return action_cls(app.args).run(app=app, interaction=interaction)
+    logger.error("Subcommand '%s' does not support mode interactive", action)
+    return action_cls(app.args).no_interactive_mode(app=app, interaction=interaction)
```
webknjaz, line 104: This can be made more flat:
```suggestion
    app_action = action_cls(app.args)
    supports_interactive = hasattr(app_action, "run")
    if not supports_interactive:
        logger.error("Subcommand '%s' does not support mode interactive", action)
    run_action = app_action.run if supports_interactive else app_action.no_interactive_mode
    return run_action(app=app, interaction=interaction)
```
Plus this verbosity allows you to assign semantical meanings to things using var names.

src/ansible_navigator/actions/exec.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _run_runner(self) -> Optional[Tuple]:
+        # pylint: disable=too-many-branches
+        # pylint: disable=too-many-statements
+        """Spin up runner.
+
+        :return: The stdout, stderr and return code from runner
+        """
+        if isinstance(self._args.set_environment_variable, dict):
+            set_envvars = {**self._args.set_environment_variable}
```
webknjaz, line 70: I think there's no point in this since it's already a dict, right?
```suggestion
            set_envvars = self._args.set_environment_variable
```

src/ansible_navigator/actions/exec.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            set_envvars = {}
```
webknjaz, line 72: Or did you mean this?
```suggestion
        set_envvars = (self._args.set_environment_variable or {}).copy() 
```

webknjaz: Extra note on the var name — `set_envvars` implies an action, if I were to see it used somewhere in the code, I would assume that it needs to be called as `set_envvars()`. If it's supposed to hold data, it should use a noun and have an implication of being an object/a concept, not an action.

src/ansible_navigator/app.py
```diff
@@ -20,6 +20,8 @@
 
 from .ui_framework import Interaction
 from .ui_framework import ui
+from .ui_framework import warning_notification
+
```
webknjaz, line 24: It seems like you're adding this extra indentation by accident:
```suggestion
```

src/ansible_navigator/app.py
```diff
@@ -83,6 +85,19 @@ def app(self) -> AppPublic:
             )
         raise AttributeError("app passed without args initialized")
 
+    def no_interactive_mode(self, interaction: Interaction, app: AppPublic) -> None:
+        # pylint: disable=unused-argument
+        """Warn the user interactive mode is not supported"""
```
webknjaz, line 90: Would this be more precise?
```suggestion
        """Show a warning notification that the user interactive mode is not supported."""
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -310,6 +320,18 @@ def container_options(entry: Entry, config: ApplicationConfiguration) -> PostPro
             entry.value.current = flatten_list(entry.value.current)
         return messages, exit_messages
 
+    @_post_processor
+    def exec_shell(self, entry: Entry, config: ApplicationConfiguration) -> PostProcessorReturn:
+        # pylint: disable=unused-argument
+        """Post process exec_shell.
```
webknjaz, line 326: What is this `exec_shell` that you're post-processing here?
Also, non-code things will potentially be checked for spelling and since it's an identifier, it's best to at least mark it as such:
```suggestion
        """Post process ``exec_shell``.
```

tests/integration/actions/exec/__init__.py
```diff
@@ -0,0 +1 @@
+"""Integration tests for the exec subcommand."""
```
webknjaz, line 1: Since this is rendered as RST:
```suggestion
"""Integration tests for the ``exec`` subcommand."""
```

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,104 @@
...
+        }
+        if isinstance(self.CONFIG_FILE, str):
+            assert os.path.exists(self.CONFIG_FILE)
+            params["config_path"] = self.CONFIG_FILE
+
+        with TmuxSession(**params) as tmux_session:
+            yield tmux_session
+
+    def test(self, request: pytest.FixtureRequest, tmux_session: TmuxSession, step: Step):
+        # pylint:disable=unused-argument
```
webknjaz, line 53:
```suggestion
        # pylint: disable=unused-argument
```

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,104 @@
...
+            assert not any(look_not in page for look_not in step.look_nots)
+
+        if not any((step.look_fors, step.look_nots)):
+            dir_path, file_name = fixture_path_from_request(request, step.step_index)
+            with open(f"{dir_path}/{file_name}") as infile:
+                expected_output = json.load(infile)["output"]
+
+            assert expected_output == received_output, "\n" + "\n".join(
+                difflib.unified_diff(expected_output, received_output, "expected", "received")
+            )
```
webknjaz, line 104: This is a nice opportunity to dedent a code block:
```suggestion
        if any((step.look_fors, step.look_nots)):
            return

        dir_path, file_name = fixture_path_from_request(request, step.step_index)
        with open(f"{dir_path}/{file_name}") as infile:
            expected_output = json.load(infile)["output"]

        assert expected_output == received_output, "\n" + "\n".join(
            difflib.unified_diff(expected_output, received_output, "expected", "received")
        )
```

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,104 @@
...
+
+from ..._common import fixture_path_from_request
+from ..._common import update_fixtures
+from ..._tmux_session import TmuxSession
+from ..._interactions import SearchFor
+from ..._interactions import Step
+from ....defaults import FIXTURES_DIR
+
+TEST_FIXTURE_DIR = os.path.join(FIXTURES_DIR, "integration", "actions", "exec")
+TEST_CONFIG_FILE = os.path.join(TEST_FIXTURE_DIR, "ansible-navigator.yaml")
```
webknjaz, line 20: This is probably out of the scope, but it'd be truly great to use `pathlib` for any interactions with file paths and only convert them to `str` where really needed.

src/ansible_navigator/actions/_actions.py
cidrblock: good catch, yeah, shouldn't need to instantiate it in the conditional, or I guess I could ahead of time

src/ansible_navigator/actions/exec.py
cidrblock: this is what is great about someone else seeing, that was a cut and paste from another action without looking at it twice. yeah, because we manipulate them a copy is cool

src/ansible_navigator/actions/exec.py
cidrblock: the parameter in the settings is 'set_environment_variables` so that got carried around up until the point it's handed to runner. So oddly in this case, the name hooks back to the settings entry and wasn't intended to suggest a function name, but I agree with as a pattern.  I'll make a change here, but that won't eliminate it everywhere

cidrblock: `self._args.set_environment_variable` can either be a Constant or a dict, I prefer doing a type check here instead, since it's not apparent that the check is for that, let me through something else in there and let me know what you think.

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,104 @@
...
+
+        received_output = tmux_session.interaction(
+            value=step.user_input,
+            search_within_response=search_within_response,
+        )
+
+        if (
+            self.UPDATE_FIXTURES
+            or os.environ.get("ANSIBLE_NAVIGATOR_UPDATE_TEST_FIXTURES") == "true"
+        ):
```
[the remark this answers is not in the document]
cidrblock: I agree with you.  This snippet of code can be found in quite a few tests so I'll add it here, but need to do a PR to make changes across the board so it remains consistent.

tests/integration/actions/exec/base.py
cidrblock: interesting....  in this case I prefer the first because it's explicit.  "What happens if there are not any...."  in the update that becomes implied by the early return, which suggests an invisible else in there. That said, opportunities to dedent should be watched for, but I'm gonna leave this one as-is for now.

tests/integration/actions/exec/base.py
cidrblock: yeah, I think there is probably many more place to make the switch to pathlib, but why not start now right?

src/ansible_navigator/actions/exec.py
cidrblock: @webknjaz it might be over the top, but let me know what you think

cidrblock:
recheck

cidrblock:
recheck

src/ansible_navigator/actions/exec.py
cidrblock: @webknjaz any thoughts here? I'll leave this for you to resolve.

docs/changelog-fragments.d/526.feature.md
```diff
@@ -0,0 +1,8 @@
+Added the ability to run ad-hoc commands directly within an execution
+environment using mode `stdout`.
+
+```
+$ ansible-navigator exec --mode stdout
```
webknjaz, line 5: There are a few things to note here:
1) It's usually useful to select the highlighting language for code blocks to avoid ambiguity. This has a better chance of being parsed with the desired lexer. For snippets of commands having `$` prompt, it should usually be `console` or its alias `shell-session`.
2) The way `console` works is that it uses different colors for the commands with prompts and their output.
3) markdownlint will probably be unhappy about having a prompt+command with no output (I've hit that rule recently). Although, you won't see this problem until the release (when this gets into `CHANGELOG.md`) because the notes are excluded from linting.
4) If you only intend to demonstrate a command, you could use `bash`/`sh` instead and drop the prompt.
````suggestion
```console
$ ansible-navigator exec --mode stdout
[example output]
````

File: docs/changelog-fragments.d/526.feature.md [hunk 24]
```diff
@@ -0,0 +1,8 @@
+Added the ability to run ad-hoc commands directly within an execution
+environment using mode `stdout`.
+
```
webknjaz, line 3: How about being a bit clearer wrt what the code snippet below demonstrates?
```suggestion

Here is an example invocation:
```

src/ansible_navigator/actions/exec.py
```diff
@@ -0,0 +1,116 @@
...
+        :return: The stdout, stderr and return code from runner
+        """
+        if isinstance(self._args.set_environment_variable, dict):
+            envvars_to_set = self._args.set_environment_variable.copy()
+        elif isinstance(self._args.set_environment_variable, Constants):
+            envvars_to_set = {}
+        else:
+            log_message = (
+                "The setting 'set_environment_variable' was neither a dictionary"
+                + " or Constants, please raise an issue. No environment variables will be set."
```
webknjaz, line 79: Prefer parse-time concatenation over runtime. With this, the interpreter will treat both lines as a single string literal:
```suggestion
                " or Constants, please raise an issue. No environment variables will be set."
```
But with a `+`, it would parse two literals and only concatenate them when the control flow reaches this place in runtime, which is slightly less performant.

cidrblock: Interesting, don't we introduce implicit string concatenation if we do that while it is explicit with the +?

docs/changelog-fragments.d/526.feature.md
cidrblock: I'm gonna leave it without a prompt or the output of the command.  The reason I added this wan't really so it showed the sample output, but to show how the command looked with `exec` and the ability to cut-paste from the change log...  My theory, completely unproven, is that someone might be more likely to try it out if they could cut and paste :)

src/ansible_navigator/actions/exec.py
webknjaz: Yep. Exactly. Except with "implicit" concatenation, there's no actual concatenation in runtime, the interpreter would "see" the whole thing as if you typed in everything as one big literal. Are you hitting any linter complaints? I think there was a WPS rule that I'd always disable because it doesn't allow implicit concat. But I think having it implicit is more common and reasonable.

docs/changelog-fragments.d/526.feature.md
webknjaz: > My theory, completely unproven, is that someone might be more likely to try it out if they could cut and paste :) (or use the clipboard icon when it's available)

AFAIR that clipboard icon Sphinx extension actually removes the prompt on copy and it's configurable what it should remove. Needs some testing, though, because I'm not sure how it's configured currently.

docs/changelog-fragments.d/526.feature.md
```diff
@@ -0,0 +1,9 @@
...
+Here is an example invocation:
+```console
```
webknjaz, line 5: Only use `console` or `shell-session` if you have a prompt with the command, followed by its output. If there's no prompt, don't use this lexer because it'll treat the content (well, anything that does not start with a prompt or a comment-hash) as "console output". Instead, you could choose something like `bash`.
```suggestion
```bash
```

These names are coming from https://pygments.org/docs/lexers/#pygments.lexers.shell.BashSessionLexer.

src/ansible_navigator/actions/_actions.py
```diff
@@ -95,7 +98,11 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
     """Call the given action's run"""
     action_cls = get(package, action)
     app, interaction = args
-    return action_cls(app.args).run(app=app, interaction=interaction)
+    try:
+        return action_cls(app.args).run(app=app, interaction=interaction)
+    except AttributeError:
+        logger.error("Subcommand '%s' does not support mode interactive", action)
+        return action_cls(app.args).no_interactive_mode(app=app, interaction=interaction)
```
webknjaz, line 105: In this place, try/except approach may be shadowing a few unobvious problem origins. This exception handler will be chosen in the following cases:
1) If `app` does not have `args` attribute
2) If internals of `action_cls()` (or *anything* down the call stack) raise `AttributeError`
3) If the object returned by `action_cls()` does not have a `run` attribute
4) If the `run` attribute exists, is callable but internally something raises `AttributeError`.

It wouldn't be nice if half of instructions/mutations inside of `run()`, error out, only to cause `no_interactive_mode()` to potentially repeat some of them, right? This is why in my original suggestion, I tried to minimize what's expected to be caught in the try/except block. This is also why https://github.com/ansible/ansible-language-server/issues/16 is a thing — it is important to scope the instructions within try/except to what you really expect and let any other exceptions bubble up naturally.

While the scenarios I described may not be happening right now, it is important to understand how dangerous this may become in the future and how hard it would be to debug such cases if they happen.

This is why I'd like to re-iterate the original proposal:
1) Initialize the instance outside of try/except
2) Extract the callable within try/except so that it's the only thing that is processed by this construct
3) Call it outside of try/except

src/ansible_navigator/actions/exec.py
webknjaz: Okay, keep an instance check if there's value in this information.

As for the var name, it's important to understand that the same piece of data may have different semantical meanings on different levels of abstraction — it's one thing on the calling side but another in a called function that operates lower-level primitives.

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,110 @@
...
+import pytest
+
+from ..._common import fixture_path_from_request
+from ..._common import update_fixtures
+from ..._tmux_session import TmuxSession
+from ..._interactions import SearchFor
+from ..._interactions import Step
+from ....defaults import FIXTURES_DIR
+
+TEST_FIXTURE_DIR = Path(FIXTURES_DIR, "integration", "actions", "exec")
```
webknjaz, line 20: WOW, I somehow didn't realize you could pass separate paths to the initializer. TIL

src/ansible_navigator/actions/exec.py
cidrblock: https://github.com/ansible/ansible-navigator/pull/740

src/ansible_navigator/actions/_actions.py
cidrblock: thanks for the explanation, I hadn't considered how deeply the except would go.

btw: https://github.com/ansible/ansible-navigator/pull/741

tests/integration/actions/exec/base.py
```diff
@@ -0,0 +1,110 @@
...
+
+        :param request: The test request
+        :param tmux_session: The tmux session
+        :param step: A step within a series of tests
+        :raises ValueError: If test mode isn't set
+        """
+        if step.search_within_response is SearchFor.PROMPT:
+            search_within_response = tmux_session.cli_prompt
+        else:
+            raise ValueError("test mode not set")
```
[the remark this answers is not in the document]
cidrblock: nice.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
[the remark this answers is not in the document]
cidrblock: @relrod what do you think?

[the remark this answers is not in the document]
relrod: I don't feel super strongly about it, I mainly am just going by what other CLI apps tend to do, but like I mentioned, we're already deviating from the norm by requiring an explicit `true` or `false` anyway.

PR #635, opened 2021-12-02
[linting, tests] Type hinting clean up

### LINKED ISSUE #633: Type hinting errors (mypy)
```
(venv) x1 ➜  ansible-navigator git:(exec) mypy src/ansible_navigator                                                   
src/ansible_navigator/runner/ansible_inventory.py:9: error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
src/ansible_navigator/runner/ansible_doc.py:12: error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
src/ansible_navigator/runner/ansible_config.py:8: error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
src/ansible_navigator/runner/ansible_config.py:8: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
src/ansible_navigator/runner/command_async.py:10: error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
src/ansible_navigator/runner/command.py:6: error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
src/ansible_navigator/actions/run.py:666: error: Argument "seconds" to "human_time" has incompatible type "float"; expected "int"
```

Most likely introduced with #623, missed b/c CI not enforcing

cidrblock:
Fixes #633 

Given the change to human_time, add tests to confirm
linting fixes in the utils test file

### DIFF
--- a/src/ansible_navigator/runner/ansible_config.py
+++ b/src/ansible_navigator/runner/ansible_config.py
@@ -5,7 +5,7 @@ to run the ansible-config command
 from typing import Optional
 from typing import Tuple
 
-from ansible_runner import get_ansible_config  # type: ignore
+from ansible_runner import get_ansible_config
 
 from .base import Base
 
--- a/src/ansible_navigator/runner/ansible_config.py
+++ b/src/ansible_navigator/runner/ansible_config.py
@@ -5,7 +5,7 @@ to run the ansible-config command
 from typing import Optional
 from typing import Tuple
 
-from ansible_runner import get_ansible_config  # type: ignore
+from ansible_runner import get_ansible_config # type: ignore[import]
 
 from .base import Base
 
--- a/src/ansible_navigator/runner/ansible_config.py
+++ b/src/ansible_navigator/runner/ansible_config.py
@@ -5,7 +5,7 @@ to run the ansible-config command
 from typing import Optional
 from typing import Tuple
 
-from ansible_runner import get_ansible_config
+from ansible_runner import get_ansible_config  # type: ignore
 
 from .base import Base
 
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,15 +1,14 @@
 """ tests for the utilities in utils
 """
 import os
-import stat
 
 from typing import List
 from typing import Optional
-from types import SimpleNamespace
+from typing import Union
 
 import pytest
 
-import ansible_navigator.utils as utils
+from ansible_navigator import utils
 
 EXTENTIONS = [".yml", ".yaml", ".json"]
 
@@ -25,9 +24,9 @@ def test_find_many_settings_home(monkeypatch) -> None:
         return arg in paths
 
     monkeypatch.setattr(os.path, "exists", check_path_exists)
-    messages, exit_messages, found = utils.find_settings_file()
+    _messages, exit_messages, _found = utils.find_settings_file()
     expected = f"Only one file among {utils.oxfordcomma(paths, 'and')}"
-    assert any([expected in exit_msg.message for exit_msg in exit_messages])
+    assert any((expected in exit_msg.message for exit_msg in exit_messages))
 
 
 def test_find_many_settings_cwd(monkeypatch) -> None:
@@ -39,9 +38,9 @@ def test_find_many_settings_cwd(monkeypatch) -> None:
         return arg in paths
 
     monkeypatch.setattr(os.path, "exists", check_path_exists)
-    messages, exit_messages, found = utils.find_settings_file()
+    _messages, exit_messages, _found = utils.find_settings_file()
     expected = f"Only one file among {utils.oxfordcomma(paths, 'and')}"
-    assert any([expected in exit_msg.message for exit_msg in exit_messages])
+    assert any((expected in exit_msg.message for exit_msg in exit_messages))
 
 
 def test_find_many_settings_precedence(monkeypatch) -> None:
@@ -54,7 +53,7 @@ def test_find_many_settings_precedence(monkeypatch) -> None:
         return arg in paths
 
     monkeypatch.setattr(os.path, "exists", check_path_exists)
-    messages, exit_messages, found = utils.find_settings_file()
+    _messages, _exit_messages, found = utils.find_settings_file()
     assert expected == found
 
 
@@ -99,3 +98,24 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
     """test for flatten list"""
     actual_result = utils.flatten_list(value)
     assert list(actual_result) == anticpated_result
+
+
+@pytest.mark.parametrize(
+    "value, anticpated_result",
+    [(1, "1s"), (60 + 1, "1m1s"), (3600 + 60 + 1, "1h1m1s"), (86400 + 3600 + 60 + 1, "1d1h1m1s")],
+    ids=["seconds", "minutes seconds", "hours minutes seconds", "days minutes seconds"],
+)
+def test_human_time(value: Union[int, float], anticpated_result: str) -> None:
+    """test for human time"""
+    # pass integer
+    int_result = utils.human_time(value)
+    assert int_result == anticpated_result
+    # pass negative integer
+    neg_int_result = utils.human_time(-value)
+    assert neg_int_result == f"-{anticpated_result}"
+    # pass float
+    float_result = utils.human_time(float(value))
+    assert float_result == anticpated_result
+    # pass negative float
+    neg_float_result = utils.human_time(-value)
+    assert neg_float_result == f"-{anticpated_result}"
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -103,7 +103,7 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
 @pytest.mark.parametrize(
     "value, anticpated_result",
     [(1, "1s"), (60 + 1, "1m1s"), (3600 + 60 + 1, "1h1m1s"), (86400 + 3600 + 60 + 1, "1d1h1m1s")],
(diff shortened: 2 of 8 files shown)


src/ansible_navigator/runner/ansible_config.py
```diff
@@ -5,7 +5,7 @@
 from typing import Optional
 from typing import Tuple
 
-from ansible_runner import get_ansible_config
+from ansible_runner import get_ansible_config  # type: ignore
```
webknjaz, line 8: Instead of adding a catch-all ignore, only ignore specific error codes. It'd look like `# type: ignore[mistake-id]`. This way you wouldn't accidentally ignore extra problems that may arise in the future.

tests/unit/test_utils.py
```diff
@@ -25,9 +24,9 @@ def check_path_exists(arg):
         return arg in paths
 
     monkeypatch.setattr(os.path, "exists", check_path_exists)
-    messages, exit_messages, found = utils.find_settings_file()
+    _messages, exit_messages, _found = utils.find_settings_file()
     expected = f"Only one file among {utils.oxfordcomma(paths, 'and')}"
-    assert any([expected in exit_msg.message for exit_msg in exit_messages])
+    assert any((expected in exit_msg.message for exit_msg in exit_messages))
```
webknjaz, line 29: The correct fix is
```suggestion
    assert any(expected in exit_msg.message for exit_msg in exit_messages)
```

tests/unit/test_utils.py
```diff
@@ -39,9 +38,9 @@ def check_path_exists(arg):
...
```
webknjaz, line 43:
```suggestion
    assert any(expected in exit_msg.message for exit_msg in exit_messages)
```

tests/unit/test_utils.py
```diff
@@ -99,3 +98,24 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
...
     assert list(actual_result) == anticpated_result
+
+
+@pytest.mark.parametrize(
+    "value, anticpated_result",
+    [(1, "1s"), (60 + 1, "1m1s"), (3600 + 60 + 1, "1h1m1s"), (86400 + 3600 + 60 + 1, "1d1h1m1s")],
+    ids=["seconds", "minutes seconds", "hours minutes seconds", "days hours minutes seconds"],
+)
+def test_human_time(value: Union[int, float], anticpated_result: str) -> None:
+    """test for human time"""
```
webknjaz, line 109: What does it text?

tests/unit/test_utils.py
```diff
@@ -99,3 +98,24 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
...
+    assert int_result == anticpated_result
+    # pass negative integer
+    neg_int_result = utils.human_time(-value)
+    assert neg_int_result == f"-{anticpated_result}"
+    # pass float
+    float_result = utils.human_time(float(value))
+    assert float_result == anticpated_result
+    # pass negative float
+    neg_float_result = utils.human_time(-value)
+    assert neg_float_result == f"-{anticpated_result}"
```
webknjaz, line 121: This looks like 4 distinct tests what would be better to either parameterize or separate.

tests/unit/test_utils.py
```diff
@@ -99,3 +98,24 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
     """test for flatten list"""
     actual_result = utils.flatten_list(value)
     assert list(actual_result) == anticpated_result
+
+
+@pytest.mark.parametrize(
+    "value, anticpated_result",
```
webknjaz, line 104: For the params list, it's more readable to use a sequence of args, not a string with coma-separated names. When I have time, I'll integrate the corresponding flake8 plugin that catches this problem.
```suggestion
    ("value", "anticpated_result"),
```

src/ansible_navigator/runner/ansible_config.py
cidrblock: this is b/c type hints aren't available for ansible_runner

```
error: Skipping analyzing "ansible_runner": found module but no type hints or library stubs
```

From the docs: 

```
If you are unable to find any existing type hints nor have time to write your own, you can instead suppress the errors. All this will do is make mypy stop reporting an error on the line containing the import: the imported module will continue to be of type Any.

To suppress a single missing import error, add a # type: ignore at the end of the line containing the import.
```

tests/unit/test_utils.py
cidrblock: doh! ha, thx

tests/unit/test_utils.py
cidrblock: it a unit test for the function called human time in the util file, that func:

```
"""convert seconds into human readable
    00d00h00m00s format"""
```
I'll add this specifically

tests/unit/test_utils.py
```diff
@@ -99,3 +98,24 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
...
+    [(1, "1s"), (60 + 1, "1m1s"), (3600 + 60 + 1, "1h1m1s"), (86400 + 3600 + 60 + 1, "1d1h1m1s")],
+    ids=["seconds", "minutes seconds", "hours minutes seconds", "days hours minutes seconds"],
+)
+def test_human_time(value: Union[int, float], anticpated_result: str) -> None:
```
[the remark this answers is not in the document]
cidrblock: the type of `seconds` for human_time changed from `int` to `Union[int, float]`, so to confirm the function did correctly handle both, the tests were added.

tests/unit/test_utils.py
cidrblock: doing away with this.........................................

tests/unit/test_utils.py
cidrblock: It was simple enough this should be fine.  Will split it out.

src/ansible_navigator/runner/ansible_config.py
webknjaz: Unresolving. This is not what I am asking about. I'm asking to append a specific error code, just like you do with pylint. You don't ignore every linting violation type possible when you see a specific one, you disable just that one thing. Same with mypy.

webknjaz: I hope this document will help you understand what I mean: 
https://mypy.readthedocs.io/en/stable/error_codes.html#error-codes

cidrblock: This was ignored because it is a 3rd party library that does not have hints.

I think this documentation may be more relevant that the one you suggested: https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library

webknjaz: Interesting, does it not have its own code? If not, at least add code comments explaining why this is necessary. In fact, code comments with reasoning are needed every time you ignore a linting problem. Otherwise, it's hard to figure it out in the future.

cidrblock: @webknjaz If you are ware of a better way to handle this, please let me know.

webknjaz: Looks like mypy is not configured to show error codes. Try adding `--show-error-codes` to see them.

webknjaz: Looks like you need
```suggestion
from ansible_runner import get_ansible_config  # type: ignore[import]
```
per https://mypy.readthedocs.io/en/stable/error_code_list.html#check-that-import-target-can-be-found-import

cidrblock: OOoooo..... if that works I'll change it! cool.

cidrblock: I love this, other instances will get cleaned up with this: https://github.com/ansible/ansible-navigator/issues/638

thanks

cidrblock:
@webknjaz Would appreciate a final approval when you have time

tests/unit/test_utils.py
```diff
@@ -99,3 +99,49 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
     """test for flatten list"""
     actual_result = utils.flatten_list(value)
     assert list(actual_result) == anticpated_result
+
+
+class HumanTimeTestData(NamedTuple):
+    """data for human time test"""
```
webknjaz, line 105:
```suggestion
    """Data for human time test."""
```

tests/unit/test_utils.py
```diff
@@ -99,3 +99,49 @@ def test_flatten_list(value: List, anticpated_result: List) -> None:
...
+    HumanTimeTestData(id="minutes seconds", value=60 + 1, expected="1m1s"),
+    HumanTimeTestData(id="hours minutes seconds", value=3600 + 60 + 1, expected="1h1m1s"),
+    HumanTimeTestData(
+        id="days hours minutes seconds", value=86400 + 3600 + 60 + 1, expected="1d1h1m1s"
+    ),
+]
+
+
+@pytest.mark.parametrize("data", human_time_test_data, ids=lambda data: data.id)
+def test_human_time_integer(data: List) -> None:
```
webknjaz, line 123: Didn't mypy complain?
```suggestion
def test_human_time_integer(data: HumanTimeTestData) -> None:
```

webknjaz (commented):
I think this is the last mistake. Feel free to merge after addressing these.

tests/unit/test_utils.py
cidrblock: Thanks, will be good to get a linter check for this in the future!

tests/unit/test_utils.py
cidrblock: Good catch.

Because of time constraints during the initial development of ansible-navigator, mypy and type hinting wasn't covered for the tests.  No time like the present to start using it thought.

PR #637, opened 2021-12-02
[linting] Switch to f strings from format

### LINKED ISSUE #636: [lint] Consider using f string
With python 3.10 and the latest pylint 2.11, the recommendation for f string is now default:

https://pylint.pycqa.org/en/latest/whatsnew/2.11.html#new-checkers

```
(venv) x1 ➜  ansible-navigator git:(lint_errors) pylint src/ansible_navigator --ignore=tm_tokenize --disable=all --enable=consider-using-f-string
************* Module ansible_navigator.utils
src/ansible_navigator/utils.py:355:15: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/utils.py:357:15: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/utils.py:359:15: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/utils.py:360:11: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.steps
src/ansible_navigator/steps.py:115:29: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.configuration_subsystem.navigator_configuration
src/ansible_navigator/configuration_subsystem/navigator_configuration.py:60:18: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.actions.inventory
src/ansible_navigator/actions/inventory.py:62:13: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.actions.doc
src/ansible_navigator/actions/doc.py:185:22: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.actions.run
src/ansible_navigator/actions/run.py:112:17: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/actions/run.py:118:17: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/actions/run.py:136:17: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
************* Module ansible_navigator.ui_framework.ui
src/ansible_navigator/ui_framework/ui.py:248:21: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/ui_framework/ui.py:261:20: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
src/ansible_navigator/ui_framework/ui.py:406:25: C0209: Formatting a regular string which could be a f-string (consider-using-f-string)
```

cidrblock:
Fixes #636

### DIFF
--- a/src/ansible_navigator/actions/doc.py
+++ b/src/ansible_navigator/actions/doc.py
@@ -182,8 +182,9 @@ class Action(App):
                 [self._plugin_name], plugin_type=self._plugin_type, playbook_dir=playbook_dir
             )
             if plugin_doc_err:
-                msg = "Error occurred while fetching doc for" " plugin {0}: '{1}'".format(
-                    self._plugin_name, plugin_doc_err
+                msg = (
+                    "Error occurred while fetching doc for plugin"
+                    f" {self._plugin_name}: '{plugin_doc_err}'"
                 )
                 self._logger.error(msg)
 
--- a/src/ansible_navigator/actions/doc.py
+++ b/src/ansible_navigator/actions/doc.py
@@ -182,12 +182,11 @@ class Action(App):
                 [self._plugin_name], plugin_type=self._plugin_type, playbook_dir=playbook_dir
             )
             if plugin_doc_err:
-                msg = (
-                    "Error occurred while fetching doc for plugin"
-                    f" {self._plugin_name}: '{plugin_doc_err}'"
+                self._logger.error(
+                    "Error occurred while fetching doc for plugin %s: '%s'",
+                    self._plugin_name,
+                    plugin_doc_err,
                 )
-                self._logger.error(msg)
-
             plugin_doc_response = self._extract_plugin_doc(plugin_doc, plugin_doc_err)
             return plugin_doc_response
         else:
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -57,7 +57,7 @@ def generate_editor_command() -> str:
     else:
         message = "EDITOR environment variable set as '{editor}'"
         initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
-        command = "%s {filename}" % os.environ.get("EDITOR")
+        command = f"{os.environ.get('EDITOR')} {{filename}}"
     message = f"Default editor_command set to: {command}"
     initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
     return command
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -55,9 +55,9 @@ def generate_editor_command() -> str:
         initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
         command = "vi +{line_number} {filename}"
     else:
-        message = "EDITOR environment variable set as '{editor}'"
+        message = f"EDITOR environment variable set as '{editor}'"
         initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
-        command = f"{editor} {{filename}}"
+        command = f"{editor!s} {{filename}}"
     message = f"Default editor_command set to: {command}"
     initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
     return command
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -57,7 +57,7 @@ def generate_editor_command() -> str:
     else:
         message = "EDITOR environment variable set as '{editor}'"
         initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
-        command = f"{os.environ.get('EDITOR')} {{filename}}"
+        command = f"{editor} {{filename}}"
     message = f"Default editor_command set to: {command}"
     initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
     return command
--- a/src/ansible_navigator/utils.py
+++ b/src/ansible_navigator/utils.py
@@ -352,7 +352,7 @@ def human_time(seconds: Union[int, float]) -> str:
     hours, seconds = divmod(seconds, 3600)
     minutes, seconds = divmod(seconds, 60)
     if days > 0:
-        return f"{sign_string}{days}d{hours}h{minutes}m{seconds}s"
+        return f"{sign_string!s}{days:d}d{hours:d}h{minutes}m{seconds}s"
     if hours > 0:
         return f"{sign_string}{hours}h{minutes}m{seconds}s"
     if minutes > 0:
--- a/src/ansible_navigator/utils.py
+++ b/src/ansible_navigator/utils.py
@@ -352,7 +352,7 @@ def human_time(seconds: Union[int, float]) -> str:
     hours, seconds = divmod(seconds, 3600)
     minutes, seconds = divmod(seconds, 60)
     if days > 0:
-        return f"{sign_string!s}{days:d}d{hours:d}h{minutes}m{seconds}s"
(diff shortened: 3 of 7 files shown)


src/ansible_navigator/actions/doc.py
```diff
@@ -182,8 +182,9 @@ def _run_runner(self) -> Union[None, dict, Tuple[str, str, int]]:
                 [self._plugin_name], plugin_type=self._plugin_type, playbook_dir=playbook_dir
             )
             if plugin_doc_err:
-                msg = "Error occurred while fetching doc for" " plugin {0}: '{1}'".format(
-                    self._plugin_name, plugin_doc_err
+                msg = (
+                    "Error occurred while fetching doc for plugin"
+                    f" {self._plugin_name}: '{plugin_doc_err}'"
```
webknjaz, line 187: I see that you pass this to the logger. For loggers, the rules are different — they interpolate strings internally, and it's lazy, meaning that it may even be skipped based on the logging settings.
The proper way is to pass the template string as the first argument, and pass the values as positional arguments.

P.S. Another reason to do this is that some tools that collect logging statistics are better at counting which invocation was hit multiple times based on the static template rather than different pre-rendered strings passed to the same logging call.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -57,7 +57,7 @@ def generate_editor_command() -> str:
     else:
         message = "EDITOR environment variable set as '{editor}'"
         initialization_messages.append(LogMessage(level=logging.DEBUG, message=message))
-        command = "%s {filename}" % os.environ.get("EDITOR")
+        command = f"{os.environ.get('EDITOR')} {{filename}}"
```
webknjaz, line 60: For the conversion to be really equivalent, use the same explicit conversion specifier
```suggestion
        command = f"{os.environ.get('EDITOR')!s} {{filename}}"
```

src/ansible_navigator/utils.py
```diff
@@ -352,12 +352,12 @@ def human_time(seconds: int) -> str:
     hours, seconds = divmod(seconds, 3600)
     minutes, seconds = divmod(seconds, 60)
     if days > 0:
-        return "%s%dd%dh%dm%ds" % (sign_string, days, hours, minutes, seconds)
+        return f"{sign_string}{days}d{hours}h{minutes}m{seconds}s"
```
webknjaz, line 355: This would be equivalent to the original:
```suggestion
        return f"{sign_string!s}{days:d}d{hours:d}h{minutes:d}m{seconds:d}s"
```

webknjaz (changes requested):
Some of the conversations are not equivalent, this is probably not exactly what you meant.

src/ansible_navigator/actions/doc.py
cidrblock: yeah it definitely going to the logger, easy enough to move, not that it was computationally expensive :)

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: @webknjaz Can you show me where os.environ.get returns anything but a string?

src/ansible_navigator/actions/doc.py
webknjaz: Right, I guess pylint missed this because it's a separate var that it didn't understand that it'll end up being passed to the logger.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
webknjaz: I think it can return `None` if there's no such key. But my comment was about a different thing — the syntax you used is not equivalent. The original string used an explicit placeholder "its a string" which for your f-string variant it's implicit. If fact, off the top of my head I can't even remember whether it's `str()` or `repr()` by default — this is why I prefer explicit over implicit.

src/ansible_navigator/utils.py
cidrblock: @webknjaz sign_string is already a string, and I don't see a need to use :d here since an f string will str() the value

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: @webknjaz in this case, it was already checked, but we should use that instead :) (fixed)

a string used in an fstring will be rendered as a string, so I don't see the difference knowing it is a string, and the outcome of the f string is a string which I guess is implied?

webknjaz: The main motivation was to point out that your conversions are not equivalent. It's best not to mix up formatting changes and potential functional changes. When we make exceptions for these rules, we tend to miss things because the focus is on the different aspect of the change. Plus from readability perspective, it's nice to say "I expect this to be a string" rather than "I hope this will get turned into a string at this point of the control flow regardless of whether something above gets refactored". When you do code changes, do you go through all the places where a variable that you're changing is going to be rendered thinking about the corner cases that may happen in the places of rendering?
Having explicit markers helps with the readability long after you've committed the change.

src/ansible_navigator/utils.py
webknjaz: Then, the behavior will not be equivalent. As I said above, explicit markers help with readability. But not only that. When you instruct it to call `int()` (which is what `:d` does essentially) on the rendered object, it's also a sanity check/safeguard for free, it will fail loudly if that thing will not be convertable to a number. Some time in the future, when you refactor things, you'd appreciate this rather than having a silent conversion and hearing about a bug from the end-users when it's shipped.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: Happy to make the change here. Hopefully it doesn't raise questions in the future because `os.environ.get` would be a string and the `!s` is not required.

src/ansible_navigator/utils.py
cidrblock: again, happy to make this change. IMO it feels unnecessary, but if it helps you and contributors in the future it's worthwhile.

cidrblock: @webknjaz Looking at this again, I don't think the use of `:d` if functionally equivalent to the initial string.   In the case divmod returns a float, the use of `%d` or the use of an f-string will convert the float to a string, where `:d` does not.

```
>>> "%dd" % 5
'5d'
>>> "%dd" % 5.0
'5d'
>>> foo = 5
>>> f"{foo:d}"
'5'
>>> foo = 5.0
>>> f"{foo:d}"
>>> f"{foo}"
'5.0'
  File "<stdin>", line 1, in <module>
ValueError: Unknown format code 'd' for object of type 'float'
``` 

If we want to use the `:d` within the fstring, I think it is best to ensure the values returned by a divmod are explicitly integers.  Although it might be questionably out of scope for the PR, I think it is worthwhile because the type hints for the helper function show a float or int can be passed but we will always return a tuple of ints.

cidrblock: The helper function should eliminate the possibility of someone modifying the function and passing a float to divmod which would result in a response from divmod of `float, float`, and cause the f-string to choke.

cidrblock:
recheck

src/ansible_navigator/utils.py
webknjaz: I guess. My point is that we need to make sure that the f-string will choke as early as possible and part of that is using explicit specifiers. Another thing is the readability aspect of this: if a person looks at just one line, it's obvious what sort of data is being rendered w/o having to backtrace what the variable is, where it's defined and how it's passed or mutated.

cidrblock: @webknjaz ok to gate and merge?

webknjaz: Okay, let's do it.

webknjaz (approved):

PR #639, opened 2021-12-03
[typing] Clarify some type ignores, remove uneeded

### LINKED ISSUE #638: [typing] Clarify use of `# type: ignore` where used
Rather than a broad ignore, include the specific check

eg 

`# type: ignore[import]`

vs

`# type: ignore`

cidrblock:
Fixes #638 
Specify the reason for some `# type: ignore` statements, remove ignore statements that are needed

### DIFF
--- a/src/ansible_navigator/actions/_actions.py
+++ b/src/ansible_navigator/actions/_actions.py
@@ -16,7 +16,7 @@ from typing import Tuple
 try:
     from importlib import resources
 except ImportError:
-    import importlib_resources as resources  # type: ignore
+    import importlib_resources as resources  # type: ignore[import, no-redef]
 
 
 # Basic structure for storing information about one action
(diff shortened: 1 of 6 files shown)


src/ansible_navigator/actions/_actions.py
```diff
@@ -16,7 +16,7 @@
 try:
     from importlib import resources
 except ImportError:
-    import importlib_resources as resources  # type: ignore
+    import importlib_resources as resources  # type: ignore[import, no-redef]
```
webknjaz, line 19: One of the idioms to use with mypy is if-else rather than try-except for fallback imports. It's a known bug that mypy does not understand that it's one or the other. But with the `if py_ver:` style, it is able to figure out what's going on and wouldn't need ignores.

cidrblock: @webknjaz  Good idea, although I think we should wait on this until we get better python version test coverage in place when CI is broght up to speed.  I'll log an issue so it can be changed later.

cidrblock: Related: #646

webknjaz (approved):

PR #641, opened 2021-12-03
[typing] Ensure `:run` passes rounded int to `utils.human_time`

### LINKED ISSUE #640: [:run] May pass a float to `utils.human_time`
```
run.py:666: error: Argument "seconds" to "human_time" has incompatible type "float"; expected "int"  [arg-type]
```

mypy raised an error explaining that utils.human_time may be passed either an integer or float from the run action

cidrblock:
depends-on: https://github.com/ansible/ansible-navigator/pull/671
Fixes #640 

- Without a guarantee of the runner duration value being an int, assume it can be either float or int, and round half up
- helper function added to utils
- unit test to help demonstrate the behavior of decimal

### DIFF
--- a/docs/changelog-fragments.d/640.misc.md
+++ b/docs/changelog-fragments.d/640.misc.md
@@ -1,0 +1,2 @@
+Ensure durations in the TUI are rounding consistently (half-up).
+-- by {user}`cidrblock`
--- a/docs/changelog-fragments.d/640.misc.md
+++ b/docs/changelog-fragments.d/640.misc.md
@@ -1,2 +1,2 @@
-Ensure durations in the TUI are rounding consistently (half-up).
+Ensure durations in the TUI are rounded consistently (half-up).
 -- by {user}`cidrblock`
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -3,6 +3,7 @@
 import os
 
 from typing import List
+from typing import NamedTuple
 from typing import Optional
 from typing import NamedTuple
 from typing import Union
@@ -157,3 +158,32 @@ def test_human_time_negative_float(data: HumanTimeTestData) -> None:
     """
     result = utils.human_time(-float(data.value))
     assert result == f"-{data.expected}"
+
+
+class RoundHalfUpTestData(NamedTuple):
+    """Data for round half up tests."""
+
+    id: str
+    value: Union[int, float]
+    expected: int
+
+
+round_half_up_test_data = [
+    RoundHalfUpTestData(id="integer", value=1, expected=1),
+    RoundHalfUpTestData(id="negative integer", value=-1, expected=-1),
+    RoundHalfUpTestData(id="down float", value=1.49999999, expected=1),
+    RoundHalfUpTestData(id="up float", value=1.50000001, expected=2),
+    RoundHalfUpTestData(id="negative down float", value=-1.49999999, expected=-1),
+    RoundHalfUpTestData(id="negative up float", value=-1.50000001, expected=-2),
+]
+
+
+@pytest.mark.parametrize("data", round_half_up_test_data, ids=lambda data: data.id)
+def test_round_half_up(data: RoundHalfUpTestData) -> None:
+    """Test for the utils.round_half_up function.
+
+    Ensure the number passed is consistently rounded to the nearset integer with ties going away from zero.
+    """
+    for _iteration in range(100):
+        result = utils.round_half_up(data.value)
+        assert result == data.expected
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -175,7 +175,7 @@ round_half_up_test_data = [
     RoundHalfUpTestData(id="negative down float", value=-1.49999999, expected=-1),
     RoundHalfUpTestData(id="negative up float", value=-1.50000001, expected=-2),
     RoundHalfUpTestData(id="half_even", value=2.5, expected=3),
-    RoundHalfUpTestData(id="half_even", value=3.5, expected=4)
+    RoundHalfUpTestData(id="half_even", value=3.5, expected=4),
 ]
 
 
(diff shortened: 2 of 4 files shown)


tests/unit/test_utils.py
```diff
@@ -157,3 +157,33 @@ def test_human_time_negative_float(data: HumanTimeTestData) -> None:
...
+
+
+@pytest.mark.parametrize("data", round_half_up_test_data, ids=lambda data: data.id)
+def test_round_half_up(data: RoundHalfUpTestData) -> None:
+    """Test for the utils.round_half_up function.
+
+    Ensure the number passed is consistently rounded to the nearset
+    integer with ties going away from zero.
+    """
+    for _iteration in range(100):
```
webknjaz, line 187: I don't understand the point of this loop. Why is it here? Are you testing that calling the function more than once gives the same result? I don't really get it...

tests/unit/test_utils.py
```diff
@@ -157,3 +157,33 @@ def test_human_time_negative_float(data: HumanTimeTestData) -> None:
...
+    RoundHalfUpTestData(id="negative integer", value=-1, expected=-1),
+    RoundHalfUpTestData(id="down float", value=1.49999999, expected=1),
+    RoundHalfUpTestData(id="up float", value=1.50000001, expected=2),
+    RoundHalfUpTestData(id="negative down float", value=-1.49999999, expected=-1),
+    RoundHalfUpTestData(id="negative up float", value=-1.50000001, expected=-2),
+]
+
+
+@pytest.mark.parametrize("data", round_half_up_test_data, ids=lambda data: data.id)
+def test_round_half_up(data: RoundHalfUpTestData) -> None:
```
webknjaz, line 181: To me, this looks like an ideal use case to integrate property-based testing: https://hypothesis.rtfd.io.

tests/unit/test_utils.py
cidrblock: Ha, the intention was to demonstrate the consistency of the rounding technique vs. round(), but this doesn't even do that :) I'll remove the loop and make a note

tests/unit/test_utils.py
cidrblock: that's a cool library, in this case we'd have to find a way to determine the expected result without using the function we were testing.  Certainly something to look at in the future.

webknjaz: I've been looking for an excuse to integrate Hypothesis for years but never got to it :) I guess we need to start somewhere...

webknjaz (approved):

webknjaz:
@cidrblock I think you could start adding change notes like `docs/changelog-fragments.d/640.bugfix.md` (is it a bugfix, or a feature, or maybe misc?) even prior to merging #654.

ssbarnea:
> @cidrblock I think you could start adding change notes like `docs/changelog-fragments.d/640.bugfix.md` (is it a bugfix, or a feature, or maybe misc?) even prior to merging #654.

NO, this would count as scope-creeping. You cannot ask someone to add unrelated stuff. Be patient and allow the team to decide if we go for towncrier or not, especially as we have more serious problems now like broken CI.

In fact I would recommend blocking any pull-request that does not directly goes towards making CI report green.

webknjaz:
> NO, this would count as scope-creeping. You cannot ask someone to add unrelated stuff. Be patient and allow the team to decide if we go for towncrier or not, especially as we have more serious problems now like broken CI.

It is absolutely related. Who is going to manually track adding proper change notes otherwise?

cidrblock:
although not required at this time, I will opt-in and add the change log

docs/changelog-fragments.d/640.misc.md
```diff
@@ -0,0 +1,2 @@
+Ensure durations in the TUI are rounded consistently (half-up).
```
webknjaz, line 1: @cidrblock I missed this but there's a few comments:
1) Markdown, when it's rendered, will produce a single line/paragraph so there will be no line break between the sentence and the signature. It'll be like this:
   ```text
   Ensure durations in the TUI are rounded consistently (half-up). -- by {user}`cidrblock`
   ```
   _(preview: https://ansible-navigator--641.org.readthedocs.build/en/641/changelog/#miscellaneous)_
   It's not documented explicitly but you can see that there's a style weirdness with an em dash after the period. This is why in the examples and existing fragments, I don't add that. I don't know if it's the best solution in the world but this is what I'm following, at least for consistency.
2) It's mentioned in https://github.com/ansible/ansible-navigator/tree/main/docs/changelog-fragments.d#alright-so-how-do-i-add-a-news-fragment but let me re-iterate — release notes reflect what changed in the experience for the end-users since the previous release. This means that you need to describe the change using the past tense (or a different way of similarly emphasizing that this has happened already at the time of reading/release). As it stands, your change note is inconsistent with the existing ones style-wise (it's quite visual if you open the changelog page on RTD and look at the draft rendered together).
   It is important to understand that the rules here are different from what is generally expected for commit messages. While the Git commits need to use the present tense and go into the _technical internal implementation details_, the release notes would need to tell the story of how the interactions of the user with the software got affected and changed. Using "ensure" at the beginning implies that maybe there was a bug that got fixed, or maybe it's a one-time thing.
   I think it's best to be explicit about whether something has changed compared to the previous version, started, or stopped happening. You may need to be specific about what part or property of the software you are talking about. You mention "durations" and "TUI" but it still does not give me an idea of what to expect when I run the software — where exactly am I supposed to look to notice the difference? Could it be "progress bar" widgets? ETAs? Some long-running process outputs?
   In release notes, you are allowed to be more verbose. Type in more, even add a paragraph of prose.
   Here's an example (with a few made-up things):
   ```text
   Started to consistently round durations using the half-up strategy in the TUI components that show
   <process> output/results -- by {user}`cidrblock`

   Before this, the resulting time representation was unpredictable resulting in jumpy rerendering of
   the UI elements displaying this information.
   ```

Issue #643, opened 2021-12-03
[lint] Lint errors in test directory


cidrblock:
```
************* Module tests.defaults
tests/defaults.py:1:0: C0114: Missing module docstring (missing-module-docstring)
************* Module tests.unit.test_image_introspection
tests/unit/test_image_introspection.py:5:0: R0402: Use 'from ansible_navigator import utils' instead (consider-using-from-import)
************* Module tests.unit.test_yaml
tests/unit/test_yaml.py:1:0: C0114: Missing module docstring (missing-module-docstring)
tests/unit/test_yaml.py:4:0: C0116: Missing function or method docstring (missing-function-docstring)
************* Module tests.unit.test_cli
tests/unit/test_cli.py:99:11: C1803: 'exit_msgs == []' can be simplified to 'not exit_msgs' as an empty sequence is falsey (use-implicit-booleaness-not-comparison)
tests/unit/test_cli.py:112:11: C1803: 'exit_msgs == []' can be simplified to 'not exit_msgs' as an empty sequence is falsey (use-implicit-booleaness-not-comparison)
************* Module tests.unit.test_log_append
tests/unit/test_log_append.py:5:0: R0402: Use 'from ansible_navigator import cli' instead (consider-using-from-import)
************* Module tests.unit.configuration_subsystem.test_invalid_params
tests/unit/configuration_subsystem/test_invalid_params.py:80:49: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
************* Module tests.unit.configuration_subsystem.conftest
tests/unit/configuration_subsystem/conftest.py:45:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module tests.unit.configuration_subsystem.test_fixture_sanity
tests/unit/configuration_subsystem/test_fixture_sanity.py:25:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module tests.unit.configuration_subsystem.test_previous_cli
tests/unit/configuration_subsystem/test_previous_cli.py:81:11: C1803: 'exit_messages == []' can be simplified to 'not exit_messages' as an empty sequence is falsey (use-implicit-booleaness-not-comparison)
tests/unit/configuration_subsystem/test_previous_cli.py:101:11: C1803: 'exit_messages == []' can be simplified to 'not exit_messages' as an empty sequence is falsey (use-implicit-booleaness-not-comparison)
[...]

PR #644, opened 2021-12-03
[tests] Fix pylint errors in `tests/unit`

cidrblock:
Related #643 

- does what it says
- several doc string/sentence fixes when seen

### DIFF
--- a/tests/unit/configuration_subsystem/conftest.py
+++ b/tests/unit/configuration_subsystem/conftest.py
@@ -42,7 +42,7 @@ def _generate_config(params=None, setting_file_name=None, initial=True) -> Gener
 
     if setting_file_name:
         settings_file_path = os.path.join(TEST_FIXTURE_DIR, setting_file_name)
-        with open(settings_file_path) as file:
+        with open(file=settings_file_path, encoding="utf-8") as file:
             try:
                 settings_contents = yaml.load(file, Loader=Loader)
             except yaml.parser.ParserError:
--- a/tests/defaults.py
+++ b/tests/defaults.py
@@ -1,3 +1,5 @@
+"""Constants with default values used throughout the tests.
+"""
 import os
 
 FIXTURES_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "fixtures"))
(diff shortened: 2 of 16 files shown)


tests/unit/configuration_subsystem/conftest.py
```diff
@@ -42,7 +42,7 @@ def _generate_config(params=None, setting_file_name=None, initial=True) -> Gener
 
     if setting_file_name:
         settings_file_path = os.path.join(TEST_FIXTURE_DIR, setting_file_name)
-        with open(settings_file_path) as file:
+        with open(file=settings_file_path, encoding="utf-8") as file:
```
webknjaz, line 45: Why was this necessary? It's the default. Also, it's best to follow what the `tmp_path` does and operate on the pathlib objects. They also have `pathlib.Path.open()` if that's really necessary.

webknjaz, line 45: Why was this necessary? It's the default. Also, it's best to follow what the `tmp_path` does and operate on the pathlib objects. They also have `pathlib.Path.open()` if that's really necessary.

cidrblock: pylint W1514 refers to PEP 597

cidrblock: @webknjaz ^

cidrblock: @webknjaz see above

cidrblock:
recheck

cidrblock:
recheck

tests/unit/configuration_subsystem/conftest.py
webknjaz: > pylint W1514 refers to PEP 597

Oh, interesting, I didn't realize. This raises another question — why didn't pytest reveal the warning. Ah! It's probably because we don't test on Python 3.10 (and maybe pytest also isn't set up right...)

But this is resolved for now, not really related to the PR.

webknjaz: Oh, it's a separate question but anyway it's not the right place to solve it. Resolved.

webknjaz (approved):

PR #648, opened 2021-12-04
[tests] Fix pylint errors in `tests/interactive`

cidrblock:
Related #643 

- does what it says
- formatted/added specifics to doc strings as encountered
- remove unneeded pylint disables as encountered
- one constant name correction made when seen

### DIFF
--- a/tests/integration/_action_run_test.py
+++ b/tests/integration/_action_run_test.py
@@ -22,7 +22,9 @@ from ansible_navigator.steps import Steps
 
 
 class ActionRunTest:
-    """directly run an action"""
+    # pylint: disable=too-many-instance-attributes
+    # pylint: disable=too-many-arguments
+    """Directly run an action."""
 
     def __init__(
         self,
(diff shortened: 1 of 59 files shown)


tests/integration/_action_run_test.py
```diff
@@ -137,23 +140,26 @@ def run_action_stdout(self, **kwargs) -> Tuple[int, str, str]:
         sys.stdin = stty  # type: ignore
 
         # set stderr and stdout to fds
-        sys.stdout = tempfile.TemporaryFile()  # type: ignore
-        sys.stderr = tempfile.TemporaryFile()  # type: ignore
+        with tempfile.TemporaryFile() as sys_stdout:
```
webknjaz, line 143: I think you could merge two with-blocks into one.

tests/integration/_tmux_session.py
```diff
@@ -161,10 +164,14 @@ def interaction(
         ignore_within_response=None,
         timeout=300,
     ):
-        """interact with the tmux session
+        # pylint: disable=too-many-branches
```
webknjaz, line 167: Is it a conscious choice not to have coma-separated disables?

cidrblock: yeah, I prefer one liners

tests/integration/_action_run_test.py
cidrblock: good idea

PR #650, opened 2021-12-05
Add MVP GHA workflows

ssbarnea:
This change aims to add a MVP implementation of GHA that replicates current zuul config.

Deliberately it does not include features that are not already covered by zuul jobs, and it also does not include a release pipeline.

The `look_nots` are being removed because `/home/runner` was used as a check in those tests to confirm that the test was not running in an EE.   Because GHA uses `runner` as the default user in it's images, we cannot use this test in the future.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -1,0 +1,78 @@
+name: tox
+
+# https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  create:  # is used for publishing to PyPI and TestPyPI
+    tags:  # any tag regardless of its name, no branches
+      - "**"
+  push:  # only publishes pushes to the main branch to TestPyPI
+    branches:  # any integration branch but not tag
+      - "main"
+  pull_request:
+
+jobs:
+  tox:
+    name: ${{ matrix.tox_env }}
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        tox_env:
+          - lint
+          - build-docs
+        python-version:
+          - 3.8
+        include:
+          - tox_env: lint-vetting
+            devel: true
+          - tox_env: py36
+            python-version: 3.6
+          - tox_env: py37
+            python-version: 3.7
+          - tox_env: py38
+            python-version: 3.8
+          - tox_env: py39
+            python-version: 3.9
+          - tox_env: py310
+            python-version: "3.10"
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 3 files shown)


webknjaz (dismissed):
Please don't merge this, I'm working on a better and more generic version.

cidrblock (approved):
Thank you for this.  I really appreciate the idea of an uncomplicated, baseline, MVP that can be iterated on as needed.

This will help avoid the transition from ansible-zuul to software factory.... which would just burn time.

cidrblock:
> Please don't merge this, I'm working on a better and more generic version.

I'm dismissing this as the scope here is now bare-minimum MVP which can be iterated on later as needed.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,78 @@
...
+          which python
+      - name: Install tox
+        run: |
+          python3 -m pip install --upgrade tox
+      - name: Log installed dists
+        run: >-
+          python3 -m pip freeze --all
+      - name: "tox -e ${{ matrix.tox_env }}"
+        continue-on-error: ${{ matrix.devel || false }}
+        run: |
```
relrod, line 65: nit: no need to use a block here, the command fits on one line.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,78 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: >-
+          Log the currently selected Python
+          version info (${{ matrix.python-version }})
+        run: |
+          python --version --version
+          which python
+      - name: Install tox
+        run: |
```
relrod, line 58: nit: no need to use a block here, the command fits on one line.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,78 @@
...
+          python3 -m pip install --upgrade tox
+      - name: Log installed dists
+        run: >-
```
relrod, line 61: nit: no need to use a block here, the command fits on one line.

relrod (approved):
:+1: GHA

.github/workflows/tox.yml
ssbarnea: I do know and is by-design, so when I can easily add a second command inside the same run. Also forcing block on all runs does make easier to read the workflow file. Anyway, i will not ask anyone to do the same.

.github/workflows/tox.yml
```diff
@@ -0,0 +1,78 @@
...
+          python3 -m pip freeze --all
+      - name: "tox -e ${{ matrix.tox_env }}"
+        continue-on-error: ${{ matrix.devel || false }}
+        run: |
+          python3 -m tox
+        env:
+          TOXENV: ${{ matrix.tox_env }}
+
+  check:
+    if: always()
```
webknjaz, line 71: @ssbarnea another place missing the correct check that will allow failed PRs.

PR #651, opened 2021-12-05
[lint, type] Linting, docstring and type improvements in `_` files

cidrblock:
depends-on: https://github.com/ansible/ansible-navigator/pull/671
- does what it says

### DIFF
--- a/src/ansible_navigator/__init__.py
+++ b/src/ansible_navigator/__init__.py
@@ -1,3 +1,2 @@
-""" init
-"""
+"""The initialization file for ansible-navigator."""
 from ._version import __version__  # noqa: F401
--- a/src/ansible_navigator/__init__.py
+++ b/src/ansible_navigator/__init__.py
@@ -1,2 +1,2 @@
-"""The initialization file for ansible-navigator."""
+"""The ansible-navigator application."""
 from ._version import __version__  # noqa: F401
--- a/src/ansible_navigator/_version.py
+++ b/src/ansible_navigator/_version.py
@@ -1,7 +1,6 @@
-""" version
+"""Version information for ansible-navigator.
 
 Note:
-
 __version__:
 
 indicates the version of the ansible_navigator applications
@@ -10,7 +9,7 @@ __version_collection_doc_cache__:
 
 indicates the version of the schema of the collection doc cache
 this is checked during initialization, if the version of the cache
-differes from below, the cache will be rebuilt.  This should be
+differs from below, the cache will be rebuilt.  This should be
 incremented when the schema changes and need not correspond to the
 application version, although keeping the major in sync is probably
 not a bad idea to minimize the amount of stale docs in the user's cache
--- a/src/ansible_navigator/_version.py
+++ b/src/ansible_navigator/_version.py
@@ -1,18 +1,19 @@
 """Version information for ansible-navigator.
 
-Note:
-__version__:
+.. note::
 
-indicates the version of the ansible_navigator applications
+   __version__:
 
-__version_collection_doc_cache__:
+   Indicates the version of the ansible_navigator application.
 
-indicates the version of the schema of the collection doc cache
-this is checked during initialization, if the version of the cache
-differs from below, the cache will be rebuilt.  This should be
-incremented when the schema changes and need not correspond to the
-application version, although keeping the major in sync is probably
-not a bad idea to minimize the amount of stale docs in the user's cache
+   __version_collection_doc_cache__:
+
+   Indicates the version of the schema of the collection doc cache
+   this is checked during initialization, if the version of the cache
+   differs from below, the cache will be rebuilt.  This should be
+   incremented when the schema changes and need not correspond to the
+   application version, although keeping the major in sync is probably
+   not a bad idea to minimize the amount of stale docs in the user's cache
 """
 __version__ = "1.1.0a1"
 __version_collection_doc_cache__ = "1.0"
--- a/src/ansible_navigator/_yaml.py
+++ b/src/ansible_navigator/_yaml.py
@@ -29,7 +29,9 @@ class YamlStyle(NamedTuple):
 
 
 def human_dump(obj: Any, filename: str = None, fmode: str = "w") -> Union[str, None]:
-    """Consistant dumping of object across the application.
+    """Serialize an object to yaml.
+
+    This allows for the consistent representation across the application.
 
     :param obj: The object to serialize
     :param filename: The filename of the file in which the obj should be written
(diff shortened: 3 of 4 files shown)


src/ansible_navigator/__init__.py
```diff
@@ -1,3 +1,2 @@
-""" init
-"""
+"""The initialization file for ansible-navigator."""
```
webknjaz, line 1: `__init__` represents a package (directory) docstring, not a single module (file). This is why such docstrings are expected to describe the package (as in a Python importable package, not a distribution package).

It'd be something like
```suggestion
"""Ansible-navigator TUI."""
```

You could optionally extend it with "doctest"-style examples or RST code block snippets. For example, look at this https://cheroot.cherrypy.dev/en/latest/pkg/cheroot.cli/ (although this specific example isn't for a package but for a single module, it should demonstrate the idea).

src/ansible_navigator/_yaml.py
```diff
@@ -22,14 +22,20 @@
...
-    """the params for yaml dump"""
+    """The parameters for yaml dump."""
 
     default_flow_style: bool = False
     explicit_start: bool = True
 
 
 def human_dump(obj: Any, filename: str = None, fmode: str = "w") -> Union[str, None]:
-    """Consistant dumping of object across the application"""
+    """Consistant dumping of object across the application.
```
webknjaz, line 32: 1) There's a typo in "Consistent"
2) Functions/methods represent actions and their docstrings should start with a verb. Does it do something or is it just a static concept you're describing?

```suggestion
    """Dump an object consistently across the application.
```

(you could optionally expand on what it actually means in a separate paragraph)

src/ansible_navigator/_yaml.py
```diff
@@ -45,10 +51,16 @@ def human_dump(obj: Any, filename: str = None, fmode: str = "w") -> Union[str, N
 
 class HumanDumper(Dumper):
     # pylint: disable=too-many-ancestors
-    """for block scalar for multiline"""
+    """Dump a multiline in a human readable format."""
```
webknjaz, line 54: This one is the exact opposite of my previous comment — it should describe a static thing/concept/object. By itself, a class instance does not do anything (unless it implements a `__call__()` method, that is; but even then, I'd put the action in the dunder-call docstring, not the class docstring), it does not dump a multiline; its methods do the dumping but merely initializing this object does not have side-effects as it currently implies.

*Hint:* for classes/constants/vars/attrs/properties/modules/packages, describe what they are, not what they "do". If there's a side effect of doing something during the initialization, something is seriously wrong with that design.

```suggestion
    """An implementation of YAML the dumping protocol

    It is designed with human readability in mind and its methods
    emit multiline string representation.
    """
```
(not sure how precise this is but I hope you got the idea)

src/ansible_navigator/_yaml.py
```diff
@@ -62,8 +74,15 @@ def represent_scalar(self, tag, value, style=None):
         return super().represent_scalar(tag, value, style)
 
 
-# from http://stackoverflow.com/a/15423007/115478
-def _is_multiline_string(value):
+def _is_multiline_string(value: str):
+    """Determine if a string is multiline.
+
+    Note:
+    from http://stackoverflow.com/a/15423007/115478
```
webknjaz, line 81: We can use an RST admonition for this:
```suggestion
    .. note::

       Inspired by http://stackoverflow.com/a/15423007/115478.
```

webknjaz: Or maybe it's better to be
```suggestion
    .. tip::

       Inspired by http://stackoverflow.com/a/15423007/115478.
```
(or "hint"?)

src/ansible_navigator/_yaml.py
```diff
@@ -62,8 +74,15 @@ def represent_scalar(self, tag, value, style=None):
...
+
+    :param value: The value to check
+    :return: a boolean indicating is the string is multiline
```
webknjaz, line 84: Other docstrings have this capitalize. So it'd be consistent to do
```suggestion
    :return: A boolean indicating if the string is multiline
```
Also `s/is/if/`

src/ansible_navigator/__init__.py
cidrblock: @webknjaz in this case the init file is for the ansible_navigator package (the entirety of the source)....  It probably should have been an underscore?

cidrblock: got it

src/ansible_navigator/_yaml.py
cidrblock: got it

src/ansible_navigator/_yaml.py
cidrblock: got it

src/ansible_navigator/_yaml.py
cidrblock: got it

src/ansible_navigator/_yaml.py
cidrblock: ty

cidrblock:
recheck

src/ansible_navigator/_version.py
```diff
@@ -1,7 +1,6 @@
-""" version
+"""Version information for ansible-navigator.
 
 Note:
```
webknjaz, line 3: Looks like this could be an RST `.. note::` admonition.

src/ansible_navigator/_version.py
```diff
@@ -1,7 +1,6 @@
...
-
```
webknjaz, line 4: I wouldn't remove this line. Other vars below have a separate empty line before them. Plus if you use an admonition, the whole block would need to be indented.

src/ansible_navigator/_yaml.py
```diff
@@ -45,10 +53,19 @@ def human_dump(obj: Any, filename: str = None, fmode: str = "w") -> Union[str, N
...
 class HumanDumper(Dumper):
     # pylint: disable=too-many-ancestors
-    """for block scalar for multiline"""
+    """An instance of a pyyaml Dumper.
+
+    This deviates from the base to dump a multiline string in a human readable format.
+    """
+
+    def represent_scalar(self, tag: str, value: str, style=Union[str, None]) -> yaml.ScalarNode:
+        """Use a block scalar for a nicer human representation of multiline strings.
```
webknjaz, line 62: What does this do exactly? "use" does not specify a side-effect, is it a converter? Maybe it'd be "Convert one thing into another"?

src/ansible_navigator/_version.py
cidrblock: done

src/ansible_navigator/_version.py
cidrblock: the others have a blank after, not before :)

cidrblock: put back

src/ansible_navigator/_yaml.py
cidrblock: updated

cidrblock:
recheck

webknjaz (approved):

PR #656, opened 2021-12-07
Add pytest-rerunfailures for tests seen as failing randomly

ssbarnea:


webknjaz (changes requested):
Let's instead use a more integrated solution: https://pypi.org/p/pytest-rerunfailures / https://github.com/pytest-dev/pytest-rerunfailures. I saw it being used in the pip's CI and it's trusted.

ssbarnea:
> Let's instead use a more integrated solution: [pypi.org/p/pytest-rerunfailures](https://pypi.org/p/pytest-rerunfailures) / [pytest-dev/pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures). I saw it being used in the pip's CI and it's trusted.

I checked it and you are welcomed to repurpose this PR and use that plugin instead, it seems as a maintained one.

webknjaz (approved):
This should be good now but need to wait for CI.

ssbarnea (commented):
I approve the new form but please do not merge until we have confirmation from zuul with 2-3 r-e-c-h-e-c-k-s

cidrblock:
Do we still need this?

ssbarnea:
I moved it to draft as to make it it clear that we need to wait until we have some tests randomly failing. Those that are in current PR are not longer doing this.

PR #658, opened 2021-12-07
Exclude change note fragments in markdownlint

webknjaz:


cidrblock (approved):

ssbarnea (changes requested):
Please do not add gate for changes that are not tested by our CI (either zuul or GHA). While using pre-commit.ci SaaS is good, it does not count as our CI and we know wheel that it runs with certain limitations that do not allow it run all kind of hook.

Also adding gate for changes when we already know unittests are broken is a very bad idea as the change will not be merged and only consume zuul time and our time, by slowing down testing of other changes.

We should focus on bringing zuul green and start merging these improvements only after that.

I am displeased as I about the fact that I keep seeing adding new changes that are not effectively testing and none that is addressing the broken unittests.

PR #659, opened 2021-12-07
Only run flake8 on manual invocations

webknjaz:
This cannot run in `pre-commit.ci` because it's too heavy.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,9 @@
 ---
 
+ci:
+  skip:
+  - flake8
+
 repos:
 - repo: https://github.com/asottile/add-trailing-comma.git
   rev: v2.0.1


cidrblock (approved):

ssbarnea (changes requested):
Can you please explain what is too heavy about it. That is the first case where I see flake8 being declared not compatible with pre-commit.ci and i am using it in like 40-50 repos for many years. They also happen to have the same maintainers...

Also if something does not work with pre-commit.ci, that is the incorrect way to skip it for that SaaS service.

webknjaz:
Looks like with stronger linter plugins it takes a lot of time, I haven't profiled exactly why but I think it may be a combination of the project complexity + a lot of linting plugins. This is being disabled only to allow for the rest of things to run. We are not going to use this for gating right now. Thanks for the config pointer, I improved it.

PR #660, opened 2021-12-07
Add a config for darglint

webknjaz:
This config explicitly specifies what is expected from the already
integrated linter plugin.

### DIFF
--- a/.darglint
+++ b/.darglint
@@ -1,0 +1,8 @@
+[darglint]
+# NOTE: All `darglint` styles except for `sphinx` hit ridiculously low
+# NOTE: performance on some of the in-project Python modules.
+# Refs:
+# * https://github.com/terrencepreilly/darglint/issues/186
+# * https://github.com/wemake-services/wemake-python-styleguide/issues/2287
+docstring_style = sphinx
+strictness = full


.darglint
```diff
@@ -0,0 +1,4 @@
+[darglint]
+docstring_style=sphinx
+enable=DAR104
+strictness=long
```
cidrblock, line 4: can we make this full?

cidrblock (approved):

.darglint
webknjaz: I guess we could but wouldn't it be too strict for the initial integration? We already have like a million violations. This would probably need a noqa in every place. Do you think this is sustainable at this point?

.darglint
```diff
@@ -0,0 +1,4 @@
+[darglint]
+docstring_style=sphinx
+enable=DAR104
```
webknjaz, line 3: @cidrblock remembering that this project uses type annotations, I'm leaning toward dropping this. WDYT?
```suggestion
```

.darglint
webknjaz: But yeah, we could totally do
```suggestion
strictness=full
```
if that's not a concern.

cidrblock: since we have all the per-file exceptions this should be fine right, and cover new files?

.darglint
cidrblock: yep makes sense

"DAR104: (disabled) The docstring parameter has no type specified"

I kind of wish it required that types in the docstring weren't present at all :)

cidrblock:
recheck

ssbarnea (dismissed):
I did a grep on`darglint` inside the repo and failed to find any place where we install and call it.

I am not against adding new linters but I need to see them running at the same time we add configuration for them, on CI.

Also the paragraph from https://github.com/terrencepreilly/darglint#project-status makes me think that darlint project is already a liability and that we may endup on a single person opinion.

Maybe I am missing something but I would really want to see when it runs and how it behaves.

webknjaz:
@ssbarnea it is a part of the wemake-python-stylegui.de linter set, this is why it's implicit. And I know that Brad uses it locally as a separate command invocation.

.darglint
webknjaz: I'm actually not fully sure but I'm not strongly opposed to accepting your suggestion either. Thought I'd mention this just in case so that it's explicit. The config is mostly copied from other projects that sometimes have different considerations.

.darglint
webknjaz: @cidrblock we could make a feature request and/or think of having our own linter plugin (it shouldn't be too hard to make one).

webknjaz (commented):
@cidrblock could you accept the suggested changes where you think it makes change and "resolve" the rest of the conversations? I don't have a strong preference, this is why I'm leaving this to you.

webknjaz:
@cidrblock @ssbarnea we need to merge this asap as it fixes the performance issue with darglint.

.darglint
```diff
@@ -0,0 +1,4 @@
+[darglint]
```
ssbarnea, line 1: Please add a comment on first line mentioning that flake8 is using this via wemake. So a random visitor will be able to identify how is this used.

webknjaz: Actually, I've checked that the issue is visible if you just invoke `darglint` directly. I wonder why Brad hasn't faced this locally.

cidrblock: I was only running darglint with select files and directories, so I probably just didn't hit it

webknjaz: Filed the upstream issue: https://github.com/terrencepreilly/darglint/issues/186.

webknjaz: The second issue: https://github.com/wemake-services/wemake-python-styleguide/issues/2287.

PR #664, opened 2021-12-07
Move global flake8 ignores to per-file

### LINKED ISSUE #662: [TODO] Convert broad flake8 ignores into per-file ones
2) move to _all_ file specific ignores

_Originally posted by @cidrblock in https://github.com/ansible/ansible-navigator/issues/655#issuecomment-988127575_

webknjaz:
Resolves #662

### DIFF
--- a/docs/changelog-fragments.d/662.internal.md
+++ b/docs/changelog-fragments.d/662.internal.md
@@ -1,0 +1,2 @@
+Started maintaining pre-existing flake8 exclusions per file
+-- by {user}`webknjaz`
--- a/docs/changelog-fragments.d/664.internal.md
+++ b/docs/changelog-fragments.d/664.internal.md
@@ -1,0 +1,1 @@
+662.internal.md
--- a/tox.ini
+++ b/tox.ini
@@ -184,105 +184,6 @@ show-source = true
 extend-ignore =
     E203,
     F401,
-    # FIXME:
-    C409,
-    C414,
-    Q000,
-    WPS100,
-    WPS110,
-    WPS111,
-    WPS112,
-    WPS114,
-    WPS115,
-    WPS118,
-    WPS121,
-    WPS122,
-    WPS201,
-    WPS202,
-    WPS204,
-    WPS210,
-    WPS211,
-    WPS212,
-    WPS213,
-    WPS214,
-    WPS218,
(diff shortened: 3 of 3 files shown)


ssbarnea (dismissed):
AGAIN, please stop creating changes that are not tested by any CI job. You know very well that we don't run lint on zuul or gha and still making changes to linters. I seen like 8-10 changes and none towards enabling linting execution in CI.

ssbarnea:
To be honest I a do not know how to vote on this change, one part of me is against one is in favour. 

If I approving it I know that I will regret it.  I would prefer to see changes that are resolving exceptions instead of playing around with their configuration, yep that is direct reference to the "leading by example" thing @cidrblock mentioned yesterday.

Basically this change would enforce people adding new code (files) a set of  rules that we do not respect in present almost at all. This mean imposing new standards on incoming pull requests, something that do not find ok.

Again, instead of doing this move, I would find far more valuable to even fix just one such exception. At least with that we could claim that our codebase is already following the rule.

In fact if someone would want to fix linting exceptions after this change is merged, it will be harder, mainly because you would have to remove the exception from 50-100 places instead of one just to get the linter to report all occurrences of it.

I will let @cidrblock decide on this and avoid adding a -1 on it.

webknjaz:
> I will let @cidrblock decide on this and avoid adding a -1 on it.

It's a part of the initial non-enforced adoption. This was specifically asked by Brad. It is hard to add excludes to every file so this may need to happen separately. Currently, this moves the ignores from a global location to a more specific one and that's it. Anything else is out of the scope here.

cidrblock:
So, if I understand correctly, this will bring attention to the errors found by some of the linters and checks that are being vetted in new files, but ignoring those in existing files.

So in theory, this PR should see those which should give us a chance to gauge the impact in a non-blocking fashion:

https://github.com/ansible/ansible-navigator/pull/624

and because I'm curious, over the next few days I will try and do all the error cleanup in one file as well as one error across the entire repo.  This might be a good way to see how one approach plays out vs the other. I don't really know which might be easier for the author and reviewers.

I'll approve this so we can see it run and see how it feels and collect more feedback on it's impact in other PRs

cidrblock (approved):

PR #653, opened 2021-12-06
Reinstate linters zuul job

ssbarnea:
Adds back lint zuul job as non-voting until we can fix all linting errors.

Related: https://github.com/ansible/ansible-navigator/pull/665

PR #655, opened 2021-12-07
Integrate linters through pre-commit.com tool

webknjaz:
This is a common approach to managing the linting setup in the wider
community that is also already used in some of our repositories.

Some of the checks are disabled and ignored to simplify the initial
adoption but must be addressed later.

P.S. Ignore the failing checks, they will disappear once #654 is merged.

PR #665, opened 2021-12-08
Fix linting regressions

ssbarnea:
Fixes problems causing flake8, black and mypy to fail.

This is making `tox -e linters,type` pass, both of them were failing before this change as we had multiple pull-requests merged without running any linting in CI.

These fixes were originally included into https://github.com/ansible/ansible-navigator/pull/653 but I was asked to make them separately, even if it that would not allow proper testing on CI.

### DIFF
--- a/docs/_ext/spelling_stub_ext.py
+++ b/docs/_ext/spelling_stub_ext.py
@@ -19,9 +19,9 @@ class SpellingNoOpDirective(SphinxDirective):
 
 def setup(app: Sphinx) -> None:
     """Initialize the extension."""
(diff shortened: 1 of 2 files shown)


ssbarnea:
Please ignore pre-commit.ci failure as that is unrelated to my change and was caused by the premature merge of https://github.com/ansible/ansible-navigator/pull/655

cidrblock (approved):
ty sorin!

webknjaz:
Sounds like this "fix" is invalid, though.

PR #666, opened 2021-12-08
Fix typing in `setup` of Sphinx spelling stub ext

webknjaz:


webknjaz:
This PR does not attempt to fix the violation with `human_time()` because https://github.com/ansible/ansible-navigator/pull/641 will do this properly.

ssbarnea (approved):
LGTM, if we no longer see the error on pre-commit.ci, we merge it.

webknjaz:
> 

I think there will still be an error with one function that is going to be addressed in another PR.

ssbarnea:
> I think there will still be an error with one function that is going to be addressed in another PR.

Are you doing to address this error by any chance? I am asking because I find it bit funny that this error happens to raise from inside the "externally managed" file. If that is externally managed, I find hard to believe that the undocumented place where it comes from does not use mypy to check typing problems, especially as the file looks like being typed.

webknjaz:
What are you talking about? What I mentioned was not related.

PR #667, opened 2021-12-08
Fix the externally managed file by partialy reverting invalid "Fix linting regressions (#665)"

webknjaz:
This fixes an invalid formatting change introduced by #665.

### DIFF
--- a/docs/_ext/spelling_stub_ext.py
+++ b/docs/_ext/spelling_stub_ext.py
@@ -19,9 +19,9 @@ class SpellingNoOpDirective(SphinxDirective):
 
 def setup(app: Sphinx) -> Dict[str, Union[bool, str]]:
     """Initialize the extension."""
-    app.add_directive("spelling", SpellingNoOpDirective)
+    app.add_directive('spelling', SpellingNoOpDirective)
 
     return {
-        "parallel_read_safe": True,
-        "version": "builtin",
+        'parallel_read_safe': True,
+        'version': 'builtin',
     }
--- a/docs/_ext/spelling_stub_ext.py
+++ b/docs/_ext/spelling_stub_ext.py
@@ -1,3 +1,4 @@
+# fmt: off
 """Sphinx extension for making the spelling directive noop."""
 
 from typing import Dict, List, Union


cidrblock (approved):

ssbarnea:
Can we please wait a little bit on changes?

That file was added by @webknjaz recently and was already not passing the linting. Basically another case where the review process was poor because it failed to spot the fact that it degraded the linting. 

This should have being altering the linter config to avoid causing an error or just fix it in original change. If we are about to start reverting thinks, why not reverting the original addition of the file?

webknjaz:
@ssbarnea this is related to the docs infra and doesn't break anything related to the linting. Removing it would be breaking, though. Please don't block this change as it's a non-breaking improvement over the previous state. Plus it reverts your change that shouldn't have been merged in the first place, it wasn't reviewed by people who understand the implications.

ssbarnea:
Instead of doing this revert which would break black please make a constructive one that also adds this file to the exclusion list -- in case that is indeed an externally tracked file (does not look really as a true one).

webknjaz:
> Instead of doing this revert which would break black please make a constructive one that also adds this file to the exclusion list -- in case that is indeed an externally tracked file (does not look really as a true one).

It is tracked in Cheroot and a few other projects. Also, I don't use black and don't know about its configuration so I'll keep it as-is for now, black is really not related, I just don't want it to be in my way making bad changes.

ssbarnea (changes requested):
I its current form this change breaks `tox -e linters`. If we are to restore the style used by these documents this needs to include alteration of linting config to skip processing these files.

As @webknjaz claims that these are externally manged files, I would raise the question of where? The `_ext` in the folder name is not from external but from extension, as they are small sphinx extensions.

There is no README mentioning that these files are externally managed so do not see them as such.

Without any documentation on this we cannot really blame anyone for fixing their string form to bring them in-sync with the rest of the project.

I really doubt we will need to do much maintenance on these files, so I do not even think it worth the effort of excluding them from black.

webknjaz:
> I its current form this change breaks `tox -e linters`.

Not true.

> If we are to restore the style used by these documents this needs to include alteration of linting config to skip processing these files.

It's excluded already as it is.

> As @webknjaz claims that these are externally manged files, I would raise the question of where? The `_ext` in the folder name is not from external but from extension, as they are small sphinx extensions.

It's a part of an incomplete Sphinx integration that is to be added in a follow-up. It allows using a Sphinx directive with the respective name as no-op when the dependency with an actual implementation is missing.

> There is no README mentioning that these files are externally managed so do not see them as such.

True, I didn't add any.

> Without any documentation on this we cannot really blame anyone for fixing their string form to bring them in-sync with the rest of the project.

True, I didn't document this in the source code but it's beside the point as I'm restoring this how it should be now. And since it does not produce *any* linting violations, there is no technical reason not to merge this.

> I really doubt we will need to do much maintenance on these files, so I do not even think it worth the effort of excluding them from black.

This is why it's marked in file.

PR #670, opened 2021-12-08
[type] Ignore attr defined

cidrblock:
This was a 3.6 specific mypy complaint

### DIFF
--- a/src/ansible_navigator/actions/_actions.py
+++ b/src/ansible_navigator/actions/_actions.py
@@ -35,6 +35,7 @@ def _import(package: str, action: str) -> None:
 
 def _import_all(package: str) -> None:
     """Import all actions in a package"""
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

src/ansible_navigator/actions/_actions.py
```diff
@@ -35,7 +35,7 @@ def _import(package: str, action: str) -> None:
 
 def _import_all(package: str) -> None:
     """Import all actions in a package"""
-    files = resources.contents(package)
+    files = resources.contents(package)  # type: ignore[attr-defined]
```
webknjaz, line 38: This needs a code comment with the motivation.

webknjaz (approved):

ssbarnea:
We bypassed zuul to save time on this as we knew it was correct and wanted to avoid the rebase.

PR #673, opened 2021-12-08
[lint] long line

cidrblock:
A long line snuck in from another PR


cidrblock:
In an effort to see #671 go in, this should fix the final and remaining lint error.  Forced in the interest of time.

PR #677, opened 2021-12-09
Add tox-ini-fmt pre-commit hook (wip)

ssbarnea:
In order to avoid random indentation and ordering of tox.ini options, we now use tox-ini-fmt to ensure a consistent content.

[tox-ini-fmt](https://github.com/tox-dev/tox-ini-fmt) was created by the current maintainer of tox so you cannot get more official than that. The main benefit is that we no longer need to manually correct human indentation mistakes, it will fix them for us.

As clearly documented, non tox sections are not touched, only moved to the end of file. In fact use of non tox sections inside tox.ini file is discouraged and something we should look into sort in a follow-up.

Note that the content of that `tox.ini` was the result of running the hook, I did not make any other manual change to it. Still, I find the outcome a net improvement over the old one.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,11 @@
 ---
 
 repos:
+- repo: https://github.com/tox-dev/tox-ini-fmt
+  rev: "0.5.1"
+  hooks:
+    - id: tox-ini-fmt
+      args: ["-p", "linters,type"]
 - repo: https://github.com/asottile/add-trailing-comma.git
   rev: v2.0.1
   hooks:
--- a/tox.ini
+++ b/tox.ini
@@ -1,190 +1,175 @@
-[base]
-pkg_name = ansible_navigator
-default_ee = quay.io/ansible/creator-ee:v0.2.0
-small_test_ee = quay.io/ansible/python-base
-mypy_tmp_dir = mypy_tmp_dir
-
 [tox]
-envlist = linters, type, py38, report, clean
-minversion = 3.16.1
-skipsdist = True
+envlist =
+    linters
+    type
+    py38
+    report
+    clean
+skipsdist = true
 skip_missing_interpreters = true
-
+minversion = 3.16.1
 
 [testenv]
 description = Run pytest under {basepython} ({envpython})
-# the pytest command line is not in the project.toml because of issues 
-# with catching breakpoints while debugging unit tests
-# the number of CPUs is set to 50% b/c CI timeout failures were more likely
-# the small test image is used an alternative to the default, but should be it
-# one of it's layer
-allowlist_externals = cat
-                      rm
-                      grep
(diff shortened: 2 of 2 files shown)


webknjaz (changes requested):
:-1: because this moves things out of place and deletes important comments. It also makes it harder to sync things with the projects the parts of the config came from.

ssbarnea:
Ouch, I did not see https://github.com/tox-dev/tox-ini-fmt/issues/42 before raising it. Clearly a deal breaker in adopting it. I will move it to draft.

webknjaz:
> Ouch, I did not see [tox-dev/tox-ini-fmt#42](https://github.com/tox-dev/tox-ini-fmt/issues/42) before raising it. Clearly a deal breaker in adopting it. I will move it to draft.

This is not the only concern. Another one is that the reusable `[base]` moved and it should be at the top because it's natural to see the common definitions and have then reused below (this doesn't really differ from any source code in this regard).

The formatted also deleted extra empty line separators between multiline commands which is also a deal breaker. The same goes for the inability to have semantical sorting of the sections (like the docs-related sections being next to each other, sorted by importance).
Plus I currently sort the settings alphabetically, which helps with synching the file across projects. If you randomly shuffle the order like this, it will make it very hard to match different sections with `nvim -d`, for example — who is going to be responsible for maintaining close copies then?

ssbarnea:
The idea of sync with another project makes very little sense for few reasons:
* there is not reference source project to sync with 
* you cannot control divergence and in time setups will diverge
* a computer sorted ini file is far more likely to be easy to sync across projects as it will always put the same options in the same place

Still, the **comments issue make this impossible to adopt**. Likely we will have to manually do some fixes to the file to make it easier to maintain, if and when we have time for that.

webknjaz:
> The idea of sync with another project makes very little sense for few reasons:
> 
> * there is not reference source project to sync with

Yeah, I didn't get to it, it's rather manual at this point. But the hope is to have something similar to https://github.com/jaraco/skeleton (partially, at least).

> * you cannot control divergence and in time setups will diverge

Yes, but with the approach linked above, it's possible to make the merge/update conflicts smaller and easy to solve. For example, structuring entries in a way that would make a part of the file identical (coming from "upstream") and another part that is project-specific. This would be something to deal with but it'll show up in the same place of the diff and not scattered across the file and combined with other changes.

> * a computer sorted ini file is far more likely to be easy to sync across projects as it will always put the same options in the same place

Yes, but the drawbacks of not having control over how it's formatted influence other important aspects that we shouldn't forget about. FWIW I'm willing to sync things manually for as long as we don't have an automatic solution that is truly acceptable.

> Still, the **comments issue make this impossible to adopt**. Likely we will have to manually do some fixes to the file to make it easier to maintain, if and when we have time for that.

I'll make sure to send a PR adjusting this indentation soon then. Leave it to me.

PR #678, opened 2021-12-09
Ensure flake8 and yamllint run from pre-commit

ssbarnea:
Instead of running flake8 and yamllint in two different ways, we run pre-commit from tox -e linters. This allows
us avoid surprises cause by different tool versions.

The plan is to do the same for the other linters in follow-up.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -49,10 +49,9 @@ commands =
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
+  pre-commit run -a {posargs}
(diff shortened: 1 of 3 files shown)


cidrblock (approved):

webknjaz (approved):

tox.ini
```diff
@@ -49,10 +49,9 @@ commands =
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
+  pre-commit run -a {posargs}
```
webknjaz, line 52: This was a mistake, it should've included `flake8` in the args.

PR #679, opened 2021-12-09
Fixed trailing whitespace

ssbarnea:
Resolve existing trailing whitespaces and prevents introduction of new ones.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -40,16 +40,6 @@ repos:
   hooks:
   # Side-effects:
   - id: trailing-whitespace
-    exclude: |
-      (?x)
-      ^
-        tests/fixtures/unit/configuration_subsystem/ansible-navigator\.yml|
-        licenses/LICENSE\.source\.json\.json|
-        tests/smoketests/scenarios/smoke-config\.yml|
-        docs/faq\.md|
-        tests/fixtures/integration/execution_environment/ansible-navigator_disable_ee\.yml|
-        tox\.ini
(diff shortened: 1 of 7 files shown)


cidrblock (approved):

licenses/LICENSE.source.json.json
```diff
@@ -7,13 +7,13 @@ All rights reserved.
 MIT License
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
-documentation files (the ""Software""), to deal in the Software without restriction, including without limitation 
-the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, 
+documentation files (the ""Software""), to deal in the Software without restriction, including without limitation
```
webknjaz, line 10: Is it safe to edit this file? I thought it has something to do with JSON format...

webknjaz (approved):
This looks great, thanks! I only have one small concern that I've started a thread for. If it's not, please comment why you think so and mark it as "Resolved". Besides that, this is ready to merge.

licenses/LICENSE.source.json.json
ssbarnea: Now that I see the double quote on that file I realise that this whitespace may be needed. Time to use git blame to find who added these and what is their purpose...

cidrblock: never intended to be valid json files, the extention got carried over from the source file name. 

these were renamed in #682

PR #683, opened 2021-12-09
Move the flake8 config into a standalone location

webknjaz:
The motivation here is to improve the maintainability and follow the wider community practices. It also allows to find out more easily where the common flake8 configuration exists rather than trying to check all of the config files where it may be present.
As a bonus, this will allow comparing/syncing flake8 configurations across projects that also follow this practice.


ssbarnea:
@webknjaz I am ready to +1 this change but it should pass pre-commit.ci. Now I see some flake8 errors reported which is not ok.

webknjaz:
> @webknjaz I am ready to +1 this change but it should pass pre-commit.ci. Now I see some flake8 errors reported which is not ok.

Yeah, I know. I was about to tell you about this because I don't fully understand *why* this is happening. I just moved things from one file to another w/o any changes. I will try to debug what's going on.

cidrblock (approved):
assuming it will pass before merged

PR #685, opened 2021-12-09
Fix end-of-file

ssbarnea:


cidrblock:
some of those files, specifically the json files in the fixtures directory are generated by the tests themselves.

specifically here: https://github.com/ansible/ansible-navigator/blob/main/tests/integration/_common.py#L45

triggered by a temporary update to the test or the use on an env var

https://github.com/ansible/ansible-navigator/blob/main/tests/integration/actions/config/base.py#L73 

So........  if we update them we should probably have a followup PR/issue to not reintroduce them when fixtures are updated or created

ssbarnea:
I was expecting that and to have to make other changes. Thanks for pointing me the right place to change.

I will update the PR to do it right from start.

cidrblock (approved):

PR #688, opened 2021-12-09
[tests] Update all fixtures

cidrblock:
This is a full update of all fixtures
- added a newline to each (this is the main reason for the PR) (see the _common.py file for the code change)
- other fixture updates expected:

1. image name update
2. uuids
3. py env name

beacuse some fixtures were missing, but have been added (not required by the specific tests), some new spelling errors have been introduced.  Will address in a different PR

This should limit the changes in #685 while tying the fixture updates to the code change that added the newline.


cidrblock:
If this is the first time seeing the pre-commit stuff, although it has failed, it does not block the PR/gate.....

ssbarnea (approved):

PR #691, opened 2021-12-09
Add lint-vetting job for all optional linters

ssbarnea:
- ensure that all optional pre-commit hooks are run with `manual`
- use `lint-vetting` as tox env name and job name for the run  linting with all optional hooks
- add non-voting zuul job named `lint-veting`
- normal run of `pre-commit.ci` will run only required/approved hooks, making its result relevant for the review process

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
       ^
         .*
       $
+    stages: ["manual"]
 
 - repo: https://github.com/timothycrosley/isort.git
   rev: 5.10.1
@@ -24,6 +25,7 @@ repos:
       ^
         .*
       $
+    stages: ["manual"]
 
 - repo: https://github.com/Lucas-C/pre-commit-hooks.git
   rev: v1.1.10
@@ -47,12 +49,14 @@ repos:
       ^
         .*
       $
+    stages: ["manual"]
   - id: end-of-file-fixer
     exclude: |
       (?x)
       ^
         .*
       $
+    stages: ["manual"]
   - id: requirements-txt-fixer
     exclude: >-
       ^requirements\.txt$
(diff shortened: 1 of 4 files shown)


cidrblock (approved):

.pre-commit-config.yaml
```diff
@@ -219,8 +226,6 @@ repos:
     language_version: python3
     additional_dependencies:
     - flake8-2020 >= 1.6.0
-    - flake8-docstrings >= 1.5.0
-    - flake8-pytest-style >= 1.2.2
```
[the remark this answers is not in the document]
cidrblock: we can introduce them one at a time with a consensus from the team

webknjaz (commented):
This change is upsetting, it breaks the easy synchronization cross projects :(

.pre-commit-config.yaml
```diff
@@ -12,6 +12,7 @@ repos:
       ^
         .*
       $
+    stages: ["manual"]
```
webknjaz, line 15: This is rather pointless since everything is excluded already.

ssbarnea: Request to do it like this came from @cidrblock 

We will move them out of manual one by one using reviews that involve entire team and use the new label "quorum-required". It is about getting approval for any new rule. It will require to fix exceptions/exclusions before doing that or at the same time. Basically anything optional is manual and rub by the lint-vetting non-voting job.

webknjaz: This also does has an inconsistent style with the rest of the places in the file.

.pre-commit-config.yaml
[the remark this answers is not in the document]
webknjaz: I think that at least `flake8-docstrings` is a part of WPS

PR #695, opened 2021-12-10
[docs] Clarify the current state of `black`

cidrblock:


.github/CONTRIBUTING.md
```diff
@@ -10,7 +10,11 @@ maintain it and you contribute.
 
 2. pylint (lints all the things)
 
-3. black (opinionated formatter)
+.. note:: In early development cycles, a decision was made to use black as a formatter
+          which is why current pull-requests are required to pass a
+         `black --diff` check of the source tree. The decision to use `black` is
+         left to individual developers as the formatting changes it makes can be
+         achieved without it.
```
webknjaz, line 17: There are two problems with this. The first one is that typically in RST/Sphinx the prose goes below the directive which allows to make better use of the space:
```suggestion
.. note::

   In early development cycles, a decision was made to use black as a formatter
   which is why current pull-requests are required to pass a
   `black --diff` check of the source tree. The decision to use `black` is
   left to individual developers as the formatting changes it makes can be
   achieved without it.
```
The second problem is that you're adding RST inside Markdown which won't work. You could use MyST syntax to make it work within Sphinx, at least, though:
````suggestion
```{note}
In early development cycles, a decision was made to use black as a formatter
which is why current pull-requests are required to pass a
`black --diff` check of the source tree. The decision to use `black` is
left to individual developers as the formatting changes it makes can be
achieved without it.
```
````

Comment (cidrblock) [reply .github/CONTRIBUTING.md]: awesome, thanks! I totally forgot about MyST part of it

File: .github/CONTRIBUTING.md [hunk 2]
```diff
@@ -10,7 +10,13 @@ maintain it and you contribute.
 
 2. pylint (lints all the things)
 
-3. black (opinionated formatter)
```
ssbarnea, line 13: TBH, I found pre-commit black integration with auto-fixable problems automatically more convenient. You avoid wasting human time on fixing formatting.

I see the current fail with diff implementation in navigator and seen a PITA to manually implement its recommendations in order to make it pass.

I love the fact that black removes some of the nitpicking discussions that I seen over reviews in various projects. If a computer can do it, let it do it..We can focus on code logic instead of imposing our artistic preferences over the incomping reviewer.

webknjaz: On the other hand, it adds more problems for me than it solves. It stands in the way of productivity. It hides serious problems indicated by linters. I would prefer black not to block my work, especially when it's unable to produce code better than me because it lacks the context and cannot grasp what is being communicated through properly structured code.

webknjaz (approved):

.github/CONTRIBUTING.md
cidrblock: This doesn't remove the current black check, it simply changes the verbage of the contributors guide such that it explains black is in place, but developers are not required to use it to satisfy that check. Let's do a fresh PR at some point to discuss whether or not the check stays in place. We have a process for adding linters, no reason we can't use the same process for removing a formatting tool that some may or may not choose to satisfy linters.

Issue #693, opened 2021-12-09
[tests] Allow `run` integration tests to create fixtures


cidrblock:
https://github.com/ansible/ansible-navigator/blob/f4c8d669f75ee68717dcfd890a0a0820bee90e30/tests/integration/actions/run/base.py#L93

This will be consistent with other tests.

Noticed in #685 even after #688 was run and merged

PR #696, opened 2021-12-10
[tests] Preset condition var

cidrblock:
as suggested in #624 reduce complexity by presetting the var used in the condition.

Additionally, correctly set the pylint: disable statements for this function across tests, removing unneeded and adding new.

(this can be cleaned up more, #693 might be a good place to do that)

### DIFF
--- a/tests/integration/actions/collections/base.py
+++ b/tests/integration/actions/collections/base.py
@@ -81,11 +81,13 @@ class BaseClass:
             line.replace(os_indendent_tmp, "FIXTURES_COLLECTION_DIR") for line in received_output
         ]
 
-        if (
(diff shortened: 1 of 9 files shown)


tests/integration/actions/collections/base.py
```diff
@@ -58,6 +58,8 @@ def test(
         comment,
     ):
         # pylint: disable=too-many-arguments
+        # pylint: disable=too-many-locals
+
```
webknjaz, line 62: Based on the style used in other places, this extra empty line is accidental + it violates PEP 257 so why not drop it
```suggestion
```

webknjaz (approved):
@cidrblock good to gate

PR #706, opened 2021-12-11
Sync annotated flake8 config with other projects

webknjaz:
This patch synchronized explicit annotated flake8 config with other projects I maintain or contribute to (including ones under Ansible). It specifically attempts not to change the current linting settings but make the forgotten settings explicit and documented.


.flake8
```diff
@@ -1,19 +1,59 @@
...
+  # Metadata of `pip wheel` cmd is autogenerated
+  pip-wheel-metadata,
+  # adjacent venv
   venv
+
+# IMPORTANT: avoid using ignore option, always use extend-ignore instead
+# Completely and unconditionally ignore the following errors:
+extend-ignore =
+  E203,  # skipped because it annoys black
+  F401,  # skipped for unknown reason
```
cidrblock, line 41: F401 module imported but unused

Since we don't know why it is skipped, and we have no violations, and it _seems_ sane, Should it be enabled, maybe not this PR because I think this is about the format of the file not removals/additions

.flake8
```diff
@@ -1,19 +1,59 @@
...
+  WPS300,  # "Found local folder import" -- nothing bad about this
+  WPS306,  # "Found class without a base class: *" -- irrelevant under Python 3
```
cidrblock, line 43: From the wemake docs:

Please, note that this rule has nothing to do with python2. We care only about consistency here, I like it here but I think the description for the exception could be better

There is a relate discussion about this one in the Discussions

.flake8
```diff
@@ -1,19 +1,59 @@
...
+  WPS422,  # "Found future import: *" -- we need these for multipython
```
cidrblock, line 44:
```
Forbid __future__ imports.

Reasoning:
Almost all __future__ imports are legacy python2 compatibility tools that are no longer required.

Solution:
Remove them. Drop python2 support.
```

The only place we violate this is and probably will be the tests fixtures directory because it may contain ansible content which was/is subject to a different rule set.  Can we limit this to just test/fixtures with an explanation?

.flake8
```diff
@@ -1,19 +1,59 @@
...
+  dist,
+  # Metadata of `pip wheel` cmd is autogenerated
+  pip-wheel-metadata,
+  # adjacent venv
   venv
+
+# IMPORTANT: avoid using ignore option, always use extend-ignore instead
+# Completely and unconditionally ignore the following errors:
+extend-ignore =
+  E203,  # skipped because it annoys black
```
cidrblock, line 40: Can we link to this: https://github.com/psf/black/issues/315 , since it explains "how" the 2 conflict?  (or maybe a lengthy reason for the exclusion)

.flake8
```diff
@@ -1,19 +1,59 @@
...
+
 builtins = _
+
+# Print the total number of errors:
+count = true
+
+# Don't even try to analyze these:
 extend-exclude =
-  # skip tool cache dirs
+  # No need to traverse egg info dir
```
cidrblock, line 10: I see there in some attempt to order these in a particular fashion?  Can that be explained so future addition here respect the ordering?

.flake8
```diff
@@ -1,19 +1,59 @@
...
+# IMPORTANT: avoid using ignore option, always use extend-ignore instead
+# Completely and unconditionally ignore the following errors:
+extend-ignore =
+  E203,  # skipped because it annoys black
+  F401,  # skipped for unknown reason
+  WPS300,  # "Found local folder import" -- nothing bad about this
+  WPS306,  # "Found class without a base class: *" -- irrelevant under Python 3
+  WPS422,  # "Found future import: *" -- we need these for multipython
+# https://wemake-python-stylegui.de/en/latest/pages/usage/formatter.html
+#format = wemake
```
cidrblock, line 46: This is commented out without explanation, can it be removed or a note added why it is currently commented out?

.flake8
```diff
@@ -1,19 +1,59 @@
...
+
+# Let's not overcomplicate the code:
+#max-complexity = 10
```
cidrblock, line 49: Same as above, commented without explanation

.flake8
```diff
@@ -1,19 +1,59 @@
...
+
+# Accessibility/large fonts and PEP8 friendly:
+#max-line-length = 79
```
cidrblock, line 52: This seems like a duplicate of the next, can it be removed?

.flake8
```diff
@@ -1,19 +1,59 @@
...
+# Accessibility/large fonts and PEP8 unfriendly:
+max-line-length = 100
+
+# Allow certain violations in certain files:
```
cidrblock, line 56: Can we add a note in the global section that:

1) exceptions should be fixed
2) if not, exception is in the file, inline
3) if not possible for some reason, exception can go here, but a majority of these are temporary
4) ACD of global needs team review (not today but in the future)

.flake8
```diff
@@ -227,3 +267,42 @@ per-file-ignores =
   tests/unit/image_manager/__init__.py: D104
   tests/unit/__init__.py: D104
   share/ansible_navigator/utils/__init__.py: D104
+
+# Count the number of occurrences of each error/warning code and print a report:
+statistics = true
+
+# flake8-eradicate
+# E800:
+eradicate-whitelist-extend = isort:\s+\w+
```
cidrblock, line 276: Since we are not using this, can this be removed?

.flake8
```diff
@@ -227,3 +267,42 @@ per-file-ignores =
...
+
+# flake8-pytest-style
+# PT001:
```
cidrblock, line 279: We appear to only have one violation of this, is it necessary as global now?

.flake8
```diff
@@ -227,3 +267,42 @@ per-file-ignores =
...
+pytest-fixture-no-parentheses = true
+# PT006:
+pytest-parametrize-names-type = tuple
```
cidrblock, line 282: This is also exempted on a per-file basis, can we have one or the other? The two ways to have the exception seem to conflict, maybe update the per-file exception in another PR?

.flake8
```diff
@@ -227,3 +267,42 @@ per-file-ignores =
...
+# PT007:
```
cidrblock, line 283: We currently do not have any violations of this, can we reconsider when we do or it comes up?

.flake8
webknjaz: Yep, exactly. I tried to produce minimum influence here but in follow-ups, we should definitely refine what's ignored globally.

.flake8
webknjaz: Ah, interesting. I just copied this from other projects where I disabled it because I don't agree with the sentiment. Consistency is good but I've never seen any of the upstream projects inheriting from `object` in every base class. As for Ansible projects, in core and collections `ansible-test` specifically has a `__metaclass__ = type` that makes py2 classes new-style enforcing this behavior in runtime.

.flake8
cidrblock: So this is not forgotten, can you log an issue with a link to this?

.flake8
webknjaz: Not necessarily. There are some of the future imports that are related to newer pythons. Like `from __future__ import annotations`.

webknjaz: Oh, wait, there's a list of allowed ones: https://wemake-python-stylegui.de/en/latest/pages/api/constants.html#wemake_python_styleguide.constants.FUTURE_IMPORTS_WHITELIST.

webknjaz: In general, I'm in favor of only ignoring this in the test fixtures. The problem is that I've seen flake8 not merging several ignores that are listed in specific file paths and matching wildcards. Basically, if you have two entries like `tests/test_file.py` and `tests/test_*.py` only one will be taken into account by flake8. Not sure if it's a bug or by design but I remember this confused me and I couldn't find a proper solution for the problem.

.flake8
webknjaz: I guess, I tried not to change the previous contents too much, leaving that for follow-ups.

.flake8
webknjaz: I guess. Since I'm synchronizing this across many projects, this is a kind of thing that would need to go into the comment above the setting and be copied over into many places. The order is alphabetical so that it produces the lowest possible amount of conflicts.

.flake8
webknjaz: This is enabled in some of the projects but not the others. The reason is that it sometimes breaks the output and I haven't yet figured why. In general, it's best to have it and re-enable once fixed + fewer differences between configs means that it's easier to sync. I've recently re-discovered one flake8 setting that may make this even easier, allowing to split the configs. But that needs to be researched first.

.flake8
webknjaz: No, I don't want this to diverge too much. Plus it's here as a demonstration of what value could be used to make the accessibility problem go away. Discussion TBA.

.flake8
webknjaz: The explanation is that I want to keep the config close to others where this is uncommented. It should be uncommented at some point but I didn't enable it because it's a new rule that was disabled previously and hasn't undergone the acceptance discussion.

.flake8
webknjaz: Such comments should be present below, once the entries are separated into what needs to be fixed vs what needs to remain ignored. But this is not the part of this PR because its sole purpose is to sync the config structure.
If we were to improve the annotations above the settings, this would be separate and synched back into other projects with a similar baseline configuration. Also, what do you mean by global? At the top of the config file?

.flake8
webknjaz: I'd rather keep it for the sake of having the configs similar. Also, this plugin is a part of WPS. So technically we are using it. We don't have isort enabled, that is true. But my feeling is that we may enable it at some point and if we end up needing its marker comments, they'll likely produce false positives in `erradicate`. Since this problem has been debugged and sorted with this config in other projects, I don't see why we shouldn't include it and face this false-positive that would definitely confuse the contributors and waste somebody's debug time.

.flake8
webknjaz: I don't know, I just synced the config with known good settings.

.flake8
webknjaz: It's not an exception. Some of the flake8 plugins have their own configurations. They are per-rule, not per-file. This setting is just a baseline for how the plugin is supposed to work.

.flake8
webknjaz: This is not an ignore, it's a plugin setting.

.flake8
cidrblock: I have no issue with it being an exception, the explanation is misleading.....

.flake8
cidrblock: No experience there, if the change is outside the scope of this PR can we log an issue to resolve with a follow up PR?

.flake8
cidrblock: NP, can we log an issue to have it updated in a follow up PR so it's not forgotten?

.flake8
cidrblock: IMO, if we expect future contributors to keep in alphabetical order it should be noted.  If not changed in this PR can you log an issue for follow-up so it's not forgotten?

.flake8
cidrblock: yeah, sorry, someplace obvious that would indicate that changes to the file are generally not necessary for exceptions as explained above.  

If not in this PR, yeah you guessed it, a followup issue & PR?

.flake8
cidrblock: I understand, can we note that it's existence in the file for forward looking to a future implementation and not in effect today?  If not in this PR, can you log an issue for followup?

.flake8
cidrblock: Can we log an issue to remove it later if it falls outside the scope of this PR so it's not fogotten?

.flake8
cidrblock: got it!

.flake8
cidrblock: got it!

.flake8
webknjaz: These settings are explicit because they differ from the given plugin defaults. I don't see how an issue would help. Do you mean that we need an issue to delete the settings if we decide that we don't need the consistency in the tests to be linted?

.flake8
webknjaz: Yep, basically, there are two styles. And one is chosen by default making the other produce a violation of the given linter rule. When the setting is changed, the violation is raised in the opposite case.

.flake8
webknjaz: https://github.com/ansible/ansible-navigator/issues/720

.flake8
webknjaz: I can mention that a different kind of consistency is currently in place: https://github.com/ansible/ansible-navigator/discussions/700#discussioncomment-1791139.

.flake8
webknjaz: I'll just drop this for now.

.flake8
webknjaz: https://github.com/ansible/ansible-navigator/issues/721

.flake8
webknjaz: https://github.com/ansible/ansible-navigator/issues/722

.flake8
webknjaz: https://github.com/ansible/ansible-navigator/issues/723

cidrblock (approved):

PR #707, opened 2021-12-11
Fixed various spelling errors

ssbarnea:
There were reported by codespell, but we will need a couple of separated changes to address all of them, as we for example even have a spelling mistake in a module name. These are safe to merge as is.

### DIFF
--- a/docs/_ext/single_sourced_data.py
+++ b/docs/_ext/single_sourced_data.py
@@ -101,7 +101,7 @@ def _rst_generate_row(row: Tuple) -> List:
 
 
 def _params_generate_tables(param_details: Dict) -> List:
-    """generate tables for paramters"""
+    """generate tables for parameters"""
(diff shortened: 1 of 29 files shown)


docs/_ext/single_sourced_data.py
```diff
@@ -101,7 +101,7 @@ def _rst_generate_row(row: Tuple) -> List:
 
 
 def _params_generate_tables(param_details: Dict) -> List:
-    """generate tables for paramters"""
+    """generate tables for parameters"""
```
webknjaz, line 104: Note for myself: need to synchronize this back to projects this file came from.

webknjaz (approved):
Thanks for taking time to do this!

cidrblock (commented):
awesome, maybe I need a new keyboard :)

PR #708, opened 2021-12-11
Use tox lint job and document it

ssbarnea:
- use 'lint' instead of 'linters' for both tox and zuul jobs, so
  they are consistent across all our projects. zuul job name is
  already named `ansible-navigator-tox-lint` so it does not need change.
- remove leftover type from envlist in tox as it was removed, as is
  part of lint.
- remove output lines from lint command that are not only outdated
  but also irrelevant for docs inclusion.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -39,15 +39,15 @@ Prerequisites:
 
 2. Use {doc}`tox <tox:index>` to run the tests.
 
-3. Before sending a PR, make sure that the linters pass:
+3. Before sending a PR, make sure that `lint` passes:
 
    ```shell-session
-   $ tox -e linters
-   linters create: .tox/linters
-   linters installdeps: -rrequirements.txt, -rtest-requirements.txt
-   linters installed: ...
-   linters run-test-pre: PYTHONHASHSEED='4242713142'
-   linters run-test: commands[0] | pylint ansible_navigator tests ...
+   $ tox -e lint
+   lint create: .tox/linters
+   lint installdeps: -rrequirements.txt, -rtest-requirements.txt
+   lint installed: ...
+   lint run-test-pre: PYTHONHASHSEED='4242713142'
+   lint run-test: commands[0] | pylint ansible_navigator tests ...
    ...
    ```
 
@@ -56,12 +56,12 @@ Prerequisites:
    annoyingly slow):
 
    ```shell-session
-   $ tox -e lint
-   lint create: .tox/lint
-   lint installdeps: pre-commit, pylint ~= 2.8.0, pylint-pytest < 1.1.0
-   lint installed: ...
-   lint run-test-pre: PYTHONHASHSEED='2351399476'
-   lint run-test: commands[0] | python -m pre_commit run ... --all-files -v
+   $ tox -e lint-vetting
+   lint-vetting create: .tox/lint-vetting
+   lint-vetting installdeps: pre-commit, pylint ~= 2.8.0, pylint-pytest < 1.1.0
+   lint-vetting installed: ...
+   lint-vetting run-test-pre: PYTHONHASHSEED='2351399476'
+   lint-vetting run-test: commands[0] | python -m pre_commit run ... --all-files -v
    ...
    [INFO] Initializing environment for https://github.com/asottile/add-trailing-comma.git.
    ...
--- a/docs/changelog-fragments.d/708.internal.md
+++ b/docs/changelog-fragments.d/708.internal.md
@@ -1,0 +1,3 @@
+Updated the linting tox environment names in the contributing document to
+match the configuration changes in the repository. It now mentions separate
+commands for `lint` and `lint-vetting` suites -- by {user}`ssbarnea`
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,2 @@
+MD013:
+  code_blocks: false
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -53,7 +53,7 @@
     name: ansible-navigator-tox-lint
     parent: ansible-tox-py38
     vars:
-      tox_envlist: linters
+      tox_envlist: lint
       tox_extra_args: -vv --skip-missing-interpreters false
 
 - job:
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ default_ee = quay.io/ansible/creator-ee:v0.2.0
 small_test_ee = quay.io/ansible/python-base
 
 [tox]
-envlist = linters, type, py38, report, clean
+envlist = lint, py38, report, clean
 minversion = 3.16.1
 skipsdist = True
 skip_missing_interpreters = true
@@ -43,7 +43,7 @@ skip_install = true
 commands =
   coverage erase
 
-[testenv:linters]
+[testenv:lint]
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =


cidrblock (approved):
looks fine from here

.github/CONTRIBUTING.md
```diff
@@ -42,12 +42,7 @@ Prerequisites:
 3. Before sending a PR, make sure that the linters pass:
 
    ```shell-session
-   $ tox -e linters
-   linters create: .tox/linters
```
webknjaz, line 46: Please keep the output examples but replace the env name there.

.github/CONTRIBUTING.md
```diff
@@ -56,20 +51,7 @@ Prerequisites:
    annoyingly slow):
 
    ```shell-session
-   $ tox -e lint
-   lint create: .tox/lint
```
webknjaz, line 60: Same here plz. I specifically added the output so there's an expectation that is visible.

docs/changelog-fragments.d/708.misc.md
```diff
@@ -0,0 +1,2 @@
+Refresh documentation regarding linting code when
+proposing new changes. -- by {user}`ssbarnea`
```
webknjaz, line 2: There are a few things to think about here:
1) [consistency] It's a bit weird to have an em dash after a period. So the samples don't have it. I think this should be dropped
2) [writing style] Your sentence looks like a git commit message (imperative mood, present tense, it reads as a TODO-list item now) but it should instead be in something like the past tense, explaining the high-level changes in the experiences for the reader compared to the previous release. Imagine you describe that's happened already.
3) [clarity] I think the explanation is not very specific. It reads as "something random happened, read the code to learn what exactly". Instead, we need to extract the highlights of what changed and make them prominent. This does not need to go into internal/implementation details, just a high-level "we have affected you with this specific thing".
4) [category] I'm not entirely sure that this belongs to `misc` — could you explain why you chose this category? This sounds like a contributor-facing change for which I'd go for `internal`. It also mentions the documentation for which we have `doc`. Do you have any thoughts on this?

I've explained some of these things in a detailed manner here https://github.com/ansible/ansible-navigator/pull/641/files#r765392465 — please read it.

ssbarnea: My toughs about this are that we spend too much time talking and complaining about form instead of just fixing issues. Too much bureaucracy for the number of fixes we need to make.

For example I would not have bothered to include a changelog frarment for that change at all. In the end documentation is out of sync, and written in way that makes impossible to keep it in sync with development. Including some output lines from tox stdout is useless, these lines would change quickly and they provide no real value to the user.

@cidrblock WDYT?

webknjaz: An inconsistent changelog is an issue, I don't enjoy having to refactor multiple fragments right before a release ​just because somebody didn't care to think about them on time.

And the output lines are useful, even when outdated. They are stripped off, for this reason, just showing mostly the beginning and the form of the output to expect.

ssbarnea: Updated with past-tense, no ending dot, used fo docs instead of misc.

docs/changelog-fragments.d/708.internal.md
```diff
@@ -0,0 +1,2 @@
+Corrected documentation regarding linting code when
+proposing new changes -- by {user}`ssbarnea`
```
webknjaz, line 2: Could you be more specific about what changed?

webknjaz:
If you could move changing the controversial parts to a separate PR, I don't see why the rest couldn't be merged.

webknjaz (commented):
If you could move changing the controversial parts to a separate PR, I don't see why the rest couldn't be merged.

docs/changelog-fragments.d/708.misc.md
cidrblock: I think I'm glad we have change logs now.
I think I will try to make my change logs past tense
I will probably always end sentences with a period.
I will probably forget if the change log gets a period or the docstring, since the 2 seem to conflict.
I also think it's might be easier to adopt changelog fragments being added first with follow on efforts to sync their grammar and tense.

webknjaz: @cidrblock that sounds reasonable. I tend to use periods only if there's more than one sentence but not before `--`. I wonder if I could lint this, it'd be easier to catch early w/o a human review.

@ssbarnea thanks for doing that, I'd like to keep the discussion going, though. I think in some cases only the PR submitter may judge properly which category is needed. We need to iterate on specificity too. I recognize that it's hard and frustrating at times. The reason it's better to do it early is that when the release time comes, it's easier to understand what's about to be released and the whole thing is closer to the final/ready state and is already high-quality by that time.

docs/changelog-fragments.d/708.internal.md
webknjaz: Does this sound precise?
```suggestion
Updated the linting tox environment names in the contributing document to
match the configuration changes in the repository. It now mentions separate
commands for ``lint`` and ``lint-vetting`` suites -- by {user}`ssbarnea`
```

ssbarnea:
> If you could move changing the controversial parts to a separate PR, I don't see why the rest couldn't be merged.

I do not think that listing outdated linting output is controversial or that it should be outside this change. This change is about ensuring that documentation is correct too. What am I supposed to do replace `linters` in those incorrect output lines with `lint` and merge it like this, knowing that  user will see a different out when calling the tool?  Even worse, I already see how I would be asked to add another annoying changelog entry for fixing few lines of documentation, basically doubling the effort.

webknjaz:
> I do not think that listing outdated linting output is controversial

Your opinion is different from mine. It is the very definition of "controversial" :man_shrugging:

webknjaz:
> that it should be outside this change. This change is about ensuring that documentation is correct too.

Correcting the output is in the scope of this PR, deleting it isn't.

> What am I supposed to do replace `linters` in those incorrect output lines with `lint`

That is exactly what I was asking you to do.

> and merge it like this, knowing that user will see a different out when calling the tool?

I don't think it's different. I was originally adding a generic tox output one would see on the first run. Do you mean you see that the env is reused on the second invocation because it's already cached on your machine?

> Even worse, I already see how I would be asked to add another annoying changelog entry for fixing few lines of documentation, basically doubling the effort.

I don't think that a fragment is necessary for such changes.

> How about this? I will remove these lines from this review but you will fix them with PR?

Just replace the env name for now. If it'll still be different, I'll check locally and correct that.

.github/CONTRIBUTING.md
```diff
@@ -56,12 +56,12 @@ Prerequisites:
...
-   lint installdeps: pre-commit, pylint ~= 2.8.0, pylint-pytest < 1.1.0
-   lint installed: ...
-   lint run-test-pre: PYTHONHASHSEED='2351399476'
-   lint run-test: commands[0] | python -m pre_commit run ... --all-files -v
+   $ tox -e lint-vetting
+   lint-vetting create: .tox/lint-vetting
+   lint-vetting installdeps: pre-commit, pylint ~= 2.8.0, pylint-pytest < 1.1.0
+   lint-vetting installed: ...
+   lint-vetting run-test-pre: PYTHONHASHSEED='2351399476'
+   lint-vetting run-test: commands[0] | python -m pre_commit run ... --all-files -v
```
ssbarnea, line 64: This is why we needed to disable MD013 on code-blocks.

ssbarnea:
recheck

.github/CONTRIBUTING.md
webknjaz: Interesting. This is why I initially shortened the example. I suppose, this can be fixed in a follow-up.

webknjaz (approved):

PR #710, opened 2021-12-11
Rename module name sentinals to sentinels (misspell)

ssbarnea:


ssbarnea:
@cidrblock This renames a module. Do we need changelog fragment or not really?

tests/fixtures/integration/actions/templar/test_welcome_interactive_noee.py/test/8.json
```diff
@@ -33,7 +33,7 @@
         " 19│      - This action will not work normally outside of lockstep strategies",
         " 20│      support: none",
         " 21│    check_mode:",
-        " 22│      description: Can run in check_mode and return changed status prediction withought",
+        " 22│      description: Can run in check_mode and return changed status prediction without",
```
cidrblock, line 36: this one can stay, because I recall us adding an exception for the fixtures folder, and it will get updated later when the fixtures are updated

cidrblock (commented):
crazy how much work one little letter can cause eh? ty and I spelled it wrong

webknjaz:
> 

I think this partially depends on whether we can treat this as public API (technically, we don't use `_private` notation anywhere so there's no clear "statement" in code). Although, it may be useful for the contributors, in which case maybe think about adding an `internal` fragment if it's useful but only for the maintainers, use `misc`.

But a more generic answer to this question would be — think if reading this sort of information would be useful to you (as an external contributor, as a maintainer or as a user). If the answer is "yes", you probably want a fragment. If the answer is "no" — it depends if somebody else can think of a reason to include it.

webknjaz:
@ssbarnea I noticed that this changes more things than mentioned in the title or description. Plz reflect those changes there.

cidrblock:
RE change log. I have no pref here today.  If we had 1000s of contributors, I might weigh in.

Wouldn't that be cool?

cidrblock (approved):

docs/changelog-fragments.d/710.internal.md
```diff
@@ -0,0 +1,2 @@
+Renamed module name sentinals to sentinels, correcting
```
webknjaz, line 1: Let's highlight the literals:
```suggestion
Renamed module name `sentinals` to `sentinels`, correcting
```

docs/changelog-fragments.d/710.internal.md
```diff
@@ -0,0 +1,2 @@
...
+a an unfortunate spelling mistake. -- by {user}`ssbarnea`
```
webknjaz, line 2: Duplicate article + period style:
```suggestion
an unfortunate spelling mistake -- by {user}`ssbarnea`
```

webknjaz (approved):

webknjaz:
@ssbarnea I've corrected a few very minor typos in the change note and requested gating. I hope that's alright with you.

PR #719, opened 2021-12-11
[lint] Remove spaces surrounding docs strings

cidrblock:
- does what it says
- also remove per-file exceptions for D210

D210 reads "No whitespaces allowed surrounding docstring text"  which is found with the pydocstyle pkg

"pydocstyle is a static analysis tool for checking compliance with Python docstring conventions."

pydocstyle is used by flake with https://pypi.org/project/flake8-docstrings/

this is not an enforced linter, but we had a mix and I had a few minutes.


webknjaz (approved):

cidrblock:
@ssbarnea does this meet your challenge :)

PR #726, opened 2021-12-12
Check against both Python 3.10 and 3.6 in MyPy

webknjaz:


ssbarnea:
Hmm, I never did this in any project while I still wondered which version is more appropriate. Can you please provide a clear example regarding why we need two of them? I am more inclined to pin it to a specific version than having multiple runs.

On the other hand I do remember once or twice deciding to pin version of python to a specific version (usually newest) to enforce extra checks that were not available under older pythons.

Also we can achieve extra coverage with zuul by changing system python to various versions without having to modify anything on linters side. Running twice when using pre-commit is no joy as it will require both interpreters for any contributor and also slowdown its execution. I am more inclined to have non-implicit `tox -e py38-lint` which would run with a specific version of python, if pre-commit respects the system python.

The next question is why testing with `py36` and `py38` and not also with `py310`?

If we are to add multiple mypy runs we need to be sure we cannot avoid it. (Examples of failure to catch specific issues on each version we test, not potential).

webknjaz:
@ssbarnea I first saw this in pip back when it supported python 2. And now I've integrated it into Cheroot too. I don't think we'll find a more elegant version, since pre-commit doesn't support matrices.

ssbarnea (changes requested):
I am delaying this until we have proof of one failure that is found by this. There are several other **maintenance and performance** related issues with doubling mypy config inside pre-commit.

If we need consistency we can configure pre-commit default python interpreter to specific version. I mention this because requiring two versions of python is worse than requiring a single specific version, especially for occasional contributors.

webknjaz:
I recall the other day there was already a problem with MyPy specifically under Python 3.6. What about that one?

ssbarnea:
I converted this to draft to remove it from our daily reminder. I personally do not find it really needed, considering that one version of mypy is enough. If you still want this to happen, bring the subject into a team meeting.

As I often say, just because we can do something, it does not mean that we must. Often the incurred costs are greater than the benefits.

webknjaz:
Requested more reviews, then.

webknjaz:
@ssbarnea how about a compromise — we make one of the executions only run in the CI. In that scope, it's not really significant but wouldn't add time locally.

cidrblock (approved):
Checking with multiple versions here seems reasonable.

Some things we should consider and iterate on in the future are the use of a config file vs. config in pre-commit, and specifically which versions we test with.

I see this as an improvement with little cost and will get a baseline config in place for more than one version.

.pre-commit-config.yaml
```diff
@@ -205,6 +207,35 @@ repos:
     - --namespace-packages
     - --non-interactive
     - --pretty
+    - --python-version=3.8
```
ssbarnea, line 210: This is incorrect, docs say `Type check code assuming it will be running on Python x.y` -- that is not the same as running on that python version and I am sure that the outcome will be different.

If you really want run **with**  different versions of python, you need to tell pre-commit to do so, it is possible to enforce python version.

When you do this also enforce the implicit one to run with minimal python and add a comment like "minimal version of python we support" and add the manual one using the maximal version of python we support (py310), also mentioning it. That will help someone else understand why these values were picked and be able to bump them accordingly.

If you want we could switch the implicit version to run max version instead of min version, I do not have a strong preference for one of them.

ssbarnea (changes requested):
Make use of https://pre-commit.com/#config-language_version

.pre-commit-config.yaml
webknjaz: > This is incorrect, docs say `Type check code assuming it will be running on Python x.y` -- that is not the same as running on that python version and I am sure that the outcome will be different.
> 
> If you really want run **with** different versions of python, you need to tell pre-commit to do so, it is possible to enforce python version.

I'm not sure if this judgment is correct. The approach I used here is derived from other projects I trust and I've seen using it. They've never had to change the runtime. In fact, I've integrated this approach in some projects recently and it was quite evident that it follows through different code paths depending on `--python-version`, requiring different sets of `additional_dependencies`.

Moreover, when you mentioned this, I've gone ahead and checked the documentation of mypy itself too — it requires Python 3.6 in runtime but still supports checking Python 2.7, for example. This supports the point that changing the runtime is not necessary:
> Although you must install Python 3 to run mypy, mypy is fully capable of type checking Python 2 code as well [...]

_(https://mypy.readthedocs.io/en/stable/getting_started.html#installing-and-running-mypy — note at the bottom of the section)_

I suspect this is because MyPy does not evaluate the code in runtime but rather inspects AST statically instead. This also explains how it works with type stubs and type comments.

The bottom line is that it is not required to change `language_version` to make it work correctly.
OTOH, if you have a reproducer that clearly demonstrates what you imply — please share it, and it would be reasonable to reconsider, then. But until then, I've only found evidence that it works as documented.

> When you do this also enforce the implicit one to run with minimal python and add a comment like "minimal version of python we support" and add the manual one using the maximal version of python we support (py310), also mentioning it. That will help someone else understand why these values were picked and be able to bump them accordingly.

This sounds reasonable. Having an explanatory motivation comment is always a good idea in such cases. In fact, there's even a WPS rule (for Python code, of course, but the ideas are similar) that encourages people not to pass arbitrarily chosen constants and have them available under named context-aware variables first.

> If you want we could switch the implicit version to run max version instead of min version, I do not have a strong preference for one of them.

I don't either. Both options have advantages and downsides which I'm not trying to explore in this PR at the moment. But if you have a preference, feel free to document your thoughts here or in a follow-up.

ssbarnea (changes requested):
Look at screenshot below that proves that mypy is not using different interpreters when you pass mentioned arg:


This was on fedora 34 with only python3.10, there was no python3.6 on the machine and surprise.... i passed.

If we are to test with multiple python versions, let's do it, not claim that we do it.

webknjaz:
You are right, it doesn't. But the point is that _it shouldn't_. It doesn't run the source code, it doesn't import those modules. It just parses the AST, which can be done from any Python runtime.

ssbarnea:
In that case the only thing that we need is to always run with `--python-version=3.6` and not have to bother about what is the current version of the interpreter. I think we already have a similar feature for configuring targeted runtime in other tools. Still, I do not see any benefit for calling it twice with different arguments.

webknjaz:
> Still, I do not see any benefit for calling it twice with different arguments.

This will check the code paths for 3.6 and 3.10 which will have different typing implications. If one has the correct types, there's no guarantee that the other has proper types too. These typing "runtimes" are different and so it is beneficial for coverage.

.pre-commit-config.yaml
webknjaz: (marking as resolved because it's been addressed)

ssbarnea (dismissed):
Sorry but in the absence of a clear example that affected this project I am against a change that:

* explodes the complexity of pre-commit config with 40-50 extra lines, still not addressing a more important known issue which is the fact that we know that running mypy with one interpreter version can produce different results than another one (regardless the argument)
* silently introduce extra configuration to mypy which is for sure unrelated to that change
* would make config harder to maintain

I already seen a similar change that added pylint-pytest plugin without real use-case. The fact that there was a case on another project (outside devtools) where something proved useful, is not a good enough reason for introduce a tool/plugin. I checked and pylint-pytest plugin was not used by any devtools project while all already used pytest successfully.

Adding a new tool/plugin/config needs proof that it helps the affected project.  I cannot see that and the description of this does not point to a recent path that was needed.

Such changes should be advertised as **preventive measures** for past problems, but I don't the the trigger.

Please ask the rest of the team and if enough think that it's ok, feel free to ignore my -1 on it.

Meanwhile I created a competing PR that aim to address the same goal, #804 so we can compare them and decide which path to go: pick one, another or none.

webknjaz:
> * is the fact that we know that running mypy with one interpreter version can produce different results than another one (regardless the argument)

This is not accurate. Different results come from different types in the dependencies. What you describe is a side effect, not the root cause. Also, I don't think that the part about "regardless the argument" is true at all. There is no proof, and all the prof I have points at exactly the opposite (see the comments above).
OTOH, I've detailed how MyPy actually works in my comments above.

> * silently introduce extra configuration to mypy which is for sure unrelated to that change

WAT? If you have _any_ proof, here's the time to show it.

Ugh... I'm not even going to reply to the rest of that ..

ganeshrn (approved):
I am good within running mypy in CI. Thank you for doing this.
Python versions to run against can be fine tuned in follow up PR's.

priyamsahoo (approved):
MyPy in the CI !
LGTM !

ssbarnea:
The current failures we receive from zuul, even on this PR, or an other ones https://dashboard.zuul.ansible.com/t/ansible/build/286aea6b783b45e09f87df641b0e9afe are a proof that this change does not make mypy execution run more reliable.

As seen above pre-commit auto-update produced a change that failed on zuul, where zuul was using py36, but passed on pre-commit.ci. If I remember correctly on GHA we default to py38 but we could switch to 36. Still, even if we match zuul version it would far better to ensure that when pre-commit runs produces same results, even at the cost of requiring developer to install a particular version of python.

As running a single job with multiple versions of python adds some challenges for GHA jobs, I am wondering if we could combine my recommendation or using `language_version` with the command line to make its outcome more reliable and avoid having it to fail on various systems, based on which version of python they have installed.

webknjaz:
I agree that the failure influences pylint but that's because of its semi-dynamic nature (it imports/evaluates stuff). I don't see a direct link with MyPy, though. I've done extra research that suggests that it shouldn't be an issue for us in the current setting. Let's revisit if we see it misbehaving.

PR #728, opened 2021-12-12
More spellcheck fixes

ssbarnea:
- adds two known spellcheck errors to ignored words, with note
- reduces the exclude list to generated files
- fixes one last minor spellcheck error

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -151,56 +151,15 @@ repos:
   rev: v2.1.0
   hooks:
   - id: codespell
+    # FIXME: Address remove current spelling exceptions:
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -151,56 +151,15 @@ repos:
   rev: v2.1.0
   hooks:
   - id: codespell
+    # FIXME: Address remove current spelling exceptions:
+    args: ["-L", "tipe,dout"]
+    # We exclude generated or external files as they are not directly under
```
cidrblock, line 156:
```suggestion
    # We exclude generated and external files as they are not directly under
```

cidrblock (approved):

PR #736, opened 2021-12-14
Bump flake8 WPS plugin to v0.16.0

webknjaz:
The 0.16.x stream is the first one compatible with flake8 v4.


cidrblock (approved):
Approved with the understanding this is for things being vetted. Please confirm :)

webknjaz:
Yep, that's still the case.

PR #737, opened 2021-12-14
Shorten output examples in CONTRIBUTING doc

webknjaz:


ssbarnea (dismissed):
Few problems with that:
* the output is not in sync with reality
* will prevent us from adding correct console output examples as it removes .markdownlint config needed for that.

IMHO, if we are to simplify the output, we should remove all of it, nothing on it is relevant for documentation purposes.

In fact I would keep something generic and not verbose like:
```
doc8.................................................(no files to check)Skipped
codespell................................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
Mixed line ending........................................................Passed
check BOM - deprecated: use fix-byte-order-marker........................Passed
Check that executables have shebangs.................(no files to check)Skipped
Check for merge conflicts................................................Passed
Debug Statements (Python)............................(no files to check)Skipped
yamllint.................................................................Passed

```

For documentation purposes the output does not need to include all lines, and only to demonstrate what the user may expect. In the end I am ok with even total removal of output lines from docs.

I will let @cidrblock decide on that, if he likes it as is I will remove my requested changes from it and gate it.

webknjaz:
> * the output is not in sync with reality

I don't think it's true, according to my testing.

webknjaz:
> * will prevent us from adding correct console output examples as it removes .markdownlint config needed for that.

Adding it was not necessary in the first place and with the improved shortened examples this band-aid should be dropped.

cidrblock:
IMO:

We want this information in the guide, so contributors know how to run it.
We also want it here so PRs are "clean" when they come in.

Based on that, it make more sense to me to have 

1) the command they should run
2) the output of what a sample "clean" run looks like

optionally, an example error they might see.

Here's my thinking, I see the command and the clean output.  I run the command and don't see clean output, I fix my errors until my output looks clean like it does in the contributors guide.  I think this is the message we want to purvey

webknjaz:
@cidrblock I think this does some of that but providing a full output is hard because it's big and distracting so I only include parts of it that are more than enough to recognize it.

cidrblock:
maybe something like this?

```
$ tox -e lint

# Output from precommit 
Tabs remover.............................................................Passed
Trim Trailing Whitespace.................................................Passed
<...>
mypy.....................................................................Passed

# Pylint output
lint run-test: commands[1] | pylint ansible_navigator tests --ignore=tm_tokenize
--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

# Black output
lint run-test: commands[2] | black -v --diff --check /home/bthornto/github/ansible-navigator

src/ansible_navigator/actions/back.py already well formatted, good job.
src/ansible_navigator/actions/filter.py already well formatted, good job.
<...>
tests/unit/test_cli.py already well formatted, good job.

All done! ✨ 🍰 ✨
219 files would be left unchanged.

# Summary for all checks
___________________________________ summary ____________________________________
  lint: commands succeeded
  congratulations :)
```

webknjaz:
@cidrblock I wanted to avoid specific linter outputs and just give a general feel of what to expect. Also, I included the tox output people see when they don't yet have a pre-cached tox env.

cidrblock:
ok, can we at least include the
```
___________________________________ summary ____________________________________
  lint: commands succeeded
  congratulations :)
```
I think it's fun.

cidrblock (approved):
I'm approving this, assuming you both agree what ever changes are made it is at least incrementally better than what is there today.

webknjaz:
> can we at least include the

Oh, I didn't see the question in this (probably there's no question mark) and missed it. But yes, this is possible. We just need to remember to balance between a very long chunk of text in the middle of a relatively small document vs the informative parts.

PR #738, opened 2021-12-14
[configuration] delayed postprocs, extra volmounts

relrod:
I've pulled this out from my lint work, where it was needed, because it feels like it should be a separate change because we probably want it regardless.

This allows post processors to be delayed until after the first pass of post processors has been completed.
At that point, it will run the delayed ones. There is only one "level" of delay (so delayed post processors can't wait for a third pass). This is intentional, to keep things simple, otherwise we'd have a whole tree of post processors and it gets complex quickly.

Doing it this way just provides a simple way to say "this post processor depends on some other things that happen in the first pass, let's wait for this before we actually run it."

```
Change:

- Allow post-processors to be delayed until after the first
  (alphabetical) pass. In the delayed pass, delayed post-processors
  are again executed alphabetically.

- In the interest of keeping configuration simple, we don't allow for
  multiple "levels" of delay. A post-processor is either run during the
  first pass, or delayed until after the first/normal pass.

- This also adds an "extra volmounts" variable to the navigator
  post-processor class, which can be used by other post-processors to
  signify that they want a volume to be mounted. The volume-mount post
  processor is now delayed to accommodate.

Test Plan:

- Using this in lint/navigator work, no new tests yet because there is
  no other post processor currently which adds an extra volume mount.

```

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/configurator.py
+++ b/src/ansible_navigator/configuration_subsystem/configurator.py
@@ -234,7 +234,17 @@ class Configurator:
                 self._messages.append(LogMessage(level=logging.INFO, message=message))
 
     def _post_process(self) -> None:
+        delayed = []
+        normal = []
+
+        # Separate normal and delayed entries so they can be processed in that order.
         for entry in self._config.entries:
+            if entry.delay_post_process:
+                delayed.append(entry)
+            else:
+                normal.append(entry)
+
+        for entry in normal + delayed:
             if self._initial or entry.change_after_initial:
                 processor = getattr(self._config.post_processor, entry.name, None)
                 if callable(processor):
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -6,6 +6,8 @@ import os
 import shlex
 import shutil
 
+from dataclasses import dataclass, field
+from enum import Enum
 from pathlib import Path
 from typing import List
 from typing import Tuple
@@ -46,12 +48,63 @@ def _post_processor(func):
     return wrapper
 
 
+class VolumeMountOption(Enum):
+    """Options that can be tagged on to the end of volume mounts.
+
+    Usually these are used for things like selinux relabeling, but there are
+    some other valid options as well, which can and should be added here as
+    needed. See ``man podman-run`` and ``man docker-run`` for valid choices and
+    keep in mind that we support both runtimes.
+    """
+
+    # Relabel as private
+    Z = "Z"
+
+    # Relabel as shared.
+    z = "z"  # pylint: disable=invalid-name
+
+
+@dataclass
+class VolumeMount:
+    """Describes EE volume mounts."""
+
+    #: The name of the config option requiring this volume mount.
+    calling_option: str
+
+    #: The source path of the volume mount.
+    src: str
+
+    #: The destination path in the container for the volume mount.
+    dest: str
+
+    #: Options for the bind mount.
+    options: List[VolumeMountOption] = field(default_factory=list)
+
+    def exists(self) -> bool:
+        """Determine if the volume mount source exists."""
+        return Path(self.src).exists()
+
+    def to_string(self) -> str:
+        """Render the volume mount in a way that (docker|podman) understands."""
+        out = f"{self.src}:{self.dest}"
+        if self.options:
+            joined_opts = ",".join(o.value for o in self.options)  # pylint: disable=not-an-iterable
+            out += f":{joined_opts}"
+        return out
+
+
 PostProcessorReturn = Tuple[List[LogMessage], List[ExitMessage]]
 
 
 class NavigatorPostProcessor:
     # pylint:disable=too-many-public-methods
     """application post processor"""
+
+    def __init__(self):
+        # Volume mounts accumulated from post processing various config entries.
+        # These get processed towards the end, in the (delayed)
+        # execution_environment_volume_mounts() post-processor.
+        self.extra_volume_mounts: List[VolumeMount] = []
 
     @staticmethod
     def _true_or_false(entry: Entry, config: ApplicationConfiguration) -> PostProcessorReturn:
@@ -233,12 +286,12 @@ class NavigatorPostProcessor:
             entry.value.current = f"{entry.value.current}:latest"
         return messages, exit_messages
 
-    @staticmethod
     @_post_processor
     def execution_environment_volume_mounts(
-        entry: Entry, config: ApplicationConfiguration
+        self, entry: Entry, config: ApplicationConfiguration
     ) -> PostProcessorReturn:
         # pylint: disable=unused-argument
+        # pylint: disable=too-many-branches
         """Post process set_environment_variable"""
         messages: List[LogMessage] = []
         exit_messages: List[ExitMessage] = []
@@ -305,6 +358,18 @@ class NavigatorPostProcessor:
                     return messages, exit_messages
 
             entry.value.current = parsed_volume_mounts
+
+        if self.extra_volume_mounts and entry.value.current is C.NOT_SET:
+            entry.value.current = []
+
+        for mount in self.extra_volume_mounts:
+            if not mount.exists():
+                exit_msg = (
+                    f"The volume mount source path '{mount.src}', needed by "
+                    f"{mount.calling_option}, does not exist."
+                )
+                exit_messages.append(ExitMessage(message=exit_msg))
+            entry.value.current.append(mount.to_string())
         return messages, exit_messages
 
     @staticmethod
--- a/tests/unit/configuration_subsystem/test_navigator_post_processor.py
+++ b/tests/unit/configuration_subsystem/test_navigator_post_processor.py
@@ -1,0 +1,31 @@
+"""Tests for the navigator config post-processor."""
+
+import pytest
+
+from ansible_navigator.configuration_subsystem.navigator_post_processor import (
+    VolumeMount,
+    VolumeMountOption,
+)
+
+
+@pytest.mark.parametrize(
+    ("volmount", "expected"),
+    (
+        (VolumeMount("test_option", "/foo", "/bar"), "/foo:/bar"),
+        (VolumeMount("test_option", "/foo", "/bar", [VolumeMountOption.z]), "/foo:/bar:z"),
+        (
+            VolumeMount("test_option", "/foo", "/bar", [VolumeMountOption.z, VolumeMountOption.Z]),
+            "/foo:/bar:z,Z",
+        ),
+        (VolumeMount("test_option", "/foo", "/bar", []), "/foo:/bar"),
+    ),
+    ids=(
+        "normal mount",
+        "mount with relabel option",
+        "mount with a list of options",
+        "mount with empty list of options",
+    ),
+)
+def test_navigator_volume_mount_to_string(volmount, expected):
+    """Make sure volume mount ``to_string`` is sane."""
+    assert volmount.to_string() == expected
--- a/src/ansible_navigator/configuration_subsystem/definitions.py
+++ b/src/ansible_navigator/configuration_subsystem/definitions.py
@@ -60,6 +60,7 @@ class Entry(SimpleNamespace):
     apply_to_subsequent_cli: Should this be applied to future CLIs parsed
     choice: valid choices for this entry
     cli_parameters: argparse specific params
+    delay_post_process: Post process in normal (alphabetical) order or wait until after first pass?
     environment_variable_override: override the default environment variable
     name: the reference name for the entry
     settings_file_path_override: over the default settings file path
@@ -76,6 +77,7 @@ class Entry(SimpleNamespace):
     change_after_initial: bool = True
     choices: List = []
     cli_parameters: Union[None, CliParameters] = None
+    delay_post_process: bool = False
     environment_variable_override: Union[None, str] = None
     settings_file_path_override: Union[None, str] = None
     subcommands: Union[List[str], Constants] = Constants.ALL
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -291,6 +291,7 @@ NavigatorConfiguration = ApplicationConfiguration(
         Entry(
             name="execution_environment_volume_mounts",
             cli_parameters=CliParameters(action="append", nargs="+", short="--eev"),
+            delay_post_process=True,
             settings_file_path_override="execution-environment.volume-mounts",
             short_description=(
                 "Specify volume to be bind mounted within an execution environment"


cidrblock (approved):
This is cool.  I was wondering how we could bypass the alphabetical order thing, a send pass is a neat idea.  Couple suggestions, nothing major.

The only other thing I was thinking about it a future developer might be able to add a new volume mount in a incorrect format.  We could create a SimpleNamespace as a carrier that had it's own function to turn itself into a string.

If we set the type of `extra_volume_mounts` to List[VolumeMount] mypy would catch it in the future.

Later on, we could eliminate some of the logic in the ee volume mount post processor by passing the dicts through the new simplenamespace SimpleNamespace(**d).to_string()

just thinking aloud here about a mistake I could make in the future :)

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -53,6 +54,27 @@ class NavigatorPostProcessor:
     # pylint:disable=too-many-public-methods
     """application post processor"""
 
+    def __init__(self):
+        # Volume mounts accumulated from post processing various config entries.
+        # These get processed towards the end, in the (delayed)
+        # execution_environment_volume_mounts() post-processor.
+        self.extra_volume_mounts = []
```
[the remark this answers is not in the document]
ssbarnea: This is implicit, not need to do that.

ssbarnea (approved):

ssbarnea:
@relrod Please ensure zuul check is passing, otherwise we will not be able merge it.

relrod:
recheck

relrod:
recheck

relrod:
this is fun

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -234,7 +234,17 @@ def _apply_cli_params(self) -> None:
                 self._messages.append(LogMessage(level=logging.INFO, message=message))
 
     def _post_process(self) -> None:
+        delayed = []
```
webknjaz, line 237: Is the order important? From a prompt look I got an impression that this could be a set, no?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +48,39 @@ def wrapper(*args, **kwargs):
...
+    def exists(self) -> bool:
+        """Does the volmount source exist?"""
+        return Path(self.src).exists()
+
+    def to_string(self) -> str:
+        """Renders the volume mount in a way that (docker|podman) understands."""
+        out = f"{self.src}:{self.dest}"
+        if self.label is not None:
+            out += f":{self.label}"
+        return out
```
webknjaz, line 68: I think if we improve the var naming like this, it could add more clarity to the code:
```suggestion
        volume_mapping = f"{self.src}:{self.dest}"

        if self.label is None:
            return volume_mapping

        return f"{volume_mapping}:{self.label}"
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +48,39 @@ def wrapper(*args, **kwargs):
...
 
 class NavigatorPostProcessor:
     # pylint:disable=too-many-public-methods
     """application post processor"""
 
+    def __init__(self):
+        # Volume mounts accumulated from post processing various config entries.
+        # These get processed towards the end, in the (delayed)
+        # execution_environment_volume_mounts() post-processor.
+        self.extra_volume_mounts: List[NavigatorVolumeMount] = []
```
webknjaz, line 82: Pro tip: if we turn the comment into an _attribute docstring_, Sphinx will pick it up and render in the docs.
```suggestion
        self.extra_volume_mounts: List[NavigatorVolumeMount] = []
        """Volume mounts accumulated from post-processing various config entries.

        These get processed towards the end, in the (delayed)
        execution_environment_volume_mounts() post-processor.
        """
```

webknjaz: @cidrblock (educational hint) FYI in case you didn't know about the existence of attr docstrings :point_up:

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -223,12 +251,12 @@ def execution_environment_image(
...
         return messages, exit_messages
 
-    @staticmethod
     @_post_processor
     def execution_environment_volume_mounts(
-        entry: Entry, config: ApplicationConfiguration
+        self, entry: Entry, config: ApplicationConfiguration
     ) -> PostProcessorReturn:
         # pylint: disable=unused-argument
+        # pylint: disable=too-many-branches
```
webknjaz, line 259: Any chance to fix this?

tests/unit/configuration_subsystem/test_navigator_post_processor.py
```diff
@@ -0,0 +1,23 @@
+"""tests for nav config post processor"""
```
webknjaz, line 1:
```suggestion
"""Tests for the navigator config post-processor."""
```

tests/unit/configuration_subsystem/test_navigator_post_processor.py
```diff
@@ -0,0 +1,23 @@
...
+
+import pytest
+
+from ansible_navigator.configuration_subsystem.navigator_post_processor import NavigatorVolumeMount
+
+
+@pytest.mark.parametrize(
+    "volmount, expected",
```
webknjaz, line 9: Readability tip:
```suggestion
    ("volmount", "expected"),
```
(there's a linter for this that is not yet enforced but similar changes are already applied to the main branch)

tests/unit/configuration_subsystem/test_navigator_post_processor.py
```diff
@@ -0,0 +1,23 @@
...
+        (NavigatorVolumeMount("/foo", "/bar", None), "/foo:/bar"),
+    ),
+    ids=(
+        "normal mount",
+        "mount with relabel option",
+        "mount with explicit None for relabel option",
+    ),
+)
+def test_navigator_volume_mount_to_string(volmount, expected):
+    """make sure vol mount to_string is sane"""
```
webknjaz, line 22: PEP 257 compat
```suggestion
    """Make sure vol mount to_string is sane."""
```

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -234,7 +234,17 @@ def _apply_cli_params(self) -> None:
...
+        normal = []
+
+        # Sort delayed and normal entries.
```
webknjaz, line 240: Would this be more precise?
```suggestion
        # Separate delayed and normal entries.
```

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -234,7 +234,17 @@ def _apply_cli_params(self) -> None:
...
         for entry in self._config.entries:
+            if entry.delay_post_process:
+                delayed.append(entry)
+            else:
+                normal.append(entry)
+
+        for entry in normal + delayed:
```
webknjaz, line 247: I don't know if the memory consumption is important here but if it is, maybe make this lazy:
```suggestion
        for entry in itertools.chain(normal, delayed):
```

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -234,7 +234,17 @@ def _apply_cli_params(self) -> None:
                 self._messages.append(LogMessage(level=logging.INFO, message=message))
 
     def _post_process(self) -> None:
+        delayed = []
+        normal = []
```
webknjaz, line 238:
```suggestion
        delayed_config_entries = []
        normal_config_entries = []
```
(maintainability+readability)

src/ansible_navigator/configuration_subsystem/configurator.py
relrod: Order could be important in theory. Right now the post processors are run in alphabetical order. This keeps that, but just lets you move some subset of them to be run (alphabetically, still) after the first ones.

If you need things "delayed" more than one time (e.g. both PP `a` and `b` are delayed already, but `b` needs to depend on `a` for some reason), then you will be at the mercy of how they PPs are sorted.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
relrod: Sure, but this is an implementation detail; I'm not sure why we'd want this in user-facing docs.

src/ansible_navigator/configuration_subsystem/configurator.py
cidrblock: In this case I don't have strong feelings about the variable names here since they are limited to the function.

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -234,7 +234,17 @@ def _apply_cli_params(self) -> None:
...
+
+        # Separate delayed and normal entries.
```
cidrblock, line 240:
```suggestion
        # Separate normal and delayed entries so they can be processed in that order.
```

cidrblock: This function doesn't have a docstring, which is 100% on me. If it did, I'd suggest in there, so it's perfect here.

src/ansible_navigator/configuration_subsystem/configurator.py
cidrblock: I hope we never have _that_ many :)   If we did, making 2 passes through the entries first for normal, second for delayed would avoid the creation of the new lists all together. Something like this:

```
for should_delay in [False, True]:
    if entry.delay_post_process is should_delay:
        if self._initial or entry.change_after_initial:
            processor = getattr(self._config.post_processor, entry.name, None)
            if callable(processor):
                messages, errors = processor(entry=entry, config=self._config)
                self._messages.extend(messages)
                self._exit_messages.extend(errors)
        else:
            message = f"'{entry.name}' cannot be reconfigured. (post process)"
            self._messages.append(LogMessage(level=logging.INFO, message=message))
```
that said, I think it's fine the way it is.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +48,39 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+@dataclass
```
cidrblock, line 51: (educational) Not intended to suggest one or the other, I am curious why not a namedtuple sine it shouldn't change....

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +48,39 @@ def wrapper(*args, **kwargs):
...
+    src: str
+    dest: str
+    label: Optional[str] = None
+
+    def exists(self) -> bool:
+        """Does the volmount source exist?"""
+        return Path(self.src).exists()
+
+    def to_string(self) -> str:
+        """Renders the volume mount in a way that (docker|podman) understands."""
```
cidrblock, line 64:
```suggestion
        """Render the volume mount in a way that (docker|podman) understands."""
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: (comment for comment sake) I prefer the original because the label gets appended to the end of the vol mount, and I think the first shows that in a more straight forward fashion

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: @webknjaz this? https://www.python.org/dev/peps/pep-0224/

That's cool......... There are a few places that could be useful where there are a ton of attrs, I'm thinking about some of the runner stuff.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: I think this would probably get fixed in a follow on PR that takes advantage of the logic in the new dataclass and replaces some of the logic here.

Even if that refactoring was in this PR, I think it would muddy the waters and distract from the new logic of delayed and normal.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -295,6 +323,18 @@ def execution_environment_volume_mounts(
...
+
+        if self.extra_volume_mounts and entry.value.current is C.NOT_SET:
+            entry.value.current = []
+
+        for mount in self.extra_volume_mounts:
+            if not mount.exists():
+                exit_msg = (
+                    f"The volume mount source path '{mount.src}', needed by "
+                    "a configuration option, does not exist."
+                )
```
cidrblock, line 335: (commentary) oh....... it this gets hit it means a different post processor added an invalid volume mount, yuck.

If the dataclass was extended to include another attribute called "source_post_processor", we could use it here so when the bug gets filed we might be able to look at the specific one.

tests/unit/configuration_subsystem/test_navigator_post_processor.py
```diff
@@ -0,0 +1,23 @@
...
+import pytest
+
+from ansible_navigator.configuration_subsystem.navigator_post_processor import NavigatorVolumeMount
+
+
+@pytest.mark.parametrize(
+    ("volmount", "expected"),
+    (
+        (NavigatorVolumeMount("/foo", "/bar"), "/foo:/bar"),
+        (NavigatorVolumeMount("/foo", "/bar", "z"), "/foo:/bar:z"),
```
cidrblock, line 12: https://github.com/ansible/ansible-navigator/discussions/757

just to keep track of it, nothing to do with this PR, and it makes the review hard because there is _so_ many of these

src/ansible_navigator/configuration_subsystem/configurator.py
cidrblock: I added the text because I had to look twice at it to see why the lists were even created if they just got put back together :)

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: @relrod I think @webknjaz is referring to the dev/api docs in this case, we don't expose them today but I think it would be cool if we did.

They look like crap now because I didn't add good doc strings last year in a lot of what I wrote.

cidrblock (approved):

src/ansible_navigator/configuration_subsystem/configurator.py
cidrblock: (untested of course)

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
webknjaz: I don't see the label being appended, I see a contextless string being appended to some var called "out" that also does not have a designated semantical meaning.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
webknjaz: > Sure, but this is an implementation detail; I'm not sure why we'd want this in user-facing docs.

The current Sphinx is set up to build a number of documents directly from the Python files. They are marked as "private API". This is not considered user-facing but is de-facto present in the Sphinx docs deployment.

webknjaz: @cidrblock

> @webknjaz this? [python.org/dev/peps/pep-0224](https://www.python.org/dev/peps/pep-0224/)
> 
> That's cool......... There are a few places that could be useful where there are a ton of attrs, I'm thinking about some of the runner stuff.

Interesting... I'm not sure if I saw this rejected PEP so these docstrings are probably not made available in runtime and a treated as no-op statements syntax-wise. But Sphinx is able to parse them out. Also, there's an alternative syntax that makes use of comments. So instead here's a different way of doing the same (using `#:`):
```suggestion
        #: Volume mounts accumulated from post processing various config entries.
        #: These get processed towards the end, in the (delayed)
        #: execution_environment_volume_mounts() post-processor.
        self.extra_volume_mounts: List[NavigatorVolumeMount] = []
```

Ref: https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute

I personally use the "docstring" style normally but there's no strong reason not to use `#:` instead.

src/ansible_navigator/configuration_subsystem/configurator.py
webknjaz: > If we did, making 2 passes through the entries first for normal, second for delayed would avoid the creation of the new lists all together.

Actually, if we need this sort of optimization, knowing that we operate on a list, I'd probably not even create these two extra lists in the first place but have a sequence of IDs pointing to one or another. Or, even better, I'd combine it with separating this into a generator function for clarity.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
webknjaz: @cidrblock dataclasses implement a far more superior and standardized concept, and they also can be frozen (and should be most of the time — immutable objects are easier to reason about, they are safer). Also, AFAIR dataclasses use `__slots__` as an optimization.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
webknjaz: Yeah, it could be filed as a refactoring issue with ideas so it's not lost.

tests/unit/configuration_subsystem/test_navigator_post_processor.py
webknjaz: Unfortunately, the existence of black stands in a way of implementing this and it also makes it harder to send suggested changes via GitHub with the quote variant that is harder to type (`Shift + '`).

src/ansible_navigator/configuration_subsystem/configurator.py
webknjaz: I have a preference for seeing a "delayed config entries list" vs a "list of unspecified delayed abstractions".

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: Would something like this be better, just so I grok it all.....

```
@dataclass
class NavigatorVolumeMount:
    """Describes EE volume mounts."""

    src: str
    #: The local source directory
    dest: str
    #: The mount point within the EE
    label: Optional[str] = None
    #: The label option(s) for the mount

    def exists(self) -> bool:
        """Does the volmount source exist?

        :return: A boolean indicating wether or not the src exists.
        """
        return Path(self.src).exists()

    def to_string(self) -> str:
        """Render the volume mount in a way that (docker|podman) understands.

        The resulting string consists of three fields, separated by colon characters (:).
        The fields must be in the correct order, and the meaning of each field is not immediately 
        obvious. In the case of bind mounts, 
        The first field is the path to the file or directory on the host machine.
        The second field is the path where the file or directory is mounted in the container.
        The third field is optional, and is a comma-separated list of options, such as ro, z, and Z.

        :return: A string formatted as a ``podman`` or ``docker`` volume mount string.
        """
        volume_mount_string = f"{self.src}:{self.dest}"
        if self.label is not None:
            volume_mount_string += f":{self.label}"
        return volume_mount_string
```

cidrblock: (description based on container engine documentation)

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+
+    def exists(self) -> bool:
+        """Does the volmount source exist?"""
+        return Path(self.src).exists()
+
+    def to_string(self) -> str:
+        """Render the volume mount in a way that (docker|podman) understands."""
+        out = f"{self.src}:{self.dest}"
+        if self.options:
+            joined_opts = ",".join([o.value for o in self.options])
```
webknjaz, line 92: I don't see why a generator expression isn't enough.

```suggestion
            joined_opts = ",".join(o.value for o in self.options)
```

relrod: Ask the linter that :)

relrod: Not that my change fixed anything - I was just trying something.

relrod: I _think_ what's happening is that the linter thinks it's not iterable because it can be `None`. It's not smart enough to know that I'm actually checking that above it.

ssbarnea: I would also check its type in the if condition and thus should be enough for mypy to determine that its must be a List. I used something similar in other cases.

webknjaz: @relrod

> I _think_ what's happening is that the linter thinks it's not iterable because it can be `None`. It's not smart enough to know that I'm actually checking that above it.

Could you be more specific about which linter and what the error message was originally?

If Sorin is right and it indeed was `mypy`, one of the ways to make a hint to it, is to add `assert self.options is not None` and if it was `Optional[List]` before, then after than line `mypy` will *know* that it's just a `List`. Another way is to use `typing.cast()`.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+    calling_option: str
+
+    # The source path of the volume mount.
+    src: str
+
+    # The destination path in the container for the volume mount.
+    dest: str
+
+    # Options for the bind mount.
+    options: Optional[List[VolumeMountOption]] = None
```
webknjaz, line 82: I wonder if
```suggestion
    options: List[VolumeMountOption] = []
```
would help.
Or maybe you could use `__post_init__()` or have a custom property that would return `[]` if it's `None`.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+        """Does the volmount source exist?"""
+        return Path(self.src).exists()
+
+    def to_string(self) -> str:
+        """Render the volume mount in a way that (docker|podman) understands."""
+        out = f"{self.src}:{self.dest}"
+        if self.options:
+            joined_opts = ",".join([o.value for o in self.options])
+            out += f":{joined_opts}"
+        return out
```
webknjaz, line 94: What is "out" in this context?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+    src: str
+
+    # The destination path in the container for the volume mount.
+    dest: str
+
+    # Options for the bind mount.
+    options: Optional[List[VolumeMountOption]] = None
+
+    def exists(self) -> bool:
+        """Does the volmount source exist?"""
```
webknjaz, line 85: > D401 First line should be in imperative mood; try rephrasing

```suggestion
        """Determine if the volume mount source exists."""
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+    # The name of the config option requiring this volume mount.
+    calling_option: str
+
+    # The source path of the volume mount.
+    src: str
+
+    # The destination path in the container for the volume mount.
+    dest: str
+
+    # Options for the bind mount.
```
webknjaz, line 81:
```suggestion
    #: Options for the bind mount.
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+class VolumeMount:
+    """Describes EE volume mounts."""
+
+    # The name of the config option requiring this volume mount.
+    calling_option: str
+
+    # The source path of the volume mount.
+    src: str
+
+    # The destination path in the container for the volume mount.
```
webknjaz, line 78:
```suggestion
    #: The destination path in the container for the volume mount.
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+
+
+@dataclass
+class VolumeMount:
+    """Describes EE volume mounts."""
+
+    # The name of the config option requiring this volume mount.
+    calling_option: str
+
+    # The source path of the volume mount.
```
webknjaz, line 75:
```suggestion
    #: The source path of the volume mount.
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -46,13 +49,64 @@ def wrapper(*args, **kwargs):
...
+
+    # Relabel as shared.
+    z = "z"  # pylint: disable=invalid-name
+
+
+@dataclass
+class VolumeMount:
+    """Describes EE volume mounts."""
+
+    # The name of the config option requiring this volume mount.
```
webknjaz, line 72:
```suggestion
    #: The name of the config option requiring this volume mount.
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
webknjaz: Oh, so looked into Zuul output and it's 
```console
src/ansible_navigator/configuration_subsystem/navigator_post_processor.py:92:53: E1133: Non-iterable value self.options is used in an iterating context (not-an-iterable)
```
Not sure if an assert would help here.

relrod:
recheck

PR #739, opened 2021-12-14
Restore execution of pre-commit from inside tox -e lint

ssbarnea:
This correct unintended bug introduced at https://github.com/ansible/ansible-navigator/pull/686/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449L52

Our intention was not to stop running the pre-commit hooks under
zuul. This brings the behavior in sync with our expectation and
documentation.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -47,6 +47,7 @@ commands =
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
+  pre-commit run -a {posargs}
   pylint {[base]pkg_name} tests --ignore=tm_tokenize
   black -v --diff --check {toxinidir}
 


cidrblock (approved):

tox.ini
```diff
@@ -47,6 +47,7 @@ commands =
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
+  pre-commit run -a {posargs}
```
webknjaz, line 50: I think in the past you asked me not to use args like `-a` with unclear meanings in scripts. Now, I'd like you to do the same.
Also, I would expect this to make multiple invocations and not to bring a feature of customizing only one linter invocation, so I'd drop `{posargs}` too:
```suggestion
  {envpython} -m \
    pre_commit run \
    markdownlint \
    --show-diff-on-failure \
    --all-files -v
  {envpython} -m \
    pre_commit run \
    codespell \
    --show-diff-on-failure \
    --all-files -v
  {envpython} -m \
    pre_commit run \
    flake8 \
    --show-diff-on-failure \
    --all-files -v
  {envpython} -m \
    pre_commit run \
    yamllint \
    --show-diff-on-failure \
    --all-files -v
  {envpython} -m \
    pre_commit run \
    mypy \
    --show-diff-on-failure \
    --all-files -v
```

webknjaz:
@ssbarnea why didn't you reply to my reviews before merging?

PR #740, opened 2021-12-14
[lint-vetting] Allow implicit string concatenation

cidrblock:
This is related to a linter that is being vetted, not enforced.

https://wemake-python-stylegui.de/en/latest/pages/usage/violations/consistency.html#wemake_python_styleguide.violations.consistency.ImplicitStringConcatenationViolation

This was discussed briefly here:

https://github.com/ansible/ansible-navigator/pull/624#discussion_r767213354


ssbarnea (approved):

ssbarnea:
I do not need internal change note because we had no real review on wemake rules. If we were about to review or announce it was about introduction of wemake (WPS) rules, not disabling them as damage-control.

I do have the impression that this rule is in direct conflict with black, so happy to see it out.

PR #741, opened 2021-12-14
[lint-vetting] Allow `hasattr` and others

cidrblock:
This is for a linter that is being vetted not enforced

https://wemake-python-stylegui.de/en/0.9.1/pages/violations/best_practices.html?highlight=FUNCTIONS_BLACKLIST#wemake_python_styleguide.violations.best_practices.WrongFunctionCallViolation

forbids:
```
FUNCTIONS_BLACKLIST = frozenset({'__import__', 'compile', 'copyright', 'credits', 'delattr', 'dir', 'eval', 'exec', 'exit', 'globals', 'hasattr', 'help', 'input', 'locals', 'quit', 'staticmethod', 'vars'})
```

Some discussion about this happened here:

https://github.com/ansible/ansible-navigator/pull/624#discussion_r768226898


.flake8
```diff
@@ -43,6 +43,7 @@ extend-ignore =
   # An opposite consistency expectation is currently enforced
   # by pylint via: useless-object-inheritance (R0205):
   WPS306,  # "Found class without a base class: *"
+  WPS421,  # "Forbids to call some built-in functions.", hasattr is good.
```
webknjaz, line 46: This is a pretty strong statement: https://hynek.me/articles/hasattr/.

In general, EAFP is preferable except when you want to make a point. So I'd say this may be context-dependent. Wouldn't it be better to ignore this in a few places where it's actually used and keep the rule on? Other forbidden functions should also be avoided in general.

ssbarnea (approved):

.flake8
ssbarnea: Just today I had a case where I had to use hasattr. I do agree with @cidrblock that there is no need for this rule. TBH, I do not remember ever seeing someone using hasattr when that was not needed, basically because is harder to use that using dot notation.

relrod: > I do not remember ever seeing someone using hasattr when that was not needed

Core [does](https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/base.py#L254) it [in some places](https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/base.py#L294) though I never figured out why.

webknjaz: @ssbarnea have you considered all of the other forbidden invocations? They make sense in general, I think. In some contexts, it may be reasonable to allow them but when we are talking about a generic rule, corner cases are something that would usually need to be excluded in exact locations where they happen.

webknjaz: > Core [does](https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/base.py?rgh-link-date=2021-12-14T16%3A30%3A56Z#L254) it with setattr/getattr [in some places](https://github.com/ansible/ansible/blob/devel/lib/ansible/playbook/base.py?rgh-link-date=2021-12-14T16%3A30%3A56Z#L294) though I never figured out why.

Probably to set attrs that don't have allocated slots (or class level or in the initializer). _That_ looks shady.

webknjaz:
@cidrblock I think this should be reverted because it allows dangerous calls (other than `hasattr`).

PR #743, opened 2021-12-14
pylint: address unspecified-encoding rule

ssbarnea:
Add explicit encoding to file operations. We already had this in several places but not on all.

Related: #742


cidrblock (approved):

cidrblock:
recheck

cidrblock:
gated too soon, check didn't finish which has the older python versions

PR #744, opened 2021-12-14
[lint] Fix/disable pylint errors

cidrblock:
Seems some pylint errors were missed in the share directory.

No attempt was made to refactor the code, most were simply disabled.


ssbarnea (approved):
LGTM, if it does not break any tests it should be ok.

PR #747, opened 2021-12-14
[lint] Disable pylint via tox, enable pylint via precommit

cidrblock:
Depends on #745 #746 #748 #749 and maybe another one if I missed something

No pylint config changes here, although this depends on a little cleanup due to the later version.

Adds `ansible-core` as a dependency so the import statements in `utils/catalog_collections.py` can be resolved

Now also depends on #751 

and #754

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -241,7 +241,6 @@ repos:
     # docs:
     - sphinx
     - setuptools-scm
-    entry: pylint --rcfile=.pylintrc
     exclude: >
       (?x)^
         (
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -223,15 +223,24 @@ repos:
   hooks:
   - id: pylint
     additional_dependencies:
-      - ansible-core
-      - ansible-runner
-      - dataclasses; python_version<"3.7"
-      - libtmux
-      - onigurumacffi
-      - pytest
-      - pyyaml
-      - sphinx
-      - setuptools-scm
+    # NOTE: The entries below must be kept in sync with the contents of `setup.cfg` at all times,
+    # NOTE: including `install_requires` and any extras that may be imported in runtime to accommodate
+    # NOTE: for `pylint`'s ability to import the modules being checked. They must also include any of the
+    # NOTE: test and platform-dependent requirements that may be imported under different circumstances.
+    - ansible-core  # dependency imported in utils/catalog_collections.py
+    # setup.cfg:
+    - ansible-runner ~= 2.0
+    - dataclasses; python_version < "3.7"
+    - importlib-resources; python_version < "3.9.0"
+    - jinja2
+    - onigurumacffi
+    - pyyaml
+    # tests:
+    - libtmux
+    - pytest
+    # docs:
+    - sphinx
+    - setuptools-scm
     entry: pylint --rcfile=.pylintrc
     exclude: >
       (?x)^
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -218,7 +218,7 @@ repos:
     - tests/
     pass_filenames: false
 
-- repo: https://github.com/PyCQA/pylint
+- repo: https://github.com/PyCQA/pylint.git
   rev: v2.12.2
   hooks:
   - id: pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -225,6 +225,7 @@ repos:
     additional_dependencies:
       - ansible-core
       - ansible-runner
+      - dataclasses; python_version<"3.7"
       - libtmux
       - onigurumacffi
       - pytest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -218,39 +218,24 @@ repos:
     - tests/
     pass_filenames: false
 
-- repo: local
+- repo: https://github.com/PyCQA/pylint
+  rev: v2.12.2
   hooks:
   - id: pylint
-    language: system
-    name: PyLint
-    files: \.py$
-    entry: python -m pylint
-    args:
-    - --ignore=tm_tokenize
-    - |-
-      --disable=
-      consider-using-with,
-      implicit-str-concat,
-      import-error,
-      invalid-name,
-      missing-class-docstring,
-      missing-function-docstring,
-      missing-module-docstring,
-      no-else-return,
-      no-member,
-      no-self-use,
-      protected-access,
-      redefined-builtin,
-      too-few-public-methods,
-      too-many-arguments,
-      too-many-branches,
-      too-many-instance-attributes,
-      too-many-locals,
-      too-many-statements,
-      unused-import,
-      useless-else-on-loop,
-      wrong-import-order,
-    stages:
-    - manual
+    additional_dependencies:
+      - ansible-runner
+      - libtmux
+      - onigurumacffi
+      - pytest
+      - pyyaml
+      - sphinx
+      - setuptools-scm
+    entry: pylint --rcfile=.pylintrc
+    exclude: >
+      (?x)^
+        (
+          src/ansible_navigator/tm_tokenize/.*
+        )
+      $
 
 ...
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -223,6 +223,7 @@ repos:
   hooks:
   - id: pylint
     additional_dependencies:
+      - ansible-core
       - ansible-runner
       - libtmux
       - onigurumacffi
--- a/tox.ini
+++ b/tox.ini
@@ -48,7 +48,6 @@ description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
   pre-commit run -a {posargs}
-  pylint {[base]pkg_name} tests --ignore=tm_tokenize
   black -v --diff --check {toxinidir}
 
 [testenv:report]


.pre-commit-config.yaml
```diff
@@ -218,39 +218,26 @@ repos:
     - tests/
     pass_filenames: false
 
-- repo: local
+- repo: https://github.com/PyCQA/pylint
```
webknjaz, line 221: Consistency with the URL style:
```suggestion
- repo: https://github.com/PyCQA/pylint.git
```

.pre-commit-config.yaml
```diff
@@ -218,39 +218,26 @@ repos:
     - tests/
     pass_filenames: false
 
-- repo: local
```
webknjaz, line 221: This does not take into account the deps that were managed externally like `pylint-pytest` (see `tox.ini`) and forces copy-paste of the dependencies in multiple locations. Like keeping the runtime/package metadata and the test deps in sync with this hook setup.
Besides, it makes it impossible to implement a reproducible env for this check w/o the dependency tree transitive constraints pinned which is otherwise possible with externally-managed deps in tox.

Is this a trade-off we're willing to make?

.pre-commit-config.yaml
```diff
@@ -218,39 +218,26 @@ repos:
...
+- repo: https://github.com/PyCQA/pylint
+  rev: v2.12.2
```
webknjaz, line 222: This suddenly jumps a few minor versions from `~= 2.8.0` set up in the tox env now. I guess it would have to be a separate PR demonstrating that this bump is okay and does not create problems on its own.

.pre-commit-config.yaml
```diff
@@ -218,39 +218,26 @@ repos:
...
+    additional_dependencies:
+      - ansible-core
+      - ansible-runner
+      - dataclasses; python_version<"3.7"
+      - libtmux
+      - onigurumacffi
+      - pytest
+      - pyyaml
+      - sphinx
+      - setuptools-scm
```
webknjaz, line 234: Let's be consistent about the indentation and unnecessary spaces + comments are needed to let people know where to copy-paste things from as a part of maintenance:
```suggestion
    # NOTE: The entries below must be kept in sync with the contents of `setup.cfg` at all times,
    # NOTE: including `install_requires` and any extras that may be imported in runtime to accommodate
    # NOTE: for `pylint`'s ability to import the modules being checked. They must also include any of the
    # NOTE: test and platform-dependent requirements that may be imported under different circumstances.
    - ansible-core  # dependency imported in utils/catalog_collections.py
    # setup.cfg:
    - ansible-runner ~= 2.0
    - dataclasses; python_version < "3.7"
    - importlib-resources; python_version < "3.9.0"
    - jinja2
    - onigurumacffi
    - pyyaml
    # tests:
    - libtmux
    - pytest
    # docs:
    - sphinx
    - setuptools-scm
```

.pre-commit-config.yaml
```diff
@@ -218,39 +218,26 @@ repos:
...
+    entry: pylint --rcfile=.pylintrc
```
webknjaz, line 235: An explicit `--rcfile` change is unnecessary because it's the default anyway. If it was needed, you'd have to use the `args:` setting instead.
```suggestion
```

tox.ini
```diff
@@ -48,7 +48,6 @@ description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
   pre-commit run -a {posargs}
-  pylint {[base]pkg_name} tests --ignore=tm_tokenize
```
webknjaz, line 51: If you remove this ignore, it must be moved under `.pylintrc`.

webknjaz: Also, don't forget that it's configured under `lint-vetting` in a certain way, to keep that implemented.

.pre-commit-config.yaml
cidrblock: I've been using the PR to find and fix anything that comes up.... If I understand this, this PR can't merge until we know this bump is ok and doesn't create problems on it's own right?

webknjaz: Yeah, and also I'm not sure if stopping to reuse the requirements sources creates more problems than it solves.

tox.ini
cidrblock: Doesn't the exclude in the pre commit yaml accomplish this?

webknjaz: I know that some people integrated linters directly in their editors/IDEs. It would be frustrating if their experience is different. This is the same issue as with flake8. If it's invoked as a VS Code plugin, a contributor would have to know to reimplement settings. But if the config is in a standard location, it's sourced and the preferences are the same.

.pre-commit-config.yaml
```diff
@@ -218,39 +218,34 @@ repos:
...
-      too-many-instance-attributes,
-      too-many-locals,
-      too-many-statements,
-      unused-import,
-      useless-else-on-loop,
-      wrong-import-order,
-    stages:
-    - manual
+    additional_dependencies:
+    # NOTE: The entries below must be kept in sync with the contents of `setup.cfg` at all times,
```
webknjaz, line 226: @ssbarnea remember you asked why I implement a `local` hook for pylint? This is why :point_up:

See also: https://github.com/pre-commit/pre-commit/issues/1963#issuecomment-874073958.

tox.ini
webknjaz: Oh, and as for `exclude`, it mostly exists for pre-commit to decide whether to even invoke the hook when it's integrated as a pre-commit hook in the local Git repo.

.pre-commit-config.yaml
```diff
@@ -218,39 +218,34 @@ repos:
...
+    - importlib-resources; python_version < "3.9.0"
+    - jinja2
+    - onigurumacffi
+    - pyyaml
+    # tests:
+    - libtmux
+    - pytest
+    # docs:
+    - sphinx
+    - setuptools-scm
```
webknjaz, line 243: Re-add the plugin lost by this migration:
```suggestion
    - setuptools-scm
    # plugins:
    - pylint-pytest < 1.1.0
```

.pre-commit-config.yaml
```diff
@@ -218,39 +218,34 @@ repos:
...
+  rev: v2.12.2
   hooks:
   - id: pylint
-    language: system
-    name: PyLint
-    files: \.py$
-    entry: python -m pylint
-    args:
-    - --ignore=tm_tokenize
-    - |-
```
webknjaz, line 230: The removal of this arg could be submitted as a separate PR too.

.pre-commit-config.yaml
cidrblock: What about just leaving everything in a test-requirements file for both linting and testing and pull that from precommit and tox....  It would be an all-in-one like it is today.

cidrblock: (I suggest this not nothing if it's possible :))

.pre-commit-config.yaml
cidrblock: Could this wait until the repo needs it? I think pylint is running clean without it...

.pre-commit-config.yaml
cidrblock: Doesn't the exclude do the same thing?

tox.ini
cidrblock: Oh.... Interesting, pylint ignoring these files is a side effect, but we should explicitly have pylint ignore them if it runs outside of precommit by a dev later?

.pre-commit-config.yaml
cidrblock: also- I don't think, but could be wrong, the current "enforcing" pylint doesn't have this

.pre-commit-config.yaml
cidrblock: Now I see explained above :)

ssbarnea: I am fully aware of the problem with immutable dependencies on pre-commit and lack of support for external dependency files. I knew it for years, bit of  PITA but I kinda get used to it and added a separated list of deps inside pylint and mypy configs to deal with it. It is bit duplicating but I think it worth, in the end project dependencies do not change often and even if they change the linter job will be the first to slap us.

Overall I think that the downsides are considerably smaller than the benefits of running under pre-commit as real hooks, not local ones. So far only two hooks needed extra attention with `additional_dependencies`. 

I only had two cleanup pre-commit cache a couple of times in the last two years as usually you can avoid it by altering the additional_dependencies even with a simple extra version condition to avoid resetting entire cache.

I plan to write a blog post about these challenges and maybe get some feedback from Anthony.

Combining unit-test dependencies with linters in a single `test-requirements.txt` is pretty bad for at least two reasons: it bloats all tox environments by installing more than needed and second, it may prevent you from identifying the correct list of dependencies due to cross-contamination.

Also, if we use local hooks, we cannot use them directly with pre-commit (hook or not) and we cannot run them with pre-commit.ci either. I would try to avoid them when possible.

tox.ini
ssbarnea: Adding ignores just here is ok. In fact my impression was that we want to address these in the future and avoid needing that exclude, right?

You can add them to pylintrc file if we do not plan to remove them, like https://stackoverflow.com/a/8427174/99834

Keep in mind that we need to pass config file to pylint when used as hook or we may be surprised it did not load the config. If you check my PRs you will see that I added it, and that ensures they both use same config.

@cidrblock I need to warn you that is a very small set of cases where adding an exclude in linter config own config is not enough when running linter under pre-commit and thus forcing us to have two sets of excludes. I think it happens very few times, but I just wanted to tell you. I remember that was needed for ansible-lint when the hook definition was with default `pass_filenames: true`. At some point I changed it to `pass_filenames: false` which basically makes pre-commit run it only once instead of once per file. That sorted the problem for ansible-lint and users had to add excludes in a single place. That was also a huge performance benefit as linter has a considerable long startup time due to ansible code... as in making it 9.5x faster.

.pre-commit-config.yaml
webknjaz: It is possible to source the requirements files in tox but not pre-commit. It would mean that this specific hook will be running in the venv that tox creates when it's `local` as pre-commit will not create a venv for this case. And we'll need to keep `manual` because it won't work in pre-commit.ci.
Basically, keep what I did originally.

.pre-commit-config.yaml
webknjaz: It'd like to have a constraints file for this and pre-commit stands in the way of stability if you use non-local hooks.

.pre-commit-config.yaml
webknjaz: We don't have this problem only because the tests are made complicated with classes. Once simplified, it's inevitable.

PR #748, opened 2021-12-14
[lint] Fix lint issues in tests/unit/runner

cidrblock:
additionally add an `__init__.py` so it gets picked up with a vanilla `pylint tests` locally


cidrblock:
recheck

ssbarnea (approved):

PR #750, opened 2021-12-15
Add py36/py37 to zuul gate

ssbarnea:
As we fixed those jobs we need to ensure we also run them inside gate.


ssbarnea:
Also add gate once approved.

cidrblock:
recheck

cidrblock (approved):

PR #751, opened 2021-12-15
[lint] Unused var and import error

cidrblock:
Fix an unused variable in actions/config
Use the same approach in 2 other places. (improves the fix in #746)

Disable the import error in utils, since that directory isn't really a package, but instead catalog_collections is just a script that gets run, so a full or relative import couldn't be used in this case.

### DIFF
--- a/share/ansible_navigator/utils/catalog_collections.py
+++ b/share/ansible_navigator/utils/catalog_collections.py
@@ -29,10 +29,8 @@ try:
 except ImportError:
     from yaml import SafeLoader  # type: ignore
 
-# pylint: disable=import-error
-from key_value_store import KeyValueStore  # type: ignore
+from key_value_store import KeyValueStore  # type: ignore  # pylint: disable=import-error
 
-# pylint: enable=import-error
 
 PROCESSES = (multiprocessing.cpu_count() - 1) or 1
 
--- a/share/ansible_navigator/utils/catalog_collections.py
+++ b/share/ansible_navigator/utils/catalog_collections.py
@@ -31,6 +31,7 @@ except ImportError:
 
 # pylint: disable=import-error
 from key_value_store import KeyValueStore  # type: ignore
+
 # pylint: enable=import-error
 
 PROCESSES = (multiprocessing.cpu_count() - 1) or 1
--- a/share/ansible_navigator/utils/catalog_collections.py
+++ b/share/ansible_navigator/utils/catalog_collections.py
@@ -29,8 +29,9 @@ try:
 except ImportError:
     from yaml import SafeLoader  # type: ignore
 
+# pylint: disable=import-error
 from key_value_store import KeyValueStore  # type: ignore
-
+# pylint: enable=import-error
 
 PROCESSES = (multiprocessing.cpu_count() - 1) or 1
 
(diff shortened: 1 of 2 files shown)


ssbarnea (approved):

cidrblock:
recheck

ssbarnea:
recheck

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -29,8 +29,10 @@
 except ImportError:
     from yaml import SafeLoader  # type: ignore
 
+# pylint: disable=import-error
 from key_value_store import KeyValueStore  # type: ignore
 
+# pylint: enable=import-error
```
webknjaz, line 35: Try to always disable things inline
```suggestion
from key_value_store import KeyValueStore  # type: ignore  # pylint: disable=import-error

```

cidrblock: cool, ty

src/ansible_navigator/actions/config.py
```diff
@@ -321,7 +321,7 @@ def _parse_and_merge(self, list_output, dump_output) -> None:
                         parsed[variable]["via"] = source
                     parsed[variable]["current"] = current
                     parsed[variable]["__current_value"] = extracted.groupdict()["current"]
-                except KeyError as exc:  # noqa: F841
+                except KeyError as _exc:  # noqa: F841
                     self._logger.error("variable '%s' not found in list output")
```
webknjaz, line 325: have you considered just adding `exc_info=exc` or dropping the `as` part?

share/ansible_navigator/utils/catalog_collections.py
webknjaz: A bit more context: comments (any comments) tend to float around the code base. If you add a comment on top of a line, there's no guarantee that in the future, an unrelated line won't appear between a comment and its context. Putting comments on the same line has a better chance of surviving such situation and you won't see cases of misleading comments explaining something that doesn't seem related to the code below (if one line can sneak in, a block of code can too, just give it time).

This is why my rule of thumb is to move the pylint comment above the line only if there are technical limitations that prevent inlining, like linters complaining about the line being too long.

src/ansible_navigator/actions/config.py
cidrblock: ha! I didn't even think of that, I'll do another PR and look for others too.

src/ansible_navigator/actions/config.py
```diff
@@ -321,7 +321,7 @@ def _parse_and_merge(self, list_output, dump_output) -> None:
                         parsed[variable]["via"] = source
                     parsed[variable]["current"] = current
                     parsed[variable]["__current_value"] = extracted.groupdict()["current"]
-                except KeyError as exc:  # noqa: F841
+                except KeyError as _exc:  # noqa: F841
```
webknjaz, line 324: Wouldn't "fixing" the variable mean that `noqa` is no longer needed?

> **F841:** Local variable name is assigned to but never used

cidrblock: yeah, brb

webknjaz: Perhaps, the real fix is
```suggestion
                except KeyError:
```

webknjaz (approved):

PR #753, opened 2021-12-15
[lint] pylint no-member

cidrblock:
This feels like a bug with pylint and py3.10:

```
************* Module conf
docs/conf.py:14:19: E1101: Instance of 'tuple' has no 'resolve' member (no-member)
************* Module single_sourced_data
docs/_ext/single_sourced_data.py:49:14: E1101: Instance of 'tuple' has no 'parent' member (no-member)
```

but, I think the changes here accomplish the same thing as the original and avoid the lint errors.

Thoughts?


docs/conf.py
```diff
@@ -11,7 +11,7 @@
 
 # -- Path setup --------------------------------------------------------------
 
-PROJECT_ROOT_DIR = Path(__file__).parents[1].resolve()
+PROJECT_ROOT_DIR = Path(__file__).resolve().parents[1]
```
webknjaz, line 14: I don't like trying to maintain these "downstream" — it'd need to be synchronized back. pylint has some configuration for declaring unobvious dynamic attrs. Also, this is likely to be runtime-dependent (it's often the case with pylint).

Oh, and IIRC `__file__` is probably already an absolute path. So is the `resolve()` really needed? Not sure how to react to pylint misbehaving, though. It's the linter bug, not ours — should we be trying to accommodate linter problems?

docs/_ext/single_sourced_data.py
```diff
@@ -45,8 +45,8 @@
 ]
 
 
-DOCS_DIR = Path(__file__).parents[1].resolve()
-PROJECT_DIR = DOCS_DIR.parent
+DOCS_DIR = Path(__file__).resolve().parents[1]
```
webknjaz, line 48: This feels a bit unnatural.
It could also be
```suggestion
DOCS_DIR = Path(__file__, '..').resolve()
```
I guess.

docs/_ext/single_sourced_data.py
```diff
@@ -45,8 +45,8 @@
...
+PROJECT_DIR = Path(__file__).resolve().parents[2]
```
webknjaz, line 49: I don't like that it's copy-paste — I specifically tried to avoid it. Otherwise, I would inline this in places where it's being used :man_shrugging:

cidrblock:
@webknjaz just `pre-commit run --all-files` in this case, with the changes proposed in #747

webknjaz:
> @webknjaz just `pre-commit run --all-files` in this case, with the changes proposed in #747

Adding comments there to explain the side-effects of such a change.

cidrblock:
left for another day :) ty @webknjaz

PR #754, opened 2021-12-15
[lint] Fix pylint `no-member` and `not-an-iterable`

cidrblock:
In the case of fields and options, both are dynamically created via a default factory.

The docs for E1101 state:

False positives: This message may report object members that are created dynamically, but exist at the time they are accessed.

http://pylint-messages.wikidot.com/messages:e1101

Although I could not find a mention for not-an-iterable (E1133), the assumption here is that it is also a false positive

### DIFF
--- a/src/ansible_navigator/actions/inventory.py
+++ b/src/ansible_navigator/actions/inventory.py
@@ -423,11 +423,13 @@ class Action(App):
             if form.cancelled:
                 return
 
+            # pylint: disable=not-an-iterable
             inventories = [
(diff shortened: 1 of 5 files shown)


webknjaz:
I guess it's fine to ignore for now but in general, it's best to declare these fields as dynamic in pylintrc and maybe even add an in-tree plugin that changes pylint's perception of these objects.

cidrblock:
hope I'm not reading that wrong...

PR #759, opened 2021-12-16
Unify calling pylint from different toxenvs

webknjaz:
Supersedes #747


cidrblock (approved):
seems sane, and better than what we have.

PR #760, opened 2021-12-16
tox: move lint-vetting section

ssbarnea:
Assure that lint and lint-vetting sections are one after another.


webknjaz (dismissed):
Let's move lint below instead of moving vetting above. This way, it'll be closer to other configs this is synchronized with.

cidrblock (approved):
no pref here, will let you both decide & agree

PR #761, opened 2021-12-16
pylint: consolidate tm_tokenize ignore

ssbarnea:
Avoid configuring same exceptions in two places while also documenting
the reasons why they were added.

This also allows external pylint execution to be in sync with
our linting.


webknjaz (changes requested):
I prefer this to be implemented as https://github.com/ansible/ansible-navigator/pull/759 instead because it also solves the problem related to the CLI option move and follows the formatting that was previously accepted.

PR #763, opened 2021-12-16
Make black run under pre-commit

ssbarnea:
Runs black with **same options** as before under pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,10 +24,10 @@ repos:
     - id: black
       language_version: python3
       args:
-        - -v
-        - --diff
-        - --check
-        - .
+      - -v
+      - --diff
+      - --check
+      - .
 
 - repo: https://github.com/Lucas-C/pre-commit-hooks.git
   rev: v1.1.10
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,6 +18,17 @@ repos:
   - id: isort
     name: Sort import statements using isort
 
+- repo: https://github.com/psf/black
+  rev: 22.1.0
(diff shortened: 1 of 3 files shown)


webknjaz (changes requested):
I think we need to go through the voting about even having it. The current team never did that.

ssbarnea:
Moved to draft to remove it from daily reminder.

cidrblock (approved):
I find having all the linters and formatters run the same way, cleaner, and easier to maintain.

This is not an endorsement of black as a formatter in any direction, simply an agreement that this improves the current configuration.

priyamsahoo (approved):
+1 for this !

.pre-commit-config.yaml
```diff
@@ -18,6 +18,17 @@ repos:
   - id: isort
     name: Sort import statements using isort
 
+- repo: https://github.com/psf/black
```
webknjaz, line 21:
```suggestion
- repo: https://github.com/psf/black.git
```

.pre-commit-config.yaml
```diff
@@ -18,6 +18,17 @@ repos:
...
+  rev: 22.1.0
+  hooks:
+    - id: black
+      language_version: python3
+      args:
+        - -v
+        - --diff
+        - --check
+        - .
```
webknjaz, line 30:
```suggestion
    - id: black
      language_version: python3
      args:
      - -v
      - --diff
      - --check
      - .
```

cidrblock:
@webknjaz 

"I'll keep hating black separately from this change."

Please submit a PR to have black removed from the repository if you feel strongly about it and seek a quorum approval.

PR #767, opened 2022-01-15
Add codecov integration

ssbarnea:


playbooks/post-run.yaml
```diff
@@ -0,0 +1,30 @@
...
+- hosts: all:!appliance
+  tasks:
+    - name: Coverage report
+      environment:
+        CODECOV_TOKEN: "{{ codecov.token }}"
+      command:
+        set -ex
+        pip install 'coverage[toml]'
+        coverage combine .tox/.coverage.*
+        curl -Os https://uploader.codecov.io/latest/linux/codecov
```
webknjaz, line 28: There's a curated executable that ansible-core uses. You should probably use it too.

playbooks/post-run.yaml
```diff
@@ -0,0 +1,30 @@
...
+          UT5ImZ8kwOGFnpvJyJsJy747dJfLiqLSxIgTjL3b4hpHJ/vHhvgaZyGrKeNHHs=
+
+- hosts: all:!appliance
+  tasks:
+    - name: Coverage report
+      environment:
+        CODECOV_TOKEN: "{{ codecov.token }}"
+      command:
+        set -ex
+        pip install 'coverage[toml]'
```
webknjaz, line 26: No need for toml. Coveragepy should use its own config as it's a de-facto practice used across other projects.

playbooks/post-run.yaml
```diff
@@ -0,0 +1,30 @@
...
+        coverage combine .tox/.coverage.*
```
webknjaz, line 27: This shouldn't be necessary unless tox itself is executed in parallel mode. AFAIK only tox is.

tox.ini
```diff
@@ -30,6 +30,7 @@ commands =
   /bin/bash -c 'py.test -vvvv -n `python -c "import multiprocessing; import os; print(int(os.environ.get(\"XDIST_CPU\") or multiprocessing.cpu_count()))"` --dist=loadfile --maxfail=15 --durations=100 --cov ansible_navigator --cov share --cov-report term-missing --cov-branch --showlocals'
 setenv =
   TERM = xterm-256color
+  COVERAGE_FILE = {env:COVERAGE_FILE:{toxworkdir}/.coverage.{envname}}
```
webknjaz, line 33: I don't think this is necessary. I suppose you may be looking for an XML report. In which case, add `--cov-report xml` maybe?

ssbarnea: Not needed for codecov itselt but it helps decluttering project root, so we don't add extra temp files to it.

ssbarnea: Removed it as is not directly needed by codecov.

.github/workflows/tox.yml
```diff
@@ -66,6 +66,14 @@ jobs:
           python3 -m tox
         env:
           TOXENV: ${{ matrix.tox_env }}
+      - name: Upload coverage data
+        uses: codecov/codecov-action@v2
+        with:
+          directory: .tox
+          files: 'coverage-*.xml'
+          name: ${{ matrix.tox_env }}
```
webknjaz, line 74: I also recommend adding flags — they are useful for determining which jobs/envs covered specific lines.

tox.ini
webknjaz: It is, unless you configure it to use a different report format IIRC

webknjaz (changes requested):
Urgh... Yet another unapproved merge 🤦🏼

PR #768, opened 2022-01-17
chore: add release drafting and label checks

ssbarnea:
Ensure that we log draft changes on https://github.com/ansible/ansible-navigator/releases
so we will know what is included in next release and be able to produce
more informative releases.

These files are the same as other projects we own, molecule,
ansible-lint, ansible-compat as they rely on shared configuration from
devtools project.


webknjaz (changes requested):
We're going for towncrier right now, don't you remember?

Issue #772, opened 2022-01-19
Unpin `astroid` in `tox.ini`


cidrblock:
astroid is a dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
attempt to remove when astroid 2.9.4 or later is released
related: https://github.com/PyCQA/pylint/issues/5131

Sample errors:

```
src/ansible_navigator/actions/run.py:42:0: E0611: No name 'round_half_up' in module 'ansible_navigator.utils' (no-name-in-module)
************* Module ansible_navigator.actions.write_file
src/ansible_navigator/actions/write_file.py:11:0: E0611: No name 'remove_dbl_un' in module 'ansible_navigator.utils' (no-name-in-module)
************* Module ansible_navigator.actions.template
src/ansible_navigator/actions/template.py:14:0: E0611: No name 'remove_dbl_un' in module 'ansible_navigator.utils' (no-name-in-module)
src/ansible_navigator/actions/template.py:15:0: E0611: No name 'templar' in module 'ansible_navigator.utils' (no-name-in-module)
************* Module ansible_navigator.image_manager.inspector
src/ansible_navigator/image_manager/inspector.py:11:0: E0611: No name 'pascal_to_snake' in module 'ansible_navigator.utils' (no-name-in-module)
************* Module ansible_navigator.image_manager.puller
src/ansible_navigator/image_manager/puller.py:8:0: E0611: No name 'ExitMessage' in module 'ansible_navigator.utils' (no-name-in-module)
src/ansible_navigator/image_manager/puller.py:9:0: E0611: No name 'ExitPrefix' in module 'ansible_navigator.utils' (no-name-in-module)
src/ansible_navigator/image_manager/puller.py:10:0: E0611: No name 'LogMessage' in module 'ansible_navigator.utils' (no-name-in-module)
************* Module tests.unit.test_image_introspection
tests/unit/test_image_introspection.py:48:47: E1101: Module 'ansible_navigator.utils' has no 'get_share_directory' member (no-member)
************* Module tests.unit.test_utils
tests/unit/test_utils.py:28:39: E1101: Module 'ansible_navigator.utils' has no 'find_settings_file' member (no-member)
tests/unit/test_utils.py:29:38: E1101: Module 'ansible_navigator.utils' has no 'oxfordcomma' member (no-member)
tests/unit/test_utils.py:42:39: E1101: Module 'ansible_navigator.utils' has no 'find_settings_file' member (no-member)
[...]

PR #773, opened 2022-01-19
pin `astroid` due to false `no-name-in-module` and `no-member` errors

cidrblock:
Related: https://github.com/ansible/ansible-navigator/issues/772

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -74,6 +74,11 @@ commands =
 deps =
   {[testenv]deps}
   -r{toxinidir}/docs/requirements.in
+  # astroid is a dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
+  # attempt to remove this when astroid 2.9.4 or later is released
+  # related: https://github.com/PyCQA/pylint/issues/5131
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -74,6 +74,11 @@ commands =
 deps =
   {[testenv]deps}
   -r{toxinidir}/docs/requirements.in
+  # astroid is a dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
```
webknjaz, line 77: Since it's an indirect dependency, please move it to the end of the list and state that it is transitive. I'd also add an entry line to make the distinction clear.

tox.ini
```diff
@@ -74,8 +74,16 @@ commands =
 deps =
   {[testenv]deps}
   -r{toxinidir}/docs/requirements.in
+  # direct dependencies
   pylint
   pylint-pytest < 1.1.0
+  # indirect dependencies
```
webknjaz, line 80: That block separator I asked for:
```suggestion

  # indirect dependencies
```

webknjaz (approved):

PR #777, opened 2022-01-20
Sort import statements within project

cidrblock:
- Adds isort configuration to `pyproject.toml` (a default location for isort configuration)
- Removes all I00* exemptions from `.flake8`
- Removes the all file exclusions from precommit (it remains a vetting/manual check)
- Updates import statements across all files as found using isort via `tox -e lint-vetting`

After several iterations this configuration for `isort` was found to most closely align with the current sort order and minimize the number of changes.

Note:
- Future PRs issues can/may/should/could be raised to revisit the `isort` configuration
- A future PR will be put in to move the `pre-commit` config out of `tox lint-vetting` to `tox lint` and modify the contributors guide
- See the `pyproject.toml` file for an explanation of each setting
- The `reverse_relative` settings will be remove in a follow-up PR, issue #780 

Closes #715


.pre-commit-config.yaml
```diff
@@ -19,13 +19,12 @@ repos:
   hooks:
   - id: isort
     args:
+    # Keep in sync with pyproject.toml
```
webknjaz, line 22: Instead, move this to `.isort.cfg`. This is what I prefer normally: https://github.com/ansible/pylibssh/blob/devel/.isort.cfg.

.pre-commit-config.yaml
```diff
@@ -19,13 +19,12 @@ repos:
...
+    - --force-single-line-imports
     - --honor-noqa
-    exclude: |
-      (?x)
-      ^
-        .*
-      $
-    stages: ["manual"]
+    - --lines-between-types=1
+    - --profile=black
```
webknjaz, line 26: I don't like this implicit thing that hides what this really means behind the internals. Look into the example above to see how it is possible to be more verbose about the settings intentions.
Another point is that using a proper per-tool annotated config allows to track the motivation and expectation behind the settings with obscured names.

.pre-commit-config.yaml
```diff
@@ -19,13 +19,12 @@ repos:
...
+    - --reverse-relative
```
webknjaz, line 27: What is this?

.pre-commit-config.yaml
```diff
@@ -19,13 +19,12 @@ repos:
   hooks:
   - id: isort
     args:
+    # Keep in sync with pyproject.toml
+    - --force-single-line-imports
```
webknjaz, line 23: Does this seem like a subjective choice? Why not use tables with mandatory trailing commas?

docs/_ext/single_sourced_data.py
```diff
@@ -6,26 +6,30 @@
 from copy import copy
 from pathlib import Path
 from re import match
-from typing import Any, Dict, List, Tuple, Union
+from typing import Any
```
webknjaz, line 9: This file should be excluded from linting because it is a copy from external repositories and it's easier to keep such files in-sync across the board when they don't produce severe merge conflicts.

webknjaz: Scratch that. I confused it with another file.

.pre-commit-config.yaml
cidrblock: I saw both worked, I'm curious why the preference?

ssbarnea: In fact look at https://github.com/ansible-community/molecule/blob/main/pyproject.toml#L24-L26 -- which will also ensure you will not have surprises from black. No real need to add yet another dot config file in repository root. AFAIK there is no clear bug that would prevent us from using pyproject.toml

docs/conf.py
```diff
@@ -8,7 +8,6 @@
 
 from setuptools_scm import get_version
 
-
```
webknjaz, line 11: This one is synchronized more often tho.

.pre-commit-config.yaml
ssbarnea: I would also prefer to avoid it and only use options for which we have clear reasoning.

pyproject.toml
```diff
@@ -49,6 +49,13 @@
 [tool.black]
 line-length = 100
 
+[tool.isort]
```
webknjaz, line 52: It's better not to put this in a shared config. Sections from multiple tools tend to snowball and are hard to locate when looking at the repo root.

.pre-commit-config.yaml
cidrblock: I take it you would prefer to have the individual lines that make up the black profile listed?  The downside here is that the burden of keeping our isort configuration compatible with black falls us.  I don't have an issue with this, but will want to add a comment that noting the configuration needs to be kept in sync with the black profile, tracking it's source.

ssbarnea: I agree with @webknjaz here, we should have zero arguments for isort inside pre-commit config. That is more important than you think because we want to ensure that even if you call isort from outside it will have the same behavior. vscode has the ability to run isort on save and you don't want it to sort files differently than pre-commit.

In most cases any arguments/config in pre-commit is a potential issue (very few exceptions).

.pre-commit-config.yaml
webknjaz, line 22: Also, feedback on writing code comments in general: try to add some context — the minimum and essential amount of information + an external reference for going deeper (if the reader would want to only). This one looks like a TODO item. But it lacks the motivation behind the request — why is it necessary? Does something not work? Does it work differently than one would expect? Is it unintuitive? Should I try to use `git blame` and do some PR/issue archeology each time I see this to figure out if this is still relevant or was even correct in the first place?

pyproject.toml
ssbarnea: It is ok, especially for isort which will not grow. If a tool needs big/growing/verbose config keeping its config separated may be considered.

.pre-commit-config.yaml
ssbarnea: In fact I am almost sure that isort from precommit just respects the pyproject.toml settings. Maybe @cidrblock forgot to stage current change to the config and was surprised to get different behavior from pre-commit?

docs/conf.py
webknjaz: (not that it needs to be excluded)

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py
```diff
@@ -1,6 +1,8 @@
 """An ansible test lookup plugin."""
 
-from __future__ import absolute_import, division, print_function
```
webknjaz, line 3: Do we run `ansible-test sanity` against these anywhere? If we do (or will), this'll break the boilerplate linting.

.pre-commit-config.yaml
cidrblock: the pattern was adopted early on, and the config was intended to minimize the changes, in an attempt to focus enabling isort and less so on the style.

My preference here would be to stay this course since the pattern is in place, and revisit it later.

this comes to mind: https://www.python.org/dev/peps/pep-0008/#imports

cidrblock: BTW, I am happy to raise an issue to revisit the isort configuration so it is not lost or forgotten.

.pre-commit-config.yaml
webknjaz: @cidrblock explained at https://github.com/ansible/ansible-navigator/pull/777/files#r788706028. TL;DR it's hard to find where the configs are un a dump of sections from different tools that may also be present in other places — with dedicated files, it's easier for everybody to look up where the related settings live.

.pre-commit-config.yaml
cidrblock: I think you mean the reverse-relative? "Reverse order of relative imports."  

the default does:
...grandparent
..parent
.sibling

this reverses that

.sibling
..parent
...grandparent

it was does to minimize the changes in the repo, I didn't capture the exact count.  

My preference here would be to revist the config after it is in place, minimizing the current changes now.

I'm happy to log an issue here, so it is not lost or forgotten

.pre-commit-config.yaml
webknjaz: > this comes to mind: [python.org/dev/peps/pep-0008/#imports](https://www.python.org/dev/peps/pep-0008/#imports)

That section allows `from ... import <multiple things>` which is what I meant. I didn't mean that bare imports should be inlined.

.pre-commit-config.yaml
webknjaz: > I take it you would prefer to have the individual lines that make up the black profile listed?

With our improvements, though. OTOH, it would probably be fine if we keep profile + things that are important to keep explicit.

.pre-commit-config.yaml
cidrblock: yep, that must have been it.  comments removed, thanks @ssbarnea @webknjaz (good points about intent in the comments)

.pre-commit-config.yaml
webknjaz: Just make sure to explain the motivation. It probably makes sense to have it so that there are fewer potential import loops.

pyproject.toml
webknjaz: It is not okay if you consider git merge driven updates from different upstreams.

pyproject.toml
```diff
@@ -49,6 +49,12 @@
...
+force_single_line = true
+honor_noqa = true
+lines_between_types = 1
```
webknjaz, line 55: What is "types" in this context?

.pre-commit-config.yaml
cidrblock: config remove from pre-commit entirely, will keep the profile for now, until it becomes a problem, thx

pyproject.toml
```diff
@@ -49,6 +49,12 @@
...
+profile = "black"
+reverse_relative = true
```
webknjaz, line 57: This one needs motivation.

pyproject.toml
```diff
@@ -49,6 +49,12 @@
 [tool.black]
 line-length = 100
 
+[tool.isort]
+force_single_line = true
+honor_noqa = true
```
webknjaz, line 54: Could you try dropping this? Originally, I've added it to my repos when fighting some isort version upgrade compatibility issues but I don't remember why anymore. So if you find out why it's needed, maybe add a code comment about that too.

src/ansible_navigator/configuration_subsystem/parser.py
```diff
@@ -1,20 +1,18 @@
 """Build the argument parser
 """
 
+from argparse import SUPPRESS
```
webknjaz, line 4: This is weird... Is it sorted like this because it's all-uppercase?

.pre-commit-config.yaml
cidrblock: Because the current pattern in place is for single line from statements, I'd prefer to keep it that way, revisit later, and focus on getting this in place.  It will eventually be one less thing that needs to be done manually, which frees up time to work on issues and features that effect the user experience.

src/ansible_navigator/runner/command_async.py
```diff
@@ -7,6 +7,7 @@
 """
 
 from queue import Queue
+
 from ansible_runner import run_command_async  # type: ignore[import]
```
webknjaz, line 11: It'd be nice to have a section declared in the isort config that would list the Ansible-land dependencies. As in things from our ecosystem/colleagues. Separately from the rest of the third-party libs.

pyproject.toml
cidrblock: I understand both locations are functionally equivalent and from a little research both commonly used.  Rather than block this from moving forward due to the location, I'm going to leave it as is.  If in the future problems arise with it in pyproject.toml, it can be easily moved.

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py
webknjaz: I'm actually not entirely sure if it enforces all of these on the same line but it's worth checking.

pyproject.toml
cidrblock:
```
import re
import subprocess
import sys

from collections import Counter
from collections import OrderedDict
from datetime import datetime

File: tests/integration/actions/doc/test_welcome_interactive_noee.py [hunk 14]
```diff
@@ -6,7 +6,6 @@
 
 from .base import BaseClass
 
-
```
webknjaz, line 9: There's an option to make this post-import two-line separator consistent regardless of the var assignment expressions being present. I'd prefer for us to use that.

src/ansible_navigator/configuration_subsystem/parser.py
cidrblock: yeah, I assumed so when I saw it

tests/unit/configuration_subsystem/test_invalid_params.py
```diff
@@ -8,6 +8,7 @@
 
 from ansible_navigator.configuration_subsystem import NavigatorConfiguration
 from ansible_navigator.utils import ExitMessage
+
```
webknjaz, line 11: I think we should use the option that drops the separator between the local and first-party imports.

pyproject.toml
webknjaz: Within the same section? Why is it needed then?

cidrblock: the goal here was to minimize the number of changes in the repo, more than happy to revisit in the future, but leaving it here for now since the pattern is in place.

webknjaz: Sounds like this line could use a context comment so that it doesn't force the future archeologists to guess what motivated its addition and if that judgment is still true.

tests/integration/actions/doc/test_welcome_interactive_noee.py
cidrblock: nice, adding.

pyproject.toml
```diff
@@ -49,6 +49,11 @@
 [tool.black]
 line-length = 100
 
+[tool.isort]
```
webknjaz, line 52: Missing config options that should be added:
```suggestion
[tool.isort]
default_section = THIRDPARTY
indent = 4
known_first_party = ansible_navigator, key_value_store
known_testing = pytest, unittest
lines_after_imports = 2  # Ensures consistency for cases when there's variable vs function/class definitions after imports
no_lines_before = LOCALFOLDER  # Keeps local imports bundled with first-party
sections = FUTURE, STDLIB, TESTING, THIRDPARTY, FIRSTPARTY, LOCALFOLDER
use_parentheses = true  # Instead of escaping EOL
```

cidrblock: I'll add these one at a time, per commit to see the changes.

docs/_ext/single_sourced_data.py
```diff
@@ -6,22 +6,28 @@
...
 from sphinx.application import Sphinx
 from sphinx.util import logging
 from sphinx.util.docutils import SphinxDirective
-from sphinx.util.nodes import nested_parse_with_titles, nodes
+from sphinx.util.nodes import nested_parse_with_titles
+from sphinx.util.nodes import nodes
 
-# isort: split
 
-from docutils import statemachine
```
webknjaz, line 18: This is something that is supposed to be close to the sphinx imports (IIRC Sphinx extends/patches docutils so its imports should happen later). It's third-party too, though.

Can we have another split? I don't remember if that would work.

pyproject.toml
webknjaz: Why did you mark this as resolved? Nothing changed.

pyproject.toml
webknjaz: Note that some are coupled (you cannot inject a section that is not declared, for example).

pyproject.toml
cidrblock: darn, I didn't intend to.... was just buzzing through and must have accidentally.  Good call, will add comment!

pyproject.toml
cidrblock: default_section = THIRDPARTY appears to be the default https://pycqa.github.io/isort/docs/configuration/options.html#default-section

cidrblock: indent = 4, makes no changes in repo, leaving it out until such time it is needed

cidrblock: use_parentheses = true , makes no changes in repo, leaving it out until such time it is needed

cidrblock: known_first_party = "ansible_navigator, key_value_store", makes no changes in repo, leaving out

tests/unit/configuration_subsystem/test_invalid_params.py
cidrblock: added: 
no_lines_before = "LOCALFOLDER"  # Keeps local imports bundled with first-party

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py
cidrblock: yeah, it complains......  skipping the file.

pyproject.toml
webknjaz: > known_first_party = "ansible_navigator, key_value_store", makes no changes in repo, leaving out

Please keep it. Probably there's no cases of confusion with third-party imports which may still happen.

cidrblock: It would appear isort is correctly determining that `ansible_navigator` is a first party though.... and the probability of the kv store causing an issue should be really low because of that directories use-case.

I'm happy to add it back if you feel it is important, but I think we should note that at the time of implementation it had no effect and provided no utility

docs/_ext/single_sourced_data.py
cidrblock: @webknjaz have a look, moved it up and added an ignore

pyproject.toml
webknjaz: Do you have a plan for automatically noticing when it stops being aligned? Formatters tend to make people ignore such problems so no hope of noticing them during manual human reviews.

src/ansible_navigator/runner/command_async.py
cidrblock: @webknjaz  I'm not sure what you change/action you are suggesting here..... can you elaborate?

pyproject.toml
webknjaz: There's also value in documenting the expectations transparently. Besides, when you integrate new linters, there may be no violations of the rules they bring but it's usually not a good reason for not linting things, right?

docs/_ext/single_sourced_data.py
```diff
@@ -6,22 +6,29 @@
...
 
 from sphinx.application import Sphinx
 from sphinx.util import logging
 from sphinx.util.docutils import SphinxDirective
-from sphinx.util.nodes import nested_parse_with_titles, nodes
+from sphinx.util.nodes import nested_parse_with_titles
+from sphinx.util.nodes import nodes
+
+
+from docutils import statemachine # isort: skip
```
webknjaz, line 22:
```suggestion
from docutils import statemachine  # isort: skip
```

pyproject.toml
cidrblock: After our discussion, I understood we will have a follow on PR immediately after this one to remove `reverse_relative` not that the circular deps checks are in place. @webknjaz I see this as resolved can you hit the button?

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py
webknjaz: So this should be reverted before it's resolved, then.

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/modules/mod_1.py
```diff
@@ -1,7 +1,11 @@
-"""An ansible test module."""
+"""An ansible test module.
+
+   isort:skip_file
```
webknjaz, line 3: Can this be a comment? We need to avoid putting identifiers in docstrings. Or any other things that don't contribute to the explanation of what the module is about for that matter.

pyproject.toml
cidrblock: @webknjaz 

The line now include an explanation of what it does and why it is there

lines_between_types = 1 # Separate import/from with 1 line, minimizing changes at time of implementation

I see this as resolved, might you please click and resolve?

docs/_ext/single_sourced_data.py
webknjaz: Okay, looks solved, except that the comment should always have at least two spaces before the start — don't the linters complain about this?

.pre-commit-config.yaml
cidrblock: @ssbarnea @webknjaz The line now reads:

force_single_line = true # Force from .. import to be 1 per line, minimizing changes at time of implementation

I see this as resolved and am comfortable with the configuration as it is.

Can you resolve this please?

.pre-commit-config.yaml
cidrblock: @webknjaz This was logged and linked so it is not forgotten in the future:

https://github.com/ansible/ansible-navigator/issues/780

I see this as resolved, can you close it or follow-up with remaining concerns/requested changes?

pyproject.toml
cidrblock: related to this: https://github.com/ansible/ansible-navigator/issues/780

docs/_ext/single_sourced_data.py
cidrblock: errrr! :) thx

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/modules/mod_1.py
cidrblock: I put it there as ansible-test still complained about it missing when it had a trailing comment. This appeared to be the next best place. Based on the docs, I saw only the 2 possible locations outside of the config: https://github.com/PyCQA/isort#skip-processing-of-imports-outside-of-configuration

.pre-commit-config.yaml
cidrblock: Since both work fine and there doesn't seem to be a technical reason, leaving it `pyproject.toml` for now.  When/If something comes up where it makes sense to move it, it should not be a large effort.

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py
cidrblock: ha, I reverted the module but missed the lookup! thanks

tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/modules/mod_1.py
cidrblock: oh, I stand corrected, there are many ways: https://pycqa.github.io/isort/docs/configuration/action_comments.html

cidrblock: I've skipped the entire collections folder in the configuration, since it just a fixture and all files would need to be excluded

docs/_ext/single_sourced_data.py
cidrblock: flake8 would have had I run it
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

docs/_ext/single_sourced_data.py:22:34: E261 at least two spaces before inline comment
from docutils import statemachine # isort: skip
                                 ^
1     E261 at least two spaces before inline comment
1
```
```

File: pyproject.toml (hunk 16 above)
Comment (cidrblock) [reply pyproject.toml:52]: For the sake of history, can you provide an example of when in the future this might become necessary while I add it back?

Comment (cidrblock) [reply pyproject.toml:52]: added it:

```
known_first_party = "ansible_navigator, key_value_store" # No effect at implementation, here anticipating future benefit

```

File: tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py (hunk 8 above)
Comment (cidrblock) [reply tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py:3]: I've reverted skipped the entire collections folder in the configuration, since it is just a fixture and all files would need to be excluded.  thanks for catching this. @webknjaz Can you resolve this?

File: pyproject.toml (hunk 16 above)
Comment (cidrblock) [reply pyproject.toml:52]: `known_tests` has been removed, here's an example:

with: 
```
import os
import shlex

import pytest

from unittest import mock
```

without
```
import os
import shlex

from unittest import mock

import pytest
```

with this setting `pylint` throws: 
```
************* Module tests.unit.test_cli
tests/unit/test_cli.py:10:0: C0411: standard import "from unittest.mock import patch" should be placed before "import pytest" (wrong-import-order)
************* Module tests.unit.configuration_subsystem.test_configurator
tests/unit/configuration_subsystem/test_configurator.py:7:0: C0411: standard import "from unittest import mock" should be placed before "import pytest" (wrong-import-order)
tests/unit/configuration_subsystem/test_configurator.py:8:0: C0411: standard import "from unittest.mock import patch" should be placed before "import pytest" (wrong-import-order)
```
It's being removed to avoid the need to modify pylint.

It can be added at such time it is necessary in the future.

@webknjaz thanks for the suggestion here, it was a good test.

Comment (webknjaz) [reply pyproject.toml:52]: I suppose we just need to disable this rule in pylint. I've found an old issue that I've apparently created a while back that has the explanation that pylint uses isort under the hood: https://github.com/PyCQA/pylint/issues/3817#issuecomment-687892090.

I still think that having all test frameworks/libs grouped is benefitial.

Comment (webknjaz) [reply pyproject.toml:52]: > For the sake of history, can you provide an example of when in the future this might become necessary while I add it back?

I think that one of the major benefits is reproducibility as isort implementation changes and also the mere fact that it is helpful as documentation of "we expect that these modules/packages are local". Explicit is better than implicit: PEP 20 did not invent this principle out of nowhere. It's a good observation that when you want things to be readable/maintainable, presenting such information to people in an easily findable manner is polite and transparent. Otherwise, we'd be basically sending anybody who'd want to understand how/why it works like this to the side quest of trying to read the source code of isort — this is the price for having implicit undocumented assumptions that nobody knows about except for the implemented at the moment of initial addition.

Comment (webknjaz) [reply pyproject.toml:52]: @cidrblock look what I've found in the FAQ: https://pylint.pycqa.org/en/latest/faq.html#i-am-using-another-popular-linter-alongside-pylint-which-messages-should-i-disable-to-avoid-duplicates.

File: .pre-commit-config.yaml (hunk 3 above)
Comment (cidrblock) [reply .pre-commit-config.yaml]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback

File: .pre-commit-config.yaml (hunk 4 above)
Comment (cidrblock) [reply .pre-commit-config.yaml]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback

File: tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py (hunk 8 above)
Comment (cidrblock) [reply tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py:3]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback

File: pyproject.toml (hunk 9 above)
Comment (cidrblock) [reply pyproject.toml]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback

File: src/ansible_navigator/runner/command_async.py (hunk 13 above)
Comment (cidrblock) [reply src/ansible_navigator/runner/command_async.py:11]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback. I am curious about this suggestion, let's continue the idea as an issue/feature request in the future.

File: tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py (hunk 8 above)
Comment (webknjaz) [reply tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py:3]: Looks good. Thanks for letting me check this :)

File: pyproject.toml (hunk 16 above)
Comment (cidrblock) [reply pyproject.toml:52]: each has been vetted, and the final list is in the configuration, resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback. Some good ideas here and example of different config options.

File: tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/modules/mod_1.py (hunk 19 above)
Comment (cidrblock) [reply tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/modules/mod_1.py]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback

File: pyproject.toml (hunk 10 above)
Comment (cidrblock) [reply pyproject.toml]: resolving conversation, please reopen if the concerns and question are unanswered, ty for the feedback.  The follow-up PR was noted in the PR description and issue logged.

File: tests/fixtures/common/collections/ansible_collections/testorg/coll_1/plugins/lookup/lookup_1.py [hunk 20]
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+
```
webknjaz, line 5: Since you've excluded the entire directory and isort no longer manages this file, do you want to also revert this change across the directory to exclude it from the PR? Or do you want to still keep it?

cidrblock: It's a good point, since isort is ignoring the entire directory, changes to these files should be left out of the PR.  thanks, keen eye :)

src/ansible_navigator/runner/command_async.py
webknjaz: https://github.com/ansible/ansible-navigator/issues/781

pyproject.toml
webknjaz: In the last comments, I was asking to keep the testing section and disable the respective pylint rule per the official FAQ entry I've linked above.

cidrblock: but with the current implementation it is not necessary as there are no conflicts...

cidrblock: thinking aloud here, because this PR does not enforce `isort` disabling import-order in pylint now would create a period of time between this PR and a follow on PR where neither was enforce and could introduce additional errors in the repo.  I'm thinking it might be better to disable pylint import order checking is necessary and after isort via pre-commit is enforceing.

cidrblock: and in the case the team does not approve using isort as enforcing, changes to pylint will not need to be made or backed out

webknjaz: https://github.com/ansible/ansible-navigator/issues/784

webknjaz: https://github.com/ansible/ansible-navigator/issues/785

webknjaz (approved):
I've filed https://github.com/ansible/ansible-navigator/issues/784 and https://github.com/ansible/ansible-navigator/issues/785 for follow-ups so this should be good to go now.

PR #782, opened 2022-01-21
Activate `pylint-pytest` plugin

webknjaz:
This patch is necessary because `pylint` does not autoactivate third-
party plugins just because they are installed in the same virtualenv.

It seems like this bit of the original configuration got lost over
time: the Python distribution package is present in the dependencies
and gets installed but `pylint` does not use it. With this change, it
becomes used again.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,10 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
     # tm_tokenize is virtually vendored and shouldn't be linted as such
     ignore = "tm_tokenize"
 
+    load-plugins = [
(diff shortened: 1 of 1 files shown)


webknjaz:
(not expecting reviews because it's a correction of a misconfiguration that was expected to work earlier)

cidrblock:
I'd like an opportunity to review this, even if it is a correction
 Can you please request reviews? I know it's might seem silly, but it is still a change to the code base.

webknjaz:
@cidrblock sure, feel free to re-gate as well. Looks like Zuul was drunk.

cidrblock (approved):

PR #787, opened 2022-01-21
Changes making the use of double quotes consistent throughout

cidrblock:
- Changes in 2 files to switch to double quotes from single

which in turn allows for:

This allows the [flake8-quotes](https://github.com/zheller/flake8-quotes) `inline-quote` style to be a double quote, eliminating the individual `Q000` entries throughout the `flake8` config file.

`flake8-quotes` is a linter pulled in by the we-make-styleguide which is a linter being vetted.

Although this change does not enforce the use of double quotes, it will allow users of the `tox -e lint-vetting` to see warning messages that double quotes are use consistently throughout the repo and a single quote is being introduced. 

Some good conversation here about this: https://issuehunt.io/r/wemake-services/wemake-python-styleguide/issues/681

Closes #757


ssbarnea (approved):

.flake8
```diff
@@ -48,6 +48,8 @@ extend-ignore =
 # https://wemake-python-stylegui.de/en/latest/pages/usage/formatter.html
 #format = wemake
 
+inline-quotes = "
```
webknjaz, line 51: This is coming from a third-party plugin and all of such settings are maintained at the end of the file.

docs/_ext/spelling_stub_ext.py
```diff
@@ -22,9 +22,9 @@ def run(self) -> List[nodes.Node]:
 
 def setup(app: Sphinx) -> Dict[str, Union[bool, str]]:
     """Initialize the extension."""
-    app.add_directive('spelling', SpellingNoOpDirective)
+    app.add_directive("spelling", SpellingNoOpDirective)
```
webknjaz, line 25: It was probably a bad idea to format a file that is copied over from other projects.

PR #788, opened 2022-01-21
Changes to make the use of trailing commas consistent throughout the project

cidrblock:
This modifies the changed files such that the use of trailing commas is consistent through out the project.

Which in turn allows all `C812` exceptions to be removed from the `flake8` configuration file which are a result of a linter being vetted called `flake8-commas` which is part of the wemake style guide.

Although this does not enforce the use of trailing commas, it will allow users of `tox -e lint-vetting` to use it to both add trailing commas where they are missing and see errors that indicate a missing trailing comma which would be inconsistent with the repository.  

The changes were made by the precommit formatter called `add-trailing-comma` followed by `black`.  Only one file needed to be manually editted to break a long string.

```
tests/integration/_tmux_session.py:277:101: E501 line too long (101 > 100 characters)
                    f"******** ERROR: TMUX '{err_message}' TIMEOUT @ {elapsed}s @ {tstamp} ********",
``` 
(the string was simply broken in half)


webknjaz:
> With black active we don't need another formatter (add-trailing-comma) that does what black already does anyway.

It does not do _exactly_ the same. `add-trailing-comma` is _a bit_ stricter but does not create conflicts with black. I think we need it integrated because it solves the trailing commas problem in a nicer way (per my experience).

ssbarnea:
recheck : why py37+ all failed, very weird as I did not see any real bytecode change with that... or maybe there is one.

webknjaz (approved):
Brilliant!

PR #789, opened 2022-01-22
Remove unnecessary generator inside inventory max modified times fuction

cidrblock:
The function that finds the maximum file modification time for file and directory based inventories had an unnecessary generator nested inside it.

This in turn eliminates a reported "missing trailing comma" error reported during the use of `tox -e lint-vetting`.

### DIFF
--- a/src/ansible_navigator/actions/inventory.py
+++ b/src/ansible_navigator/actions/inventory.py
@@ -142,10 +142,8 @@ class Action(App):
             if os.path.isdir(inventory):
                 mtimes.append(
                     max(
-                        (
-                            os.path.getmtime(e)
-                            for e in glob.glob(os.path.join(inventory, "**"), recursive=True)
-                        )
+                        os.path.getmtime(e)
+                        for e in glob.glob(os.path.join(inventory, "**"), recursive=True)
                     ),
                 )
             elif os.path.isfile(inventory):


cidrblock:
recheck

webknjaz:
@cidrblock don't re-trigger the https://github.com/apps/softwarefactory-project-zuul jobs — we've seen that they are broken yesterday, remember? We shouldn't care about that one because it's misconfigured at the moment, and is non-blocking.

webknjaz:
Oh, I see it was a legitimate failure in the correct CI.

src/ansible_navigator/actions/inventory.py
```diff
@@ -142,10 +142,8 @@ def _set_inventories_mtime(self) -> None:
             if os.path.isdir(inventory):
                 mtimes.append(
                     max(
-                        (
-                            os.path.getmtime(e)
-                            for e in glob.glob(os.path.join(inventory, "**"), recursive=True)
-                        )
+                        os.path.getmtime(e)
```
webknjaz, line 145: As a possible (not really in the scope, though) improvement — use longer variable names. Even in comprehensions/generator expressions: just looking at this patch, I can't immediately guess what's `e` here, for example. It would be nice to see a more meaningful name.

webknjaz (approved):

src/ansible_navigator/actions/inventory.py
webknjaz: (resolving so that it doesn't block gating)

PR #790, opened 2022-01-22
Form button validation, pass a `List` not `Generator`

cidrblock:
Form buttons should really only have one of 2 validators, either `all_true` in the case of a `Submit` button, or `none` in the case of a `Cancel` button.

This changes results in passing a `List` to the button's conditional validator rather than a `Generator`.

Several typing updates to correct what is being passed through.

This also eliminates a  trailing-comma error reported when using `tox -e lint-vetting`

### DIFF
--- a/src/ansible_navigator/ui_framework/field_button.py
+++ b/src/ansible_navigator/ui_framework/field_button.py
@@ -2,6 +2,7 @@
 """
 from dataclasses import dataclass
 from typing import Callable
+from typing import List
 from typing import Union
 
 from .curses_window import Window
@@ -29,7 +30,7 @@ class FieldButton:
         """no default to add into the prompt for checkbox"""
         return ""
 
-    def validate(self, response: "FieldButton") -> None:
+    def validate(self, response: List) -> None:
         """validate this instance"""
         validation = self.validator(response)
         if validation.error_msg:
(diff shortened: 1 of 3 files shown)


ssbarnea:
recheck

src/ansible_navigator/ui_framework/field_button.py
```diff
@@ -29,15 +30,15 @@ def full_prompt(self) -> str:
         """no default to add into the prompt for checkbox"""
         return ""
 
-    def validate(self, response: "FieldButton") -> None:
+    def validate(self, response: List) -> None:
```
ssbarnea, line 33: Usually is a bad idea to use container objects without a type for the contained things. 

It is a list of what? I cannot really guess from the signature and description. Can you make it more specific?

cidrblock: good catch, thanks, working on it

cidrblock: thanks again, in this case it should be a list of either Unknown or a bool indicating the validaoty of the other fields in the form

cidrblock:
recheck

ssbarnea (approved):
While I am not pleased with the overly complex type used I do find this as an improvement over older code, so approving it. As a note for the future, we should avoid using Unions.

Why we need the Unknown sentinel? Cannot we make use of None for that?
`Union[Unknown, bool]` -> `Union[Optional[bool]]` ?

PR #792, opened 2022-01-23
Disable warnings related to the use of assert in the tests directory as reported with `lint-vetting`

cidrblock:
This adds a few additional files to the `per-file-ignore` list within the flake8 configuration with an exemption for `S101` which is reported by `flake8-bandit` as part of the wemake style guide which is being vetted.

The list was also alphabetized to ensure new entries were placed in a sane location.

A note was added to the top of the flake8 per-file-ignores indicating that the glob for the tests folder ignoring s101 can be added later. Currently there is a conflict between globs and per-file entries.  See https://github.com/PyCQA/flake8/issues/670

This eliminates the `S101` entry from the output of `tox -e lint-vetting` so users of it can see messages only related to the changes they are making and not for files that were previously added to the repository:

```diff
1     C812 missing trailing comma
1     RST304 Unknown interpreted text role "mod".
- 13    S101 Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
1     S404 Consider possible security implications associated with the subprocess module.
1     S603 subprocess call - check for execution of untrusted input.
2     S604 Function call with shell=True parameter identified, possible security issue.
2     WPS111 Found too short name: p < 2
2     WPS115 Found upper-case constant in a class: KEGEX
3     WPS201 Found module with too many imports: 13 > 12
6     WPS210 Found too many local variables: 6 > 5
3     WPS220 Found too deep nesting: 24 > 20
2     WPS221 Found line with high Jones Complexity: 16 > 14
3     WPS226 Found string literal over-use: test_option > 3
9     WPS305 Found `f` string
1     WPS316 Found context manager with too many assignments
6     WPS317 Found incorrect multi-line parameters
2     WPS323 Found `%` string formatting
4     WPS324 Found inconsistent `return` statement
2     WPS414 Found incorrect unpacking target
1     WPS430 Found nested function: getcwd
14    WPS436 Found protected module import: _curses
3     WPS450 Found protected object import: _CursesWindow
1     WPS602 Found using `@staticmethod`
```

Related #713

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -70,7 +70,8 @@ per-file-ignores =
   # S101: Allow the use of assert within the tests directory
   # tests/*: S101
 
-  # The following were present during the initial implemetation
+  # The following were present during the initial implementation.
+  # They are expected to be fixed and unignored over time.
   # Please try to keep it in alphbetical order to make finding files easier for the next person
 
   docs/_ext/single_sourced_data.py: D400, D403, DAR101, DAR201, N817, P103, WPS110, WPS111, WPS201, WPS202, WPS210, WPS213, WPS221, WPS229, WPS231, WPS237, WPS305, WPS318, WPS323, WPS331, WPS336, WPS347, WPS407, WPS432, WPS440, WPS441, WPS453, WPS504
--- a/.flake8
+++ b/.flake8
@@ -63,8 +63,9 @@ max-line-length = 100
 per-file-ignores =
   # The following ignores have been researched and should be considered permanent
   # each should be preceeded with an explanation of each of the error codes
-  # When ignores are added for a specific file, these will need to be added to that line as well.
-  
+  # If other ignores are added for a specific file in the section following this,
+  # these will need to be added to that line as well.
+
   # S101: Allow the use of assert within the tests directory
   tests/**.py: S101
 
@@ -267,7 +268,6 @@ per-file-ignores =
   tests/unit/__init__.py: D104
   tests/unit/actions/__init__.py: D104
   tests/unit/actions/test_config.py: D200, D400, D403, S101, WPS110, WPS204, WPS218, WPS221, WPS226, WPS305
-  tests/unit/actions/test_exec.py: S101
   tests/unit/actions/test_inventory.py: D200, D400, D403, S101, WPS110, WPS204, WPS218, WPS221, WPS226, WPS305
   tests/unit/actions/test_run.py: D200, D400, DAR101, DAR201, N813, PT019, S101, S108, WPS110, WPS121, WPS122, WPS201, WPS226, WPS237, WPS305, WPS425, WPS432, WPS437
   tests/unit/configuration_subsystem/__init__.py: D104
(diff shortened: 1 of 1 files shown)


cidrblock:
recheck

.flake8
```diff
@@ -60,14 +60,32 @@ max-line-length = 100
...
 per-file-ignores =
-  docs/conf.py: WPS221, WPS226, WPS301, WPS305, WPS323, WPS420, WPS429, WPS433
+  # The following ignores have been researched and should be considered permanent
+  # each should be preceeded with an explanation of each of the error codes
+  # These might be commented out as they overlap with the perfile ignores
+  # once the per file ignores are removed or added within the file itself
+  # they should be able to be uncommented
+
+  # S101: Allow the use of assert within the tests directory
+  # tests/*: S101
```
webknjaz, line 70: Just uncomment it right now. The duplication won't make a difference. But when the per-file ignores are 
```suggestion
  tests/**.py: S101
```

Here's what I recommend usually — have all of the ignores necessary and explain why it makes sense to ignore each in a comment on top of that section: https://github.com/ansible-community/ansible-lint/blob/718e4cd/.flake8#L168.

.flake8
```diff
@@ -60,14 +60,32 @@ max-line-length = 100
...
+
+  # The following were present during the initial implemetation
+  # Please try to keep it in alphbetical order to make finding files easier for the next person
+
   docs/_ext/single_sourced_data.py: D400, D403, DAR101, DAR201, N817, P103, WPS110, WPS111, WPS201, WPS202, WPS210, WPS213, WPS221, WPS229, WPS231, WPS237, WPS305, WPS318, WPS323, WPS331, WPS336, WPS347, WPS407, WPS432, WPS440, WPS441, WPS453, WPS504
   docs/_ext/spelling_stub_ext.py: DAR101, DAR201
+  docs/conf.py: WPS221, WPS226, WPS301, WPS305, WPS323, WPS420, WPS429, WPS433
```
webknjaz, line 77: This doesn't look related to asserts or tests. So I'd argue that this part should be submitted separately because it doesn't belong to the atomic effort of setting up the ignores for the tests.

.flake8
```diff
@@ -60,14 +60,32 @@ max-line-length = 100
...
+  # each should be preceeded with an explanation of each of the error codes
+  # These might be commented out as they overlap with the perfile ignores
+  # once the per file ignores are removed or added within the file itself
+  # they should be able to be uncommented
+
+  # S101: Allow the use of assert within the tests directory
+  # tests/*: S101
+
+  # The following were present during the initial implemetation
+  # Please try to keep it in alphbetical order to make finding files easier for the next person
```
webknjaz, line 73: // I wonder if codespell can lint this
```suggestion
  # Please try to keep it in alphabetical order to make finding files easier for the next person
```

.flake8
```diff
@@ -60,14 +60,32 @@ max-line-length = 100
...
+  # The following ignores have been researched and should be considered permanent
+  # each should be preceeded with an explanation of each of the error codes
+  # These might be commented out as they overlap with the perfile ignores
+  # once the per file ignores are removed or added within the file itself
+  # they should be able to be uncommented
+
+  # S101: Allow the use of assert within the tests directory
+  # tests/*: S101
+
+  # The following were present during the initial implemetation
```
webknjaz, line 72: (plus missing "n")
```suggestion
  # The following were present during the initial implementation. They are expected to be fixed and unignored over time.
```

.flake8
cidrblock: I'll work with whomever originally implemented the list to see if it can be sorted correctly.

cidrblock: See: https://github.com/ansible/ansible-navigator/pull/794

webknjaz:
@cidrblock FYI in diffs there's usually some indentation in the unchanged lines.
It's (like in typical diff output)
```diff
  unchanged
- dropped
+ added
  lines
```
vs (your version)
```diff
unchanged
- dropped
+ added
lines
```
The idea is that all the text content has the same starting point and is aligned accordingly.
I know that it's inconvenient to do in GH comment inputs so I often produce such diffs externally to make it easier. It's possible to do by creating two text files `before` and `after` and feeding them to the diff like `diff -u before after`. Alternatively, text editors like Vim allow producing block indents. So I add two spaces to all the lines like this and then replace some of the spaces with `-`/`+` and  duplicate them where necessary.

webknjaz (approved):

PR #799, opened 2022-01-25
Configure pylint to run as native pre-commit hook

ssbarnea:
This changes the `pylint` entry configured for pre-commit to be run as a native pre-commit hook rather than a local hook.

local hooks are described here:  https://pre-commit.com/#repository-local-hooks

Pros:
- pylint will now run automatically as a pre-commit hook for contributors that have installed the pre-commit hooks with `pre-commit install`
-  the version of the pre-commit hook as specified in the `rev` key is a reference to the version of the hook and in this case specifies the `pylint` version to use
- In the future, if necessary, the python version can also be specified in the pre-commit configuration if differences are found between the hook being run locally and via pre-commit

Cons:
- due to the nature of pre-commit hooks and caching, the dependencies and sometime their versions for pylint to run need to be specified in the pre-commit configuration in addition to the tox.ini as well as the test-requirements.txt, This comes with operational overhead when they need to set or modified.
- there is documentation in the dev version of `pylint` that suggests the hook only works with a local installation of `pylint`. It is unknown today exactly what this might mean in the future: https://pylint.pycqa.org/en/latest/user_guide/pre-commit-integration.html

If it becomes necessary to revert to a local hook in the future, the amount of work to do so should not be major, it would basically be a fail forward PR effectively reverting this one.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -251,21 +251,30 @@ repos:
     - src/
     - tests/
 
-- repo: local
+- repo: https://github.com/pycqa/pylint.git
+  rev: v2.12.2
   hooks:
   - id: pylint
-    language: system
-    name: PyLint
-    files: \.py$
-    entry: python -m pylint
     args:
     # FIXME: uncomment once it doesn't have violations
     # - docs/
     - share/
     - src/
     - tests/
+    additional_dependencies:
+    - ansible-core
+    # astroid is a transitive dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
+    # attempt to remove this when astroid 2.9.4 or later is released
+    # related: https://github.com/PyCQA/pylint/issues/5131
+    # related: https://github.com/ansible/ansible-navigator/issues/772
+    - astroid == 2.9.0
+    - ansible-runner
+    - jinja2
+    - libtmux
+    - onigurumacffi
+    - pylint-pytest < 1.1.0
+    - setuptools-scm
+    - sphinx
     pass_filenames: false
-    stages:
-    - manual
 
 ...
--- a/tox.ini
+++ b/tox.ini
@@ -61,7 +61,7 @@ allowlist_externals = ansible-playbook
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
-  pre-commit run {--show-diff-on-failure \
+  pre-commit run {posargs:--show-diff-on-failure \
     --hook-stage manual \
     --all-files -v}
   black -v --diff --check {toxinidir}
--- a/tox.ini
+++ b/tox.ini
@@ -61,29 +61,13 @@ allowlist_externals = ansible-playbook
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
-  pre-commit run -a {posargs}
-
-  {envpython} -m \
-    pre_commit run \
-    {posargs:pylint \
-    --show-diff-on-failure \
+  pre-commit run {--show-diff-on-failure \
     --hook-stage manual \
     --all-files -v}
-
   black -v --diff --check {toxinidir}
 deps =
   {[testenv]deps}
   -r{toxinidir}/docs/requirements.in
-  # direct dependencies
-  pylint
-  pylint-pytest < 1.1.0
-
-  # indirect dependencies
-  # astroid is a transitive dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
-  # attempt to remove this when astroid 2.9.4 or later is released
-  # related: https://github.com/PyCQA/pylint/issues/5131
-  # related: https://github.com/ansible/ansible-navigator/issues/772
-  astroid == 2.9.0
 setenv =
   # NOTE: Running some of the MyPy checks is disabled here but they may
   # NOTE: be reconfigured in the pre-commit configuration file. So the


.pre-commit-config.yaml
```diff
@@ -209,21 +209,24 @@ repos:
     - tests/
     pass_filenames: false
 
-- repo: local
+- repo: https://github.com/pre-commit/mirrors-pylint
```
webknjaz, line 212:
```suggestion
- repo: https://github.com/pre-commit/mirrors-pylint.git
```
(consistency)

pyproject.toml
```diff
@@ -60,10 +60,14 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
...
     [tool.pylint.master]
-    ignore="tm_tokenize"
+    ignore="tm_tokenize" # virtually vendored and shouldn't be linted as such
 
     [tool.pylint.messages_control]
-    disable = "C0330, C0326"
+    disable = [
+      "duplicate-code",
+      "unsubscriptable-object",
+      "fixme"
```
webknjaz, line 69: This should have trailing commas on every line.

.pre-commit-config.yaml
```diff
@@ -209,21 +209,24 @@ repos:
...
-    files: \.py$
-    entry: python -m pylint
     args:
     # FIXME: uncomment once it doesn't have violations
     # - docs/
     - share/
     - src/
     - tests/
+    additional_dependencies:
+      - ansible-core
```
webknjaz, line 223: This introduces differing indentation, keep it the same as above.
Also, what is the justification for maintaining a copy-paste of the dependency list and loosing the ability to pin the transitive deps via constraints/lock files?

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -17,6 +17,8 @@
 from ..utils import oxfordcomma
 from .definitions import ApplicationConfiguration
 from .definitions import Constants as C
+
+# pylint: disable=deprecated-module
```
webknjaz, line 21: Why is it deprecated in the first place?

ssbarnea: I do not know but @cidrblock is investigating.

ssbarnea: Sorted by switching to official pylint repo and last stable release.

.pre-commit-config.yaml
webknjaz: Reopening since this was marked resolved without any reason or explanation.

tox.ini
```diff
@@ -61,29 +61,13 @@ allowlist_externals = ansible-playbook
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
 commands =
-  pre-commit run -a {posargs}
-
-  {envpython} -m \
-    pre_commit run \
-    {posargs:pylint \
-    --show-diff-on-failure \
+  pre-commit run {--show-diff-on-failure \
```
webknjaz, line 64:
```suggestion
  pre-commit run {posargs:--show-diff-on-failure \
```

webknjaz (changes requested):
Blocking until https://github.com/ansible/ansible-navigator/pull/799#discussion_r792007395 that keeps being marked by @ssbarnea as resolved will actually start being discussed.

ssbarnea:
> Blocking until [#799 (comment)](https://github.com/ansible/ansible-navigator/pull/799#discussion_r792007395) that keeps being marked by @ssbarnea as resolved will actually start being discussed.

Can you please be explicit because I am **sure** there is no comment that applies to the current code. You had lots of comments which were resolved.

.pre-commit-config.yaml
cidrblock: it looks like the entry is now:

```
- repo: https://github.com/pycqa/pylint.git
```

pyproject.toml
cidrblock: this file isn't modified in the Pr anymore, I think this was from the original implementation, 

I think it was covered by: https://github.com/ansible/ansible-navigator/pull/800

.pre-commit-config.yaml
cidrblock: these appear to be indented like other entries in the file:
```
    additional_dependencies:
    - ansible-core
    # astroid is a transitive dep for pylint, > 2.9.0 introduces false no-name-in-module and no-member errors
    # attempt to remove this when astroid 2.9.4 or later is released
    # related: https://github.com/PyCQA/pylint/issues/5131
    # related: https://github.com/ansible/ansible-navigator/issues/772
    - astroid == 2.9.0
    - ansible-runner
    - jinja2
    - libtmux
    - onigurumacffi
    - pylint-pytest < 1.1.0
    - setuptools-scm
    - sphinx
```

.pre-commit-config.yaml
webknjaz: It is :+1: 
(although, I'd prefer to also keep the case of `PyCQA` in the URL too but that's minor)

pyproject.toml
webknjaz: Yep.

src/ansible_navigator/configuration_subsystem/configurator.py
cidrblock: I don't see this in the current PR, if I recall this was a bug in pylint related to the deprecation of `Parser` in 3.9 but was fixed in a later pylint release  (we have a `Parser` too, unique to navigator)

.pre-commit-config.yaml
webknjaz: I know. Although, the second part of the comment was never addressed.

tox.ini
cidrblock: looks like this is now:
```
  pre-commit run {posargs:--show-diff-on-failure \
```

webknjaz:
> Can you please be explicit because I am **sure** there is no comment that applies to the current code. You had lots of comments which were resolved.

When you follow that link, this is what you'll see (I've copied it for you):
> This introduces differing indentation, keep it the same as above.
> Also, _**what is the justification for maintaining a copy-paste of the dependency list and loosing the ability to pin the transitive deps via constraints/lock files?**_

I've emphasized the ignored but most important part of the comment.

cidrblock (approved):
looks good from here

webknjaz:
The PR description is rendered as "No description provided." meaning that it does not address the concern either.

.pre-commit-config.yaml
cidrblock: researching, so I grok it

cidrblock: Is this because we are using the native pre-commit hook rather than local which would use the `test-requirements` file?

webknjaz: That's exactly my concern — the inability to pin the whole dependency tree to avoid transitive dependencies causing problems, deeper than the direct one (like `astroid`). Why is it better to lose this ability?

cidrblock: By using the native pre-commit hook, and specifying a version there, does it isolate use from the python version on the runner and therefore make it consistent until the precommit native hook version is changed?  I did see some inconsistencies with #797 (running locally vs zuul) I suspected were because of this.

cidrblock: This is a good read, explaining why pre-commit will never use the requirements file, here for future ref: https://github.com/pre-commit/pre-commit/issues/730

cidrblock: I should note, the list of dependencies for navigator should remain small, and new deps rarely added and only after a lot of conversation and consideration.  While this doesn't avoid the version issue, the list should remain small.

cidrblock: Should we consider making a note here and in the test-requirements/tox file that they need to remain in sync?

ssbarnea:
Well that second part was missed. Still, even reading it now I find it outside the scope of the PR.

We did not had any pining before, we don't lose ability to pin after. We use  pre-commit for all linters and I want to be able to use it directly, something that is do almost always before exploding a change. Having it as a manual check that works only from tox makes little to no sense.

local pre-commit hooks are problematic as the tool is unable to install them.

webknjaz:
> Well that second part was missed. Still, even reading it now I find it outside the scope of the PR.

This PR existence by itself goes against those concerns and the original intent of setting it up like this. So it is most definitely in the scope. At least a legitimate reason for blocking this PR.

webknjaz:
> We did not had any pining before, we don't lose ability to pin after.

It was on my list. And losing the ability has nothing to do with the prior presence of the constraints.

webknjaz:
> Having it as a manual check that works only from tox makes little to no sense.

It does, right after starting to take the reproducibility into the consideration.

webknjaz:
> We use pre-commit for all linters and I want to be able to use it directly, something that is do almost always before exploding a change.

Using something is not a good enough reason to actively prevent reproducibility, is it? It's definitely more important not to have random problems for unknown reasons reproducible in only specific env.

webknjaz:
> local pre-commit hooks are problematic as the tool is unable to install them.

That depends on the definition of what's problematic. Having non-reproducible linting runs is arguably more problematic.

cidrblock:
Does having this as a native pre-commit hook include it in the list of checks run by pre-commit when it is installed? `pre-commit install`  If this is the case, that would be one benefit.  I've recently installed the pre-commit hooks and appreciate the checks automatically happening prior to the commit....

webknjaz:
@cidrblock okay, if it's clearly documented in the PR description/commit that there are benefits and disadvantages, and that some of the problems were decided to be acceptable, I can live with that. But it should be transparent and it would be useful to also explain how to revert/fix the problems once they happen so that the tribal knowledge doesn't only live in a few people's heads.

.pre-commit-config.yaml
webknjaz: Feel free to resolve this once the last comment in the main thread is addressed.

cidrblock:
@webknjaz This is perfectly reasonable, a great idea, and no one should expect less. I particularly like the idea of having trade off mentioned. I'll remember to ask for that when it's not clear.

cidrblock:
@ssbarnea I'll add a better description with some pros/cons and rollback details in the AM if you don't beat me to it.

.pre-commit-config.yaml
ssbarnea: > @ssbarnea I'll add a better description with some pros/cons and rollback details in the AM if you don't beat me to it.

Feel free to alter it and merge it. As long my `alias pc=pre-commit run -a` includes execution of pylint, we are all ok.

As a note, we already had this setup in all the other repos and it was fairly stable. From linting perspective version of dependencies is far less important than testing code. I am in favour of pinning for unit and worked with Sviat to do it.

I already think we spent too much time on simple change like this, so I will let you deal the final blow ;)

> @ssbarnea I'll add a better description with some pros/cons and rollback details in the AM if you don't beat me to it.

Feel free to alter it and merge it. As long my `alias pc=pre-commit run -a` includes execution of pylint, we are all ok.

As a note, we already had this setup in all the other repos and it was fairly stable. From linting perspective version of dependencies is far less important than testing code. I am in favour of pinning for unit and worked with Sviat to do it.

I already think we spent too much time/effort on simple change like this, so I will @cidrblock polish it and merge it so we avoid increasing the thread on this change.

cidrblock:
@ssbarnea @webknjaz Can you review the updated description for accuracy?  I think I understand this, at least now more than I did a couple months ago.

PR #800, opened 2022-01-25
Changes necessary to have the `pylint` configuration in only one location

cidrblock:
The resulting configuration in `pyproject.toml` should reflect what is needed now.

Having it in two places was confusing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,10 +60,11 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
 
 [tool.pylint]
     [tool.pylint.master]
-    ignore="tm_tokenize"
+    # tm_tokenize is virtually vendored and shouldn't be linted as such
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -60,10 +60,11 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
 
 [tool.pylint]
     [tool.pylint.master]
-    ignore="tm_tokenize"
+    # tm_tokenize is virtually vendored and shouldn't be linted as such
+    ignore = "tm_tokenize"
 
     [tool.pylint.messages_control]
-    disable = "C0330, C0326"
+    disable = "duplicate-code,unsubscriptable-object,fixme"
```
webknjaz, line 67: It's a list and should be kept as a list, not a string with things.

cidrblock: on it!

PR #804, opened 2022-01-26
POC for running mypy with multiple python versions

ssbarnea:
That should be seen as an alternative to #726 that avoids cluttering
the mypy hook configuration.

Closes: https://github.com/ansible/ansible-navigator/pull/726/files

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -44,6 +44,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0  # needed by setuptools-scm
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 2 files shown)


webknjaz (changes requested):
Changing the runtime is a bad idea because it relies on the external interpreter being of a specific version. MyPy is designed to work with the AST which is runtime-agnostic as explained in #726, making this approach pointless compared to using `--python-version` properly.

ssbarnea:
> Changing the runtime is a bad idea because it relies on the external interpreter being of a specific version. MyPy is designed to work with the AST which is runtime-agnostic as explained in #726, making this approach pointless compared to using `--python-version` properly.

Read https://github.com/ansible/ansible-navigator/pull/726#issuecomment-1023010608 -- it took only a couple of hours to get a proof that mypy outcome depends on python interpreter and that CI failure came from a change originating from pre-commit.ci auto-update, not a human. A minor change that failed on zuul.

webknjaz:
> it took only a couple of hours to get a proof that mypy outcome depends on python interpreter

Look closer into that log and the PR — that commit does not contain split mypy hook changes. It's been failing prior to that improvement before the PR got merged. It only proves that there were problems before my PR and now those problems are fixed as a result.

cidrblock:
moved to draft because I believe it was implimented

PR #806, opened 2022-01-26
Rename `shell` to `use_shell` in one unit test

cidrblock:
Given the possibility of some future static analysis tool misinterpreting `shell=True` as a subprocess parameter/call, it was changed here to read `use_shell`.

This also eliminates the 2 false positives being reported by `flake8-bandit` which is part of the wemake style guide being vetted.

```
tests/unit/actions/test_exec.py:34:1: S604 Function call with shell=True parameter identified, possible security issue.
        shell=True,
^
tests/unit/actions/test_exec.py:49:1: S604 Function call with shell=True parameter identified, possible security issue.
        shell=True,
^
```

### DIFF
--- a/tests/unit/actions/test_exec.py
+++ b/tests/unit/actions/test_exec.py
@@ -13,7 +13,7 @@ class CommandTestData(NamedTuple):
 
     name: str
     command: str
-    shell: bool
+    use_shell: bool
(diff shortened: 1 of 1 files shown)


cidrblock:
recheck

PR #808, opened 2022-01-27
Pin python version used by pre-commit

ssbarnea:
This ensures that all environments listed below should produce the same
results:
- github actions lint job
- zuul lint job
- pre-commit.ci
- local run via `tox -e lint`

The root cause for doing is that we observed various outcomes from mypy but we already know that this can also impact other linters. Once we add a runtime requirement for pre-commit for one python version it makes sense to just make it the default python for all hook.

Related: #726


webknjaz (commented):
If we're doing this, I would suggest adding a small self-validation script that would compare the values across the CI setups.

ssbarnea:
> If we're doing this, I would suggest adding a small self-validation script that would compare the values across the CI setups.

There is no need for such script as pre-commit with throw error if it cannot find requested python version.

Still, I will keep this PR as draft until we identify a case where this would prevent an error or confusion. If any of you @webknjaz @cidrblock do spot a misbehavior, likely with mypy or pylint, try this one again and see if it improves the developer experience. Once we have we can undraft and merge it.

I am aware that each approach has its ups and downs, so let's decide on what seems easier for us.

webknjaz:
I still think it's useful to have it because the worker VMs come with pre-installed interpreters that may be used silently and it's possible that what is specified in YAML will not be used. This can easily lead to confusing a reader. The configs exist for humans as well, otherwise, we'd be using binary files if there was no need to understand what's in them.

ssbarnea:
> I still think it's useful to have it because the worker VMs come with pre-installed interpreters that may be used silently and it's possible that what is specified in YAML will not be used. This can easily lead to confusing a reader. The configs exist for humans as well, otherwise, we'd be using binary files if there was no need to understand what's in them.



That confuses me as you disapprove making it draft but you also did not approve it. If you support it, lets just merge it and move on.

webknjaz:
I don't actively support it but I still can leave some feedback without rejecting. Also, I sometimes post these from phone and the mobile app does not let me create detached comments like the web UI. This is why I'm sometimes forced to use the review feature by the app in which case I select a neutral comment type and not an approval or rejection.

webknjaz (approved):
FWIW I'm approving to unblock you but don't treat this as an active endorsement of not having a validation script. I just don't care enough to block.

PR #811, opened 2022-01-27
Fix wrong check condition on GHA

ssbarnea:
Removes condition that was part of all-green implementation, which is not used in our case. We do not want check to always run, even after its dependencies failed.


webknjaz (approved):
(demo approval — DNM)

PR #813, opened 2022-01-27
Fix the GHA `check` job to treat skips as failures

webknjaz:
Closes #811


webknjaz:
I've enabled `auto-merge` to (partially) demonstrate that this works when all of the branch protection rules are taken into account.

ssbarnea (approved):

PR #814, opened 2022-01-27
test: reduce verbosity for container pulling

ssbarnea:


cidrblock (approved):
seems reasonable, and easy to add back if we have pulling problems... ty

PR #818, opened 2022-01-28
Add `__all__` to subsystem `__init__` files

cidrblock:
<s>Depends-On: #822
Depends-On: #835</s> 

Although it is not common practice within the code to import everything from a subsystem, it does appear to be common practice to include a `__all__` in each of these `__init__` files.

This is mentioned in several places here: https://www.python.org/dev/peps/pep-0008/  (search for `__all__`)

This is the statement that caught my attention: 
```
To better support introspection, modules should explicitly declare the names in their public API using the __all__ attribute. Setting __all__ to an empty list indicates that the module has no public API.
```
 
This also eliminates the need for the global `F401` exception for flake8 which is described here, and is handled in #830 https://flake8.pycqa.org/en/2.6.0/warnings.html, which should allow future contributors to"see" and act on the error when reported.

```
module imported but unused
```

The lists were put in alphabetical order to make future additions easier, not sure if it matters here, please LMK


webknjaz (commented):
> The lists were put in alphabetical order to make future additions easier, not sure if it matters here, please LMK

It's okay, I think I even saw a linter or something for ordering them like that. Probably in the pre-commit ecosystem.

One ask I have here, though is to not use lists but replace them with tuples. Not sure why it's not linted but there's a general recommendation to have `__all__` immutable. Could you update this?

cidrblock:
> > The lists were put in alphabetical order to make future additions easier, not sure if it matters here, please LMK
> 
> It's okay, I think I even saw a linter or something for ordering them like that. Probably in the pre-commit ecosystem.
> 
> One ask I have here, though is to not use lists but replace them with tuples. Not sure why it's not linted but there's a general recommendation to have `__all__` immutable. Could you update this?


for my education, can you provide a link to the general recommendation?  The only thing I've been able to find it PEP8 which shows:

```
__all__ = ['a', 'b', 'c']
```
and states:

```
Setting __all__ to an empty list indicates that the module has no public API.
```
which leads me to believe `__all__` was expected to be a list.

BTW, I've gone ahead and made the changes, until we need it to be mutable/generated I see no difference and have no preference.

webknjaz:
I did some googling. It seems like in the past this caused some problems with pydocstyle (formerly known as pep257) — https://github.com/PyCQA/pydocstyle/issues/62.
Some people think that lists are semantically more correct. I don't buy it. `__all__` is supposed to be declarative, making it generated/dynamic defeats the whole purpose of having it. Immutable structures make it easier to reason about stuff in a predictable way.
Of course, it's even more correct not to expose those imports via `__init__` and put them into `api` submodules but that requires some extra work and the rule that could check for this code smell is currently disabled.

ssbarnea:
**Converted to draft** as reviews should stay as draft until they pass CI. That prevents spamming reviewers or confuse them on which pull-requests they can review and merge. Feel free to convert it both ways as many time you need as nothing is lost when you do that.

webknjaz:
Don't do this. It's perfectly reviewable. A passing CI is not a prerequisite for human reviews. The protection blocks merges separately.

webknjaz:
@cidrblock is it me or did the vetting linters start running in the enforced lint stage? I can't check right now from phone but I'll try to take a look at that later today when I get to the laptop.

cidrblock:
> @cidrblock is it me or did the vetting linters start running in the enforced lint stage? I can't check right now from phone but I'll try to take a look at that later today when I get to the laptop.

fixed in #835

ssbarnea (approved):

PR #819, opened 2022-01-28
Remove link in `actions/run.py` to Interaction

cidrblock:
This caused an error in #818 

```
/home/docs/checkouts/readthedocs.org/user_builds/ansible-navigator/envs/818/lib/python3.8/site-packages/ansible_navigator/actions/run.py:docstring of ansible_navigator.actions.run.Action._prepare_to_quit:: WARNING: more than one target found for cross-reference 'Interaction': ansible_navigator.ui_framework.Interaction, ansible_navigator.ui_framework.ui.Interaction
```

Eventually we would like to carry over the type hints in the signature, so this should be necessary.  (The link to Interaction and None does currently work, it just has to clicked above the return)

### DIFF
--- a/src/ansible_navigator/actions/run.py
+++ b/src/ansible_navigator/actions/run.py
@@ -254,8 +254,7 @@ class Action(App):
 
         :param interaction: The interaction from the user
         :param app: The app instance
-        :return: The pending :class:`~ansible_navigator.ui_framework.ui.Interaction` or
-            :data:`None`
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

cidrblock:
It should be noted this did not resolve the error because it was a change in the wrong place.  

But, it'll stay in because the ref was not necessary as explained in the description

PR #820, opened 2022-01-28
[DNM] Revising the approach to code base improvements through linting/formatting

cidrblock:
DNM: An alternative approach is being worked on that should accomplish the same thing.
Related #825 

Depends-On #818
Depends-On #819 

This limits the linters that are being vetted to just 2, both related to the quality of the documentation, which I believe makes the code easier to contribute to and troubleshoot.

https://github.com/PyCQA/pydocstyle
https://github.com/terrencepreilly/darglint

Although I have not gone through each error in great detail, I do find the suggestions here constructive, actionable, and generally sane, and believe contributors will be willing to act on them in future PR when warnings from lint-vetting are added to the review.

All unneeded exceptions and extensions have been removed.

I also think it would be best to limit the number of linters being vetted to just one or 2, as I find it easier to act on and remediate.  

I will continue to follow up with additional PRs that improve our documentation, and therefore limit the number of ignore.  I suspect this process, if done alone, will take 1-2 months. At which time I believe we will be able to see the improvements in the code, and provide feedback based on experience and examples through a quorum process.

I am grateful for work that has brought us this far, we are in a position to identity and better understand where our efforts can be put to make a material difference in the project.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -38,15 +38,6 @@ extend-exclude =
 # Completely and unconditionally ignore the following errors:
 extend-ignore =
   E203,  # skipped because it annoys black
-  F401,  # skipped for unknown reason
-  WPS300,  # "Found local folder import" -- nothing bad about this
-  # An opposite consistency expectation is currently enforced
-  # by pylint via: useless-object-inheritance (R0205):
-  WPS306,  # "Found class without a base class: *"
-  WPS326,  # "Forbid implicit string concatenation."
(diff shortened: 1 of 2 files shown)


webknjaz (commented):
As discussed in a call, here's an alternative approach with a smaller patch and a bit more flexibility: https://github.com/ansible/ansible-navigator/pull/825.

Issue #812, opened 2022-01-27
[tests] Periodic test failure after switch from zuul to GHA


cidrblock:
Please log them here with URL, PR number, and summary information when they occur so a trend can be found if one exists

Related #758

cidrblock:
FAILED tests/integration/actions/templar/test_direct_interactive_noee.py::Test::test[9-:open {{ task_path }}-goto vi]
FAILED tests/integration/actions/templar/test_welcome_interactive_ee.py::Test::test[6-:doc-doc for task]
FAILED tests/integration/actions/templar/test_welcome_interactive_ee.py::Test::test[8-:back-show doc]
FAILED tests/integration/actions/templar/test_welcome_interactive_ee.py::Test::test[9-:back-show task]
FAILED tests/integration/actions/templar/test_welcome_interactive_ee.py::Test::test[10-:open {{ task_path }}-goto vi]

(PR unknown)

ssbarnea:
```
FAILED tests/integration/actions/templar/test_direct_interactive_ee.py::Test::test[7-:back-show doc]

```

cidrblock:
In #814 
```
FAILED tests/integration/actions/templar/test_direct_interactive_ee.py::Test::test[9-:open {{ task_path }}-goto vi]
FAILED tests/integration/actions/templar/test_direct_interactive_noee.py::Test::test[9-:open {{ task_path }}-goto vi]
FAILED tests/integration/actions/templar/test_welcome_interactive_ee.py::Test::test[8-:back-show doc]
```
```
=================================== FAILURES ===================================
__________________ Test.test[9-:open {{ task_path }}-goto vi] __________________
[gw0] linux -- Python 3.10.2 /home/runner/work/ansible-navigator/ansible-navigator/.tox/py310/bin/python

self = <tests.integration.actions.templar.test_direct_interactive_ee.Test object at 0x7f0910292dd0>
request = <FixtureRequest for <Function test[9-:open {{ task_path }}-goto vi]>>
tmux_session = <tests.integration._tmux_session.TmuxSession object at 0x7f0910291bd0>
step = Step(user_input=':open {{ task_path }}', comment='goto vi', look_fors=['---'], look_nots=[], mask=True, playbook_status=None, step_index=9, search_within_response='---')

    def test(self, request, tmux_session, step):
        # pylint: disable=too-many-branches
        # pylint: disable=too-many-locals
        """test interactive and ``stdout`` mode ``config``"""
    
[...]

PR #823, opened 2022-01-28
[tests] Changes to frequently failing templar integration tests

cidrblock:
Noticed in #812 , many of the test failures were related to the templar integration tests.

Several changes were made, hoping to reduce the number of failures

1) Exit vi after opening it, I don't suspect this made a difference, but it seem like a polite thing to do and was left in
2) Added a new local to the base class called, unmasked output, because we log locals, we should now see both the masked output and unmasked output from the tmux session if this happens again
3) Changes the search_within_response and look_for for the vi test to seach and look for something that was not previously on the screen, I suspect we were assuming vi had started when in fact we simply caught the `---` from the previous TASK screen

Fixtures were updated and added needed

Note: the files that were changes in the `filename:` line can be ignored, that is not used for comparison and is a byproduct of env differences on the developer machine were the fixtures were built.  It should be sanitized at some point later....

### DIFF
--- a/tests/fixtures/integration/actions/templar/test_direct_interactive_ee.py/test/10.json
+++ b/tests/fixtures/integration/actions/templar/test_direct_interactive_ee.py/test/10.json
@@ -1,0 +1,35 @@
+{
+    "name": "test[10-:q!-exit vi]",
+    "index": 10,
+    "comment": "exit vi",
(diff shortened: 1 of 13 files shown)


webknjaz (approved):
Haven't looked into this too deeply but seems reasonable.

PR #825, opened 2022-01-28
Introduce an intermediate lint-candidates target

webknjaz:
This change exposes an additional vetting-like linting environment
that is limited to rules that are actively being addressed by the
contributors. This rules sub-list is to be kept small so that it
could be promoted put to voting and made enforced easily. Once the
rules are in a good shape, a new focused list of related rules can
take their place and become linting candidates.


cidrblock (approved):
looks good let's try it out!

PR #832, opened 2022-01-29
Avoid using bash when calling pytest

ssbarnea:
This change simplified the way we call pytest from tox, while still allowing user to customize the way pytest runs locally via [PYTEST_ADDOPTS](https://docs.pytest.org/en/latest/example/simple.html#how-to-change-command-line-options-defaults), which is more flexible.

In fact this change gives more control to the user as it can define any other options while removing the makeshift hack for controlling the multiprocessing only.

This change needs @cidrblock review in particular as it removing th XDIST_CPU which is used by him and will require him to evaluate the benefits of the new approach.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,11 @@ commands =
   /bin/bash -c 'podman pull -q {[base]default_ee} || docker pull -q {[base]default_ee}'
   /bin/bash -c 'podman pull -q {[base]small_test_ee} || docker pull -q {[base]small_test_ee}'
   # most coverage options are kept in pyproject.toml
-  /bin/bash -c 'py.test -vvvv -n `python -c "import multiprocessing; import os; print(int(os.environ.get(\"XDIST_CPU\") or multiprocessing.cpu_count()))"` --dist=loadfile --maxfail=15 --durations=100 --cov --cov-report=xml:{toxworkdir}/coverage-{envname}.xml --showlocals {posargs}'
+  # if you want to limit multiprocessing locally, define PYTEST_ADDOPTS="-n0 --dist no"
+  pytest {posargs:-vvvv \
+    -n auto --dist=loadfile --maxfail=15 \
+    --durations=100 --cov --cov-report=xml:{toxworkdir}/coverage-{envname}.xml --showlocals \
+    {env:PYTEST_ADDOPTS}}
 setenv =
   TERM = xterm-256color
 passenv =


webknjaz (commented):
Ideally, many of these options should migrate to `pytest.ini` and `tox.ini` would keep the ones that the contributors would want to redefine via something like `tox -- -n0 --dist no`. I don't think that the comment should advertise the env var as a primary override method for interactive invocations, only for the full test suite runs.

ssbarnea:
> Ideally, many of these options should migrate to `pytest.ini` and `tox.ini` would keep the ones that the contributors would want to redefine via something like `tox -- -n0 --dist no`. I don't think that the comment should advertise the env var as a primary override method for interactive invocations, only for the full test suite runs.

I think that you are missing a common use case pattern, one that is used at least by me and Brad. Environment variables can be used to tune the way testing tool work without touching the tracked files and also allowing use to configure them  per user account, or per project or group of projects ([direnv](https://direnv.net/)). This allows us to disable xdist, or enable `--pdb -s` for default pytest runs without having to add extra command line parameters. Adding command line args is painful as you almost always run without them by mistake just to cancel testing.

In most cases I run pytest without any virtualenv, no tox, directly on each of our projects and it just works fine (i usually keep most of them installed as editable). On more complex projects I rely on longer `tox -e py -- ...`, but that is uncommon. Usually if I need to do this it is a sign that the test config is less than ideal. 

BTW, running `tox -- ...` is almost for sure not going to work well because it will pass posargs too all default environments and lint(pre-commit) and units (pytest) never have compatible arguments.

The commend I added there was mainly for Brad, but I find it useful for many users. Moving pytest options was not the subject of this PR, lets avoid scope creeping. I have good reasons for not wanting to touch that subject as I know that if you put addops in pytest config, we force user to have all the plugins installed or he would not be able to run just `pytest`. Sometimes I remove pytest-cov and I hate when I encounter a project that fail to run because there is a `--cov` inside pytest config addopts.

webknjaz:
I still think that the main use-case to advertise in the comment is posargs, env vars being secondary. I don't have a problem using Ctrl+R where necessary and I never had to set env vars. Plus things like direnv are used for all invocations, not for cases where you only want to run one thing. It would be pointless to have `-n0` in env vars on dev machines anyway, plus it has lower precedence. Besides, advertising the use of env vars is misleading, especially for people who don't understand the precedence.

cidrblock (approved):
looks cleaner.... ty

I may need some more specific instructions on how set limit the CPU count in the future if it becomes necessary, but that is a rare thing.

tox.ini
```diff
@@ -27,7 +27,11 @@ commands =
   /bin/bash -c 'grep -iFInrq "UPDATE_FIXTURES = True" ./tests && exit 1 || exit 0'
   /bin/bash -c 'podman pull -q {[base]default_ee} || docker pull -q {[base]default_ee}'
   /bin/bash -c 'podman pull -q {[base]small_test_ee} || docker pull -q {[base]small_test_ee}'
-  /bin/bash -c 'py.test -vvvv -n `python -c "import multiprocessing; import os; print(int(os.environ.get(\"XDIST_CPU\") or multiprocessing.cpu_count()))"` --dist=loadfile --maxfail=15 --durations=100 --cov ansible_navigator --cov share --cov-report term-missing --cov-branch --cov-report=xml:{toxworkdir}/coverage-{envname}.xml --showlocals {posargs}'
+  # if you want to limit multiprocessing locally, define PYTEST_ADDOPTS="-n0 --dist no"
```
webknjaz, line 30: This suggestion is lying — it will not work unless posargs are _also_ overridden according to the how the pytest args precedence works.

webknjaz (changes requested):
The comment must stop being misleading for this PR to be acceptable.

tox.ini
cidrblock: @webknjaz intended to suggest......

This suggestion might be incorrect — it will not work unless posargs are also overridden according to the how the pytest args precedence works

ssbarnea: I updated the change to ensure that `PYTEST_ADDOPTS` does not get shadowed by our default arguments from tox.ini.

As we do not use any hardcoded pytest args inside tox.ini, both of these calling methods have the same effect:
```
PYTEST_ADDOPTS="-n0 --dist no" tox -e py
tox -e py -- -n0 --dist no
```

tox.ini
```diff
@@ -28,7 +28,11 @@ commands =
   /bin/bash -c 'podman pull -q {[base]default_ee} || docker pull -q {[base]default_ee}'
   /bin/bash -c 'podman pull -q {[base]small_test_ee} || docker pull -q {[base]small_test_ee}'
   # most coverage options are kept in pyproject.toml
-  /bin/bash -c 'py.test -vvvv -n `python -c "import multiprocessing; import os; print(int(os.environ.get(\"XDIST_CPU\") or multiprocessing.cpu_count()))"` --dist=loadfile --maxfail=15 --durations=100 --cov --cov-report=xml:{toxworkdir}/coverage-{envname}.xml --showlocals {posargs}'
+  # if you want to limit multiprocessing locally, define PYTEST_ADDOPTS="-n0 --dist no"
+  pytest {posargs:-vvvv \
+    -n auto --dist=loadfile --maxfail=15 \
+    --durations=100 --cov --cov-report=xml:{toxworkdir}/coverage-{envname}.xml --showlocals \
+    {env:PYTEST_ADDOPTS}}
```
webknjaz, line 35: This changes the pytest behavior to be different from what a regular pytest user would expect.
The official documentation shows how this environment variable is supposed to be used, which is exactly opposite to what this does — https://docs.pytest.org/en/6.2.x/reference.html#envvar-PYTEST_ADDOPTS.

webknjaz (changes requested):
Redefining the upstream feature to do something opposite to what's documented and well-known shouldn't happen here.

tox.ini
ssbarnea: Last two comments lack one essential part: suggestion an acceptable alternative for addressing the goal of the this pull request. How can we achieve the stated goals?

IMHO, the re-use of that option is not really a big issue. Obviously that we could use another environment variable name to avoid that, bug do we really need another one?

webknjaz: Okay, use another variable but don't change the semantics of the well-known one.

cidrblock:
converting to draft b/c is seems a little inactive now

cidrblock:
This has gotten a little stale, we can reopen/revisit in the future as needed

PR #833, opened 2022-01-29
Move most coverage options to pyproject.toml

ssbarnea:
This simplifies pytest calling and also makes most coverage settings tox agnostic, so other tools can load them.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,14 @@
 [tool.black]
 line-length = 100
 
+[tool.coverage.run]
+    branch = true
+    source_pkgs = ["ansible_navigator"]
+    source = ["share"]
+
+[tool.coverage.report]
+    show_missing = true
+
 [tool.isort]
 force_single_line = true # Force from .. import to be 1 per line, minimizing changes at time of implementation
 known_first_party = "ansible_navigator, key_value_store" # No effect at implementation, here anticipating future benefit
--- a/tox.ini
+++ b/tox.ini
@@ -27,7 +27,8 @@ commands =
   /bin/bash -c 'grep -iFInrq "UPDATE_FIXTURES = True" ./tests && exit 1 || exit 0'
   /bin/bash -c 'podman pull -q {[base]default_ee} || docker pull -q {[base]default_ee}'
   /bin/bash -c 'podman pull -q {[base]small_test_ee} || docker pull -q {[base]small_test_ee}'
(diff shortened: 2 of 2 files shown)


webknjaz (changes requested):
I don't think that piling up this many tool options is a single file is in any way maintainable. It's best to use tool-specific configs.

ssbarnea:
> I don't think that piling up this many tool options is a single file is in any way maintainable. It's best to use tool-specific configs.

That counts a personal preference, so it should not be a reason to -1, especially as proposed config is an net improvement over current situation.

The benefit of toml file is that it declutters the project root directory. If we would already had the standalone config, I would not have proposed moving it to toml, but w/o one, i found the move appealing. standalone configs are good for tools that have very verbose config, but for those that you configure once and you are quite unlikely to touch after, is better to avoid create new files in repo root. The list of already existing tools we have there is... self explanatory.

cidrblock (approved):
Cool, let's try it.

I do like the reduced complexity of the tox file.

webknjaz:
> The benefit of toml file is that it declutters the project root directory. 

This counts as a personal preference. Wanting to make it harder to discover tool settings is not a good enough reason for this PR to exist 🤷‍♂️

webknjaz:
> If we would already had the standalone config, I would not have proposed moving it to toml

Funny that it's exactly what happened in other devtools projects — you added duplicate configs to `pyproject.toml` without reviews despite dedicated config being present.

FWIW I've made a case in the past that synchronizing configs from skeleton repos is substantially easier with standalone files. This point still stands.

webknjaz:
> I do like the reduced complexity of the tox file.

The right place for moving most of those options is `pytest.ini`. Even pytest-cov settings should mostly go there. Having the underlying tool config influencing the pytest settings implicitly is arguably not the best idea maintenance-wise.

cidrblock:
I'd like to propose an idea here, since it's hard sometimes to look into the future to really know what one config file per tool vs. one for everything that can go it......... let's stay the course with the pyproject.toml file, and when the CI/tox/precommit/GHA stuff is all cleaned up one of use put together a PR that breaks it all out.

The benefit here is everyone on the team will be able to "see" the difference, one vs. many.

The reason I like this PR is because it moves config out of the tox.ini that doesn't need to be there, and moves it into a config file the tool uses, independent of the name of that file.

PR #836, opened 2022-01-29
Documentation improvements in `docs/_ext/single_sourced_data.py` (D/DAR)

cidrblock:

- doc strings updated to reflect parameters in function signature
- descriptions updated to reflect functionality
- fix spelling mistakes as encountered

This additionally allows for the removal of D* and DAR* exceptions from the `flake8` configuration file for this file.

Please check for accuracy, as my understanding of the sphinx internals is limited.

### DIFF
--- a/docs/_ext/single_sourced_data.py
+++ b/docs/_ext/single_sourced_data.py
@@ -42,7 +42,7 @@ PARAM_TABLE_HEADER = [
     "  :header-rows: 1",
 ]
 RST_FIRST_ROW_ENTRY = "  * - {}"
-RST_ADDITONAL_ROW_ENTRY = "    - {}"
+RST_ADDITIONAL_ROW_ENTRY = "    - {}"
 RST_NL_CELL_FIRST = "    - | {}"
 RST_NL_IN_CELL = "      | {}"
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

docs/_ext/single_sourced_data.py
```diff
@@ -76,6 +76,11 @@ def _nodes_from_rst(
     """Turn an RST string into a list of nodes.
 
     These nodes can be used in the document.
+
+    :param state: The superclass for a docutils statemachine which contains a list or transitions
```
webknjaz, line 80: Technically, it's not a class but an instance that is initialized and managed by Sphinx. Back when I had to figure this out, it was not obvious, it's good that you're documenting this.

docs/_ext/single_sourced_data.py
```diff
@@ -76,6 +76,11 @@ def _nodes_from_rst(
...
+        and transition methods
+    :param rst_source: The source of the RST content from which nodes will be extracted
```
webknjaz, line 82: While it's called RST, really it'd work with other integrated sources, like MyST Markdown. But I didn't know that at the time of the initial implementation.

PR #837, opened 2022-01-30
Documentation improvements in several files (D/DAR) 

cidrblock:

doc strings updated to reflect parameters in function signature
descriptions updated to reflect functionality
fix spelling mistakes as encountered

This additionally allows for the removal of D* and DAR* exceptions from the flake8 configuration file for these files.


share/ansible_navigator/utils/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Utility scripts that will be mounted into the execution environment.
```
webknjaz, line 1: I don't think we need future possibilities in the title. Just tell people what it does not what something external does with it in some cases. At least, not in the title.
```suggestion
"""Utility scripts mounted into the execution environment.
```

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -91,7 +93,12 @@ def _catalog_plugins(self, collection: Dict) -> None:
 
     @staticmethod
     def _generate_chksum(file_path: str, relative_path: str) -> Dict:
-        """generate a std checksum for a file"""
+        """Generate a standard checksum for a file.
+
+        :param file_path: The path to the file to generate a checksum for
+        :param relative_path: The relative path within the collection directory structure
+        :returns: A diction or details about the file, including the checksum
```
webknjaz, line 100: "diction or details"? Feels like some autocorrect gone rogue. "Of"?

webknjaz (commented):
I did not look closely at everything but some of the patterns I've pointed out repeat across the patch.

share/ansible_navigator/utils/catalog_collections.py
cidrblock: lol

webknjaz (approved):

ssbarnea (approved):

PR #840, opened 2022-01-30
Pin pylint to 2.11.1

cidrblock:
This pins pylint to a version less than a version that requires astroid 2.9+.  

(pylint 2.12.2 was creating some fale `no-member` errors as seen below)

Although the issue was not seen in pre-commit, although it can be recreated like this:

```

```
which results in:
```


```

this PR was tested by checking out the PR in the container:

```

```


cidrblock:
didn't work

.pre-commit-config.yaml
```diff
@@ -271,7 +271,11 @@ repos:
     - tests/
 
 - repo: https://github.com/pycqa/pylint.git
-  rev: v2.12.2
```
ssbarnea, line 274: And the big challenge would be on how to prevent pre-commit to bump this. We will no longer be able to run `pre-commit autoupdate` without undoing this change.

I am wondering if that issue is not related to an outdated pip because if we ping astroid and pylint asks for something else pip should fail to install. AFAIK, that works fine with recent pip versions.

ssbarnea: I am inclined to believe that silencing no-member error until we get a fix would be easier.

PR #842, opened 2022-01-30
Note python version requirement for tox in contributors guide

cidrblock:
Make a note in the contributor guide about why the python version is set, and that it will need to be installed locally for tox to run


.github/CONTRIBUTING.md
```diff
@@ -58,6 +58,15 @@ Prerequisites:
    congratulations :)
    ```
 
+   ```{note}
```
webknjaz, line 61: Maybe use
```suggestion
   ```{tip}
```

cidrblock: let's see how it looks

cidrblock: I think it's better, TY

webknjaz (approved):

PR #845, opened 2022-01-31
Move `flake8-quotes` entry to bottom of file and add explanation

cidrblock:
This move the flake8-quotes entry to the bottom of the file as it confused the highlighting.

before:

image

after

image


.flake8
```diff
@@ -328,3 +326,10 @@ rst-roles =
 
 # wemake-python-styleguide
 show-source = true
+
+# flake8-quotes (pulled in by wemake style guide)
```
webknjaz, line 330: Maybe we could follow the same commenting pattern as with the pytest-style plugin above? A line with the plugin name then rules that specific options influence. We probably shouldn't send people off to the docs to try to figure out this connection.

cidrblock: I added the we-make thing because the flake8 extension is not otherwise configured and someone might try to remove it w/o knowing the connection to wemake

.flake8
```diff
@@ -328,3 +326,10 @@ rst-roles =
...
+# https://github.com/zheller/flake8-quotes
+# at bottom of configuration because some IDEs
```
webknjaz, line 332: Also, it's probably worth specifying that because of `inline-quotes`, the set of settings should remain at the bottom of the whole file. There's already a pattern of keeping any settings contributed by third-party plugins below flake8 "stdlib" options that I've adopted long ago for clarity.

.flake8
webknjaz: Note that several sections above are also coming from WPS. So we may want to do something about that in a more consistent manner.

.flake8
cidrblock: it was clear to me :)  I'll rephrase to be more specifc

webknjaz (approved):

.flake8
cidrblock: resolving after approved and implemented

PR #846, opened 2022-01-31
Move `f`-string check from per-file to global

cidrblock:
This relocates the f-string exemption from per file to the global ignore list.

- We allow the use of fstring, and prefer them
- This reduces the unexplained list of per-file exclusions

This particular message is from the unenforced, we-make-style guide which is being researched and vetted  

Related #717

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -39,9 +39,9 @@ extend-exclude =
 extend-ignore =
   E203,  # skipped because it annoys black
   WPS300,  # "Found local folder import" -- nothing bad about this
+  WPS305,  # "Found `f` string" -- f strings are preferred for readability
   # An opposite consistency expectation is currently enforced
   # by pylint via: useless-object-inheritance (R0205):
-  WPS305,  # "Found `f` string", allow the use of f strings
   WPS306,  # "Found class without a base class: *"
   WPS326,  # "Forbid implicit string concatenation."
   WPS421,  # "Forbids to call some built-in functions.", hasattr is good.
--- a/.flake8
+++ b/.flake8
@@ -41,6 +41,7 @@ extend-ignore =
   WPS300,  # "Found local folder import" -- nothing bad about this
   # An opposite consistency expectation is currently enforced
   # by pylint via: useless-object-inheritance (R0205):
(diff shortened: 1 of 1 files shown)


.flake8
```diff
@@ -41,6 +41,7 @@ extend-ignore =
   WPS300,  # "Found local folder import" -- nothing bad about this
   # An opposite consistency expectation is currently enforced
   # by pylint via: useless-object-inheritance (R0205):
+  WPS305,  # "Found `f` string", allow the use of f strings
```
webknjaz, line 44: This is the wrong line to add this. Just look at the comment above that refers to the line below (WPS306) by having a colon at the end of the comment.

Move this addition to line 42. Plus maybe add motivation — as in "we really want to use f-strings and it's legit". I'd recommend extending the comment by adding the explanation after `--` like it's done on line 41 to prevent future similar separation of comments from the referred content.

webknjaz: Could also refer to https://github.com/ansible/ansible-navigator/discussions/717#discussioncomment-1791048 in the comment.

cidrblock: thx, updated. Many reasons to use them. I find readability the biggest win

.flake8
```diff
@@ -39,6 +39,7 @@ extend-exclude =
 extend-ignore =
   E203,  # skipped because it annoys black
   WPS300,  # "Found local folder import" -- nothing bad about this
+  WPS305,  # "Found `f` string" -- f strings are preferred for readability
```
ssbarnea, line 42: I would mention that pylint already covers for f-strings.

ssbarnea (approved):
Why do we even need to bother about fstrings when pylint is already able to detect and report them? I think that

webknjaz (approved):

.flake8
cidrblock: resolving as approved after change was implemented

.flake8
cidrblock: I'm gonna keep the comment lean here.... I suspect the topic of overlap will come up again if and when someone brings WPS forward as an enforced standard for the repo

Issue #459, opened 2021-06-07
Include ansible-builder functionality


ffirg:
### Summary

Have ansible-builder functionality within navigator.

### Issue Type

Feature Idea

### Additional Information

I would like to be able to produce and test an EE using ansible-navigator builder as an option.

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

s-hertel:
+1, seems like this would be handy.

mnecas:
+1

ffirg:
sweet! nice work!

PR #848, opened 2022-02-01
Add builder subcommand to manage EE builds

ganeshrn:
Resolves https://github.com/ansible/ansible-navigator/issues/459

*  Add builder subcommand to manage execution-environment
   creation.
*  The builder subcommand invokes `ansible-builder` utility
   under the hood and runs only in `stdout` mode
*  Add tests to cover builder subcommand scenarios


docs/changelog-fragments.d/459.feature.md
```diff
@@ -0,0 +1,9 @@
+Added the ability to run builder commands to create execution
+environment using mode `stdout`.
```
webknjaz, line 2: (important) Let's try to use the full names of things in the changelog. This is because (1) there's a generic Ansible docs writing style guide emphasizing this and (2) it is useful for the consumers of the changelog to have enough context into what the entry is about that they wouldn't need to do extra googling/code reading unless they actually want to dig deep.
```suggestion
Added the ability to run ansible-builder commands to create
execution environment using mode `stdout`.
```

docs/changelog-fragments.d/459.feature.md
```diff
@@ -0,0 +1,9 @@
...
+
+Here is an example invocation:
+```bash
+ansible-navigator builder build --tag ee_image_name --workdir ~/manifest-file --container-runtime podman -v 3 --mode stdout
```
webknjaz, line 6: (question, optional) I wonder if it would be better from the readability/rendering perspective to provide multiline commands in code snippets. This would be useful for people with larger fonts (i.e. less letters per line), narrow browser windows, mobile screens.
```suggestion
ansible-navigator builder \
  build \
  --tag ee_image_name \
  --workdir ~/manifest-file \
  --container-runtime podman \
  -v 3 \
  --mode stdout
```

setup.cfg
```diff
@@ -35,6 +35,7 @@ package_dir =
   = src
 packages = find:
 install_requires =
+    ansible-builder >=1,<2
```
webknjaz, line 38: (educational, optional)  There's a possibility to use `~=` as an alias to what's used here like this. If we don't know if `ansible-builder` is going to be following strict SemVer, it _may be_ better to use it (per your discretion):
```suggestion
    ansible-builder ~= 1
```
Another possibility to improve this is to add spacing to version specifiers — it's a bit easier/nicer to consume when it's sparse:
```suggestion
    ansible-builder >=1, <2
```

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
+"""Run the builder subcommand."""
```
webknjaz, line 1: (educational, important) Modules are not actions. Just importing them does not create side effects of performing actions (well, at least it shouldn't). When you say that something does things, it usually happens in the context of invoking functions or methods — for those, do use verbs/imperative mood. Module docstrings, OTOH, should just describe what the module contains. This one contains a class definition. By importing the module the class becomes defined/available in runtime but this is not a request to execute things. Well, looking at the source, there's one action, in fact — self-registration within the actions subsystem. Although this may be a controversial practice by itself, it seems to be a default approach in this project. I guess it may be worth mentioning too.
```suggestion
"""Builder subcommand implementation.

Importing this module registers this subcommand in the external
global subcommand registry.
"""
```

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
...
+
+    KEGEX = "^b(?:uilder)?$"
+
+    def __init__(self, args: ApplicationConfiguration):
+        """Initialize the action.
+
+        :param args: The current application configuration.
+        """
+        super().__init__(args=args, logger_name=__name__, name="builder")
+        self._runner: Command
```
webknjaz, line 28: Is it intentional that this does not have a default value?
```suggestion
        self._runner: Command = None
```

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def run_stdout(self) -> Union[None, int]:
```
webknjaz, line 30: (educational, optional) Hint: it's the same as
```suggestion
    def run_stdout(self) -> Optional[int]:
```

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
...
+        """Run in mode stdout.
+
+        :returns: The return code or None
```
webknjaz, line 33: (important) What is the semantical meaning of `None`? This would be a good place to mention it.

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
...
+        kwargs = {
+            "execution_environment": False,
+            "host_cwd": os.path.abspath(os.path.expanduser(self._args.workdir)),
+            "navigator_mode": self._args.mode,
+            "pass_environment_variable": self._args.pass_environment_variable,
+            "set_environment_variable": envvars_to_set,
+            "timeout": self._args.ansible_runner_timeout,
+        }
+
+        ansible_builder_path = shutil.which("ansible-builder")
```
webknjaz, line 78: (optional) `"ansible-builder"` is used in multiple places so I'd probably suggest putting it in a class attribute.
Also, it seems like it's best to do the resolution in the initializer and expose the full path as an instance attribute. The `is None` check would need to remain in this method, though.

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,95 @@
...
+        if ansible_builder_path is None:
+            msg = "'ansible-builder' executable not found"
+            self._logger.error(msg)
+            raise RuntimeError(msg)
```
webknjaz, line 82: (educational, optional) This is what's called a "guard expression". They are commonly used as an early control flow interruption. Traditionally, it's best to put guard expressions at the beginning of the code block/scope/function/method.
Looking at it, I see that it does not depend on anything pre-calculated earlier.

An additional benefit would be a small optimization — exiting early would mean that all those unnecessary runtime operations that do not need to be performed anymore will be skipped.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -483,5 +501,13 @@ class Internals(SimpleNamespace):
             ),
             value=SettingsEntryValue(),
         ),
+        SettingsEntry(
+            name="workdir",
+            cli_parameters=CliParameters(short="--bwd"),
+            settings_file_path_override="builder.workdir",
+            short_description="Specify the path that contains ansible-builder manifest files",
+            subcommands=["builder"],
+            value=SettingsEntryValue(default=os.getcwd()),
```
webknjaz, line 510: (important) Should this be the root of the project where the navigator config is present (or `.git/`, for example)?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -429,6 +429,37 @@ def exec_shell(
         """
         return self._true_or_false(entry, config)
 
+    @_post_processor
+    def help_builder(
+        self,
+        entry: SettingsEntry,
+        config: ApplicationConfiguration,
+    ) -> PostProcessorReturn:
+        # pylint: disable=unused-argument
```
webknjaz, line 438: (accuracy, optional) Is it possible to put this on the same line as the unused argument?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -429,6 +429,37 @@ def exec_shell(
...
+        """Post process help_builder"""
+        messages, exit_messages = self._true_or_false(entry, config)
+        if all(
+            (
+                entry.value.current is True,
+                config.app == "builder",
+                config.mode == "interactive",
+            )
+        ):
```
webknjaz, line 447: (important)
1) This is not a good use for `all()`. There is no need to create an iterable from already pre-calculated values only for "all" to iterate over them. It's actually less performant and consumes more memory than an alternative with an early exist which is to use condition chaining with an `and`.
2) This is so big and nested that it's hard to digest when embedded in a huge if-clause. It's better to assign a meaningful semantically significant name to the whole complex condition and maybe to its sub-conditions too. I tried to do some guessing what the meaning is below (it's probably wrong but it's the best one can do when reading this patch within the provided context):
    ```suggestion
            params_need_checking = (
                    entry.value.current is True and
                    config.app == "builder" and
                    config.mode == "interactive"
            )
            if params_need_checking:
    ```
    ```suggestion
            entry_value_is_selected = entry.value.current is True
            requested_app_is_builder = config.app == "builder"
            runtime_is_interactive = config.mode == "interactive"
            an_interactive_builder_cmd_is_selected = (
                entry_value_is_selected and
                requested_app_is_builder and
                runtime_is_interactive
            )
            if an_interactive_builder_cmd_is_selected:
    ```

Tell me if I guessed any of these correctly :)

webknjaz:
FYI I've fixed several typos in the PR description (in case you'd want to incorporate that in the Git commit message).

docs/changelog-fragments.d/459.feature.md
ganeshrn: Fixed

docs/changelog-fragments.d/459.feature.md
ganeshrn: I find multi-line comment is difficult to read especially on a monitor with a medium font. Hence I will keep the format as it is.

setup.cfg
ganeshrn: FIxed

src/ansible_navigator/actions/builder.py
ganeshrn: Fixed

src/ansible_navigator/actions/builder.py
ganeshrn: Good catch. For builder `runner` is not required to be a class attribute. Hence it is removed now.

src/ansible_navigator/actions/builder.py
ganeshrn: Fixed

src/ansible_navigator/actions/builder.py
ganeshrn: Fixed

src/ansible_navigator/actions/builder.py
ganeshrn: `ansible-builder` executable path is used only in `_run_runner` method and the path information is required only while invoking `ansible-runner` API hence it is not a class attribute.

src/ansible_navigator/actions/builder.py
ganeshrn: Good catch! Fixed now.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
ganeshrn: If the navigator config file is present in CWD it takes precedence over other shared paths. Also after developing collection if the user wants to build EE chances are the builder manifest files will most likely be present in the CWD itself.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
ganeshrn: It is not required and removed now. Fixed!

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
ganeshrn: Yes you guessed it correctly :-). 

Nice suggestion. Have fixed it now.

src/ansible_navigator/actions/builder.py
```diff
@@ -0,0 +1,102 @@
...
+from ..configuration_subsystem.definitions import Constants
+from ..runner import Command
+from . import _actions as actions
+
+
+@actions.register
+class Action(App):
+    """Run the builder subcommand."""
+
+    KEGEX = "^b(?:uilder)?$"
```
cidrblock, line 24: since this is stdout only, we can probably have this be just "builder", which will leave `:b` for something in the TUI... (unless of course we think we'll add it someday)

cidrblock (approved):
This so great........... ty

src/ansible_navigator/actions/builder.py
cidrblock: Come to think of it, it's not worth changing until we need b

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: I like this default.

src/ansible_navigator/actions/builder.py
cidrblock: Changing it later if needed shouldn't be a major issue. I think Ganesh's comment about scope here is accurate. If we need it commonly between multiple funcs later it should be a quick change.

tests/unit/configuration_subsystem/data.py
```diff
@@ -244,6 +245,7 @@ def cli_data():
     ("plugin_type", "become", "become"),
     ("pull_policy", "never", "never"),
     ("set_environment_variable", "T1=A,T2=B,T3=C", {"T1": "A", "T2": "B", "T3": "C"}),
+    ("workdir", "/tmp/", "/tmp/"),
```
ssbarnea, line 248: @cidrblock That default looks pretty bad to me, as I know for sure that MacOS and few other systems do not allow users to write to /tmp for security reasons.

PR #852, opened 2022-02-01
[exec] *actually* set env vars from config

relrod:
Change:
- A misindentation caused 'ansible-navigator exec' to not actually get
  env vars from config.

Test Plan:
- new integration


cidrblock (approved):
DOH! nice catch

PR #855, opened 2022-02-01
Eliminate a few flake8 violations with a vetted `autopep8`

webknjaz:
This patch takes care of small flake8/pydocstyle violations that can be autofixed like empty lines before docstrings or assigned lambdas (E731).

Extension of the pre-commit config is there to help out early adopters of linting candidates and does not affect anybody else in any other envs.

### DIFF
--- a/.pep8
+++ b/.pep8
@@ -1,0 +1,10 @@
+[pep8]
+aggressive = 0
+in-place = true
+# Fixing E203 is disabled because it annoys black
+ignore =
+  E203,
+jobs = 0
+# Accessibility/large fonts and PEP8 unfriendly:
+max-line-length = 100
+recursive = true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,13 @@ repos:
     - --py36-plus
     stages: ["manual"]
 
+- repo: https://github.com/pre-commit/mirrors-autopep8.git
+  rev: v1.6.0
+  hooks:
+  - id: autopep8
+    stages:
+    - manual
+
 - repo: https://github.com/PyCQA/isort
   rev: 5.10.1
   hooks:
--- a/src/ansible_navigator/actions/collections.py
+++ b/src/ansible_navigator/actions/collections.py
@@ -29,7 +29,6 @@ from . import run_action
 
 def color_menu(colno: int, colname: str, entry: Dict[str, Any]) -> Tuple[int, int]:
     # pylint: disable=unused-argument
-
     """color the menu"""
     if entry.get("__shadowed") is True:
         return 8, 0
--- a/src/ansible_navigator/actions/config.py
+++ b/src/ansible_navigator/actions/config.py
@@ -31,7 +31,6 @@ from . import run_action
 
 def color_menu(colno: int, colname: str, entry: Dict[str, Any]) -> Tuple[int, int]:
     # pylint: disable=unused-argument
-
     """color the menu"""
     if entry["__default"] is False:
         return 3, 0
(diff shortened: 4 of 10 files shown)


cidrblock:
Because this introduces a new linter and fixes a only small number of otherwise simple issues which are not very common, I'm gonna take a day to consider this one.

ssbarnea (changes requested):
After doing bit of googling about the subject and finding [E731](https://github.com/ray-project/ray/issues/21315) and also seeing the added `max-line-length = 100` line,  I am starting to become bit paranoid... i kinda have the impression that this aim on creating conflicts with our use of black!

After we already had to undo https://github.com/ansible/ansible-navigator/pull/855 I find hard to no to think "oh, not another tool/plugin/dependency!".

This change also seem to have also modify get_color and color implementation when it should have not done this, combining code change with linter config another reason to -1 it.

webknjaz:
> After doing bit of googling about the subject and finding [E731](https://github.com/ray-project/ray/issues/21315) 
> [..] `max-line-length = 100` line, I am starting to become bit paranoid... i kinda have the impression that this aim on creating conflicts with our use of black!

It's just a copied value from another config. Totally unrelated, I can't even imagine what this connection suggestion was based on but it's as far from true as it can get. Besides, this rule is *already enforced* in flake8. So I don't see how this point can even relate to this PR in any way.

> After we already had to undo #855 I find hard to no to think "oh, not another tool/plugin/dependency!".

It just helps address a number of autofixable flake8 rules so you don't have to do it manually :man_shrugging:

> This change also seem to have also modify get_color and color implementation when it should have not done this

This statement is incorrect because the only changes that have been applied were to make sure that the linters are still green. You can't turn a lambda into a function definition and not have the docstring-related linters notice it, for example.

ssbarnea (changes requested):
I am requesting `quorum-required` on this change because we did not enable black auto-formatting on this repository and until this happens, there is a chance that this would cause conflicts.

Once we do have black enabled I would even agree to run autopep not as manual but as automatic, so we can effectively see that it does not conflict with black and that it does fix something that black does not.

Basically, now is not the right time but not against the idea in general. Still I am really worried about potential conflicts. We need to check how many projects auto both autopep8 and black in pre-commit. I do not fancy at all an endless loop where one of them undoes changes made by the other.

Added -1 because I don't want to see it auto-merged with only one review.

webknjaz:
> I am requesting `quorum-required` on this change because we did not enable black

Unfortunately, it is already enforced, not in vetting. So this statement is incorrect. It is literally impossible to contribute without having to suffer black.

> auto-formatting on this repository and until this happens, there is a chance that this would cause conflicts.

No, I checked that autopep8 doesn't undo black's mess. It only fixes PEP 8 violations.

> Once we do have black enabled I would even agree to run autopep not as manual but as automatic, so we can effectively see that it does not conflict with black and that it does fix something that black does not.

It is already visible, the change is result of both.

> Basically, now is not the right time but not against the idea in general. Still I am really worried about potential conflicts. We need to check how many projects auto both autopep8 and black in pre-commit. I do not fancy at all an endless loop where one of them undoes changes made by the other.
> 

As I said above, I checked it already.

cidrblock:
I'm -1 on this one, just a not now really.....

I find our CI/lint/formatting set-up pretty complex right now, and until it's cleaned up a little and the entire team can grok why and how it is set up I'd rather not add to it.

Additionally, work is happening to clean up doc string, when that is done, (and not until it is done), I think we should ask the team if they want pydocstyle and darglint enforced.  I think that conversation would be less crisp if it included a conversation about autopep8 at the same time. I think the decision about docstrings should be about docstring standards and consistency and not linters. IMO linters are only there to support the decision about the quality of the docstrings.

I think we would probably be in a position to raise this separately, after we talk and decide about docstring quality.

I would like the fixes to go in though, especially if the parameters and return values were added to the docstrings of the converted lambdas (optional but encouraged :) )

cidrblock:
converting to draft for now

cidrblock:
This has gotten a little stale, we can reopen in the future if/as needed.

PR #858, opened 2022-02-02
Stop ignoring `D412` rule in the flake8 and precommit configurations

cidrblock:

Clean-up was done in #651

This updates the .flake8 file to reflect those changes.

Additionally remove from pre-commit config, since this is the last occurance

(missed in #856 )

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -77,7 +77,7 @@ per-file-ignores =
   share/ansible_navigator/utils/key_value_store.py: WPS110, WPS214, WPS526, WPS600, WPS603
   src/ansible_navigator/__init__.py: WPS300, WPS412, WPS436
   src/ansible_navigator/__main__.py: RST304, WPS300
-  src/ansible_navigator/_version.py: D412, WPS410
+  src/ansible_navigator/_version.py: WPS410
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

ssbarnea (approved):

ssbarnea (dismissed):
Still present in https://github.com/ansible/ansible-navigator/blob/88e0d64981aa1899f22cf4e96727d34f9f0267c1/.pre-commit-config.yaml#L146

webknjaz:
> Still present in

That's a different config and a different hook (which should probably be removed anyway because it is exactly the same as `flake8-docstrings`).
I don't think that this review is relevant to this PR as it specifically addresses flake8 violations and not other things.

cidrblock:
> Still present in
> 
> https://github.com/ansible/ansible-navigator/blob/88e0d64981aa1899f22cf4e96727d34f9f0267c1/.pre-commit-config.yaml#L146

It's not clear to me why that is in pre-commit, and is generally not something I pay attention to.

If it can be removed, let's do that.

webknjaz:
> If it can be removed (the entire entry), let's do that.

Yes, I think it's just a result of synching stuff across projects.

PR #652, opened 2021-12-05
[lint, type] Linting, docstring and type improvements in a*.py files

cidrblock:
depends-on: https://github.com/ansible/ansible-navigator/pull/671
- does what it says
- Additionally remove unused function (it got replaced by _argparse_error_handler in configurator long ago)

PR #859, opened 2022-02-02
Documentation improvements in `action_runner.py` file (D/DAR)

cidrblock:

A majority of the work was done in #652.

Some additional information was added here about how navigator works since this file is so critical in the whole flow and the first step in bringing up the TUI.

doc strings updated to reflect parameters in function signature
descriptions updated to reflect functionality
fix spelling mistakes as encountered

This additionally allows for the removal of D* and DAR* exceptions from the flake8 configuration file for the file(s) modified in the PR.

### DIFF
--- a/src/ansible_navigator/action_runner.py
+++ b/src/ansible_navigator/action_runner.py
@@ -1,4 +1,16 @@
-"""Jump to one action."""
+"""A mechanism to start the main application loop for the text user interface.
+
+Invoked after settings are parsed, a loop is entered and only
+exited when the user requests to quit.
+
+After initializing the user interface, an interaction is
+created from the fully initialized settings, and the requested
+action is run.
+
+From that point forward flow control of the application is handled
+by each action in the action stack, control returned here when
+`:quit` is requested by the user.
+"""
 import os
 
 from typing import TYPE_CHECKING
--- a/src/ansible_navigator/action_runner.py
+++ b/src/ansible_navigator/action_runner.py
@@ -1,10 +1,11 @@
 """A mechanism to start the main application loop for the text user interface.
 
-Invoked after settings are parsed, a loop is entered and only
+A instance of :class:`~ansible_navigator.action_runner.ActionRunner`
+is invoked once, after settings are parsed. A loop is entered and only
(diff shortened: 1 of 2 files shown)


src/ansible_navigator/action_runner.py
```diff
@@ -1,4 +1,16 @@
...
+Invoked after settings are parsed, a loop is entered and only
+exited when the user requests to quit.
+
+After initializing the user interface, an interaction is
+created from the fully initialized settings, and the requested
+action is run.
+
+From that point forward flow control of the application is handled
+by each action in the action stack, control returned here when
+`:quit` is requested by the user.
```
webknjaz, line 12:
```suggestion
``:quit`` is requested by the user.
```

src/ansible_navigator/action_runner.py
```diff
@@ -1,4 +1,16 @@
-"""Jump to one action."""
+"""A mechanism to start the main application loop for the text user interface.
+
+Invoked after settings are parsed, a loop is entered and only
```
webknjaz, line 3: Modules are not invoked but imported, class instances are initialized and their methods are invoked. So this could probably go into some of those places it tries to refer to?

webknjaz: Or at least explicitly mention what it is that is invoked if this is meant to be at the beginning of the document.

cidrblock: this is why knowing something about the code works against me.... One file, one class, named the same, it is only implied that that comment refers the sole class in the file. updating, thx

src/ansible_navigator/action_runner.py
```diff
@@ -1,4 +1,17 @@
-"""Jump to one action."""
+"""A mechanism to start the main application loop for the text user interface.
+
+A instance of :class:`~ansible_navigator.action_runner.ActionRunner`
```
webknjaz, line 3:
```suggestion
An instance of :class:`~ansible_navigator.action_runner.ActionRunner`
```

webknjaz (approved):

PR #860, opened 2022-02-02
Documentation improvements in `actions` package (D/DAR)

cidrblock:

- doc strings updated to reflect parameters in function signature
- descriptions updated to reflect functionality
- fix spelling mistakes as encountered

This additionally allows for the removal of D* and DAR* exceptions from the flake8 configuration file for the file(s) modified in the PR.

apologies in advance any errors here, while I understand how this is stitched together, I struggled to describe it in a manner that would make sense to the reader.


src/ansible_navigator/actions/__init__.py
```diff
@@ -1,4 +1,17 @@
-"""actions for Explorer"""
+"""A package containing all available actions.
+
+The actions package is a plugin like implementation for all available application actions.
```
webknjaz, line 3:
```suggestion
The actions package is a plugin-like implementation for all available application actions.
```

src/ansible_navigator/actions/__init__.py
```diff
@@ -1,4 +1,17 @@
...
+
+Actions can be added without registration and are discovered and loaded
+when their ``KEGEX`` is matched against user input.
```
webknjaz, line 6: (minor, maybe for follow-ups)
I wonder if this `KEGEX` thing should be marked with a `:term:` role and maybe belongs to a Sphinx glossary.

src/ansible_navigator/actions/__init__.py
```diff
@@ -1,4 +1,17 @@
...
+
+This allows for each action to be loaded multiple times, creating a unique
+and isolated instance of the action.
+
+Currently, the actions package is the only package supported for actions and
+is identified in the
+:class:`~ansible_navigator.configuration_subsystem.navigator_configuration.Internals`
```
webknjaz, line 13:
```suggestion
:class:`~ansible_navigator.configuration_subsystem.navigator_configuration.Internals`.
```

src/ansible_navigator/actions/__init__.py
```diff
@@ -1,4 +1,17 @@
...
+
+The actions package is a plugin like implementation for all available application actions.
+
+Actions can be added without registration and are discovered and loaded
+when their ``KEGEX`` is matched against user input.
+
+This allows for each action to be loaded multiple times, creating a unique
+and isolated instance of the action.
+
+Currently, the actions package is the only package supported for actions and
```
webknjaz, line 11: Should we highlight package names?
```suggestion
Currently, the ``actions`` package is the only package supported for actions and
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -1,4 +1,4 @@
-"""Some action stuff"""
+"""Helper functions for the actions package."""
```
webknjaz, line 1:
```suggestion
"""Helper functions for the ``actions`` package."""
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -36,12 +36,19 @@
...
+    :param action: The action to import
+    """
     importlib.import_module(f"{package}.{action}")
 
 
 def _import_all(package: str) -> None:
-    """Import all actions in a package"""
+    """Import all actions in a package.
+
+    :param package: The name of the package.
```
webknjaz, line 50: Did we reach an agreement on sentences with trailing periods vs phrases in params? Looks like even within this PR it's inconsistent.

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
+    :param action: The name of the action
+    :returns: The action's registered class
+    """
     _import(package, action)
     return _ACTIONS[package][action].cls
 
 
 def get_factory(package: str) -> Callable:
-    """Create a get() function for one package"""
+    """Create a ``get`` function for one package.
```
webknjaz, line 85: Maybe this would be more accurate?
```suggestion
    """Limit a ``get()`` function to a given package.
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
+
+    :param package: The name of the package
+    :returns: The action's registered class
+    """
     return functools.partial(get, package)
 
 
 def kegex(package: str, action: str) -> Tuple:
-    """Return a tuple of name, kegex for a action"""
+    """Return a tuple of name, class, kegex for a action.
```
webknjaz, line 94:
```suggestion
    """Return a tuple of name, class, ``kegex`` for an action.
```
Maybe this'd be a bit better? Since kegex is not a word.

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
     _import_all(package)
     return sorted(_ACTIONS[package])
 
 
 def names_factory(package: str) -> Callable:
-    """Create a names() function for one package"""
+    """Create a ``names`` function for one package.
+
+    :param package: The name of the package
+    :returns: a ``names`` method for the package
```
webknjaz, line 137: I suppose this could link either https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable or a glossary term explanation like it's done here https://docs.aiohttp.org/en/stable/glossary.html#term-callable.
```suggestion
    :returns: a ``names`` :class:`callable <collections.abc.Callable>` for the package
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
 
 
 def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> Any:
-    """Call the given action's run"""
+    r"""Call the given actions's run method.
+
+    :param package: The name of the package
+    :param action: The name of the action
+    :param \*args: The arguments passeed to the action's run method
+    :param \**_kwargs: The keyword arguments passed to the action's run method
```
webknjaz, line 148: I don't think it's customary to use asterisks like that. Have you seen it anywhere else?
```suggestion
    :param _kwargs: The keyword arguments passed to the action's run method
```

webknjaz: I did some googling and while Sphinx's docs don't describe this separately at https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists, its own source code has examples like https://github.com/sphinx-doc/sphinx/blob/f38bd8e9529d50e5cceffe3ca55be4b758529ff7/sphinx/application.py#L558 and it's rendered like this: https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_node (without asterisks in the parameter description area but the signature itself has asterisks).

webknjaz: Same with `*args`: https://github.com/sphinx-doc/sphinx/blob/f38bd8e9529d50e5cceffe3ca55be4b758529ff7/sphinx/application.py#L440.

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
+    :returns: The outcome of running the actions' run method
+    """  # noqa: DAR101, DAR102 # *args, **kwargs escaped for sphinx
```
webknjaz, line 150: The above means that this "noqa" is unnecessary.

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
+
+    :param package: The name of the package
+    :returns: a ``names`` method for the package
+    """
     return functools.partial(names, package)
 
 
 def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> Any:
-    """Call the given action's run"""
+    r"""Call the given actions's run method.
```
webknjaz, line 143: With the comments below, this would not need to remain a raw-string.

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
+    :returns: The outcome of running the actions' run_stdout method
+    """  # noqa: DAR101, DAR102 # *args, **kwargs escaped for sphinx
     action_cls = get(package, action)
     args = args[0]
     return action_cls(args).run_stdout()
 
 
 def run_stdout_factory(package: str) -> Callable:
-    """Create a run_stdout() function for one package"""
+    """Create a ``run_stdout`` function for one package.
```
webknjaz, line 185: I'd like for such references to keep `()` because this is what Sphinx appends to normal callable-like roles when they are rendered:
```suggestion
    """Create a ``run_stdout()`` function for one package.
```

src/ansible_navigator/actions/__init__.py
cidrblock: for the sake of history, KEGEX was a mash-up of key and regex from the early days of ansible-navigator when it was only single-key driven.....  It probably should be renamed REGEX, which more accurately describes it, or maybe ACTION_REGEX.....  it would not be a trivial change, but a good experience for someone not familiar with how it is used

src/ansible_navigator/actions/__init__.py
cidrblock: why not!

src/ansible_navigator/actions/_actions.py
cidrblock: nice

src/ansible_navigator/actions/_actions.py
cidrblock: If I recall it still will be............................................ but we can try

src/ansible_navigator/actions/_actions.py
cidrblock: I _thought?_ this caused a sphinx error, we'll know soon enough

src/ansible_navigator/actions/_actions.py
cidrblock: I've been trying to leave it off, thx

src/ansible_navigator/actions/_actions.py
cidrblock: all cleaned up, not sure how I got in that rat hole.

src/ansible_navigator/actions/_actions.py
cidrblock: yep! nice.

src/ansible_navigator/actions/_actions.py
cidrblock: gone and gone

src/ansible_navigator/actions/_actions.py
webknjaz: No, Sphinx doesn't care about things inside _double_ backticks — for it, it's just something that is arbitrary inline code thingy, there's no semantics inside.

src/ansible_navigator/actions/_actions.py
webknjaz: You probably googled and found some of the StackOverflow answers + docutils escaping guidelines. I've seen those too but then, I tried to look into what sphinx does and found that it doesn't bother to add asterisks — they are not a part of var names anyway.

webknjaz: Besides, it's the same as with type annotations — they annotate what gets into those variables in runtime.

src/ansible_navigator/actions/_actions.py
webknjaz: I guess this one may be hard to make accurate.

src/ansible_navigator/actions/__init__.py
webknjaz: :+1:

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
+    :param package: The name of the package
+    :returns: A generator for all ``kegexs``
+    """
     _import_all(package)
     return (kegex(package, name) for name in names(package))
 
 
 def kegexes_factory(package: str) -> Callable:
-    """Create a kegexs() function for all packages"""
+    """Create a ``kegexs()`` function for all packages.
```
webknjaz, line 115:
```suggestion
    """Create a ``kegexes()`` function for all packages.
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -52,54 +59,95 @@ def _import_all(package: str) -> None:
...
 
 def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> Any:
-    """Call the given action's run"""
+    """Call the given actions's ``run()`` method.
+
+    :param package: The name of the package
+    :param action: The name of the action
+    :param args: The arguments passeed to the action's run method
+    :param _kwargs: The keyword arguments passed to the action's run method
+    :returns: The outcome of running the actions' run method
```
webknjaz, line 149:
```suggestion
    :returns: The outcome of running the action's run method
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
 
 def run_stdout(package: str, action: str, *args: Any, **_kwargs: Any) -> Any:
-    """Call the given action's run_stdout"""
+    """Call the given action's ``run_stdout()`` method.
+
+    :param package: The name of the package
+    :param action: The name of the action
+    :param args: The arguments passeed to the action's run_stdout method
+    :param _kwargs: The keyword arguments passed to the action's run_stdout method
+    :returns: The outcome of running the actions' run_stdout method
```
webknjaz, line 177:
```suggestion
    :returns: The outcome of running the action's run_stdout method
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
+    """  # noqa: D402 # Refers to the actions's run_stdout, not this function
```
webknjaz, line 178:
```suggestion
    :returns: The outcome of running the action's ``run_stdout()`` method
    """
```

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
     args = args[0]
     return action_cls(args).run_stdout()
 
 
 def run_stdout_factory(package: str) -> Callable:
-    """Create a run_stdout() function for one package"""
+    """Create a ``run_stdout()`` function for one package.
+
+    :param package: The name of the package
+    :returns: A partial ``run_stdout`` method for the package
```
webknjaz, line 188:
```suggestion
    :returns: A partial ``run_stdout()`` method for the package
```

src/ansible_navigator/actions/_actions.py
cidrblock: appears the noqa is still needed
```
src/ansible_navigator/actions/_actions.py:171:1: D402 First line should not be the function's "signature"
    """Call the given action's ``run_stdout()`` method.

    :param package: The name of the package
    :param action: The name of the action
    :param args: The arguments passeed to the action's run_stdout method
    :param _kwargs: The keyword arguments passed to the action's run_stdout method
    :returns: The outcome of running the action's run_stdout method
    """
^
1     D402 First line should not be the function's "signature"
1
```

webknjaz: Oh, that's why! Then, maybe dropping `()` is fine for the first line.

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
 
 def run_stdout(package: str, action: str, *args: Any, **_kwargs: Any) -> Any:
-    """Call the given action's run_stdout"""
+    """Call the given action's ``run_stdout()`` method.
+
+    :param package: The name of the package
+    :param action: The name of the action
+    :param args: The arguments passeed to the action's run_stdout method
+    :param _kwargs: The keyword arguments passed to the action's run_stdout method
+    :returns: The outcome of running the action's run_stdout method
```
webknjaz, line 177:
```suggestion
    :returns: The outcome of running the action's ``run_stdout()`` method
```

src/ansible_navigator/actions/_actions.py
cidrblock: I'd rather have a noqa than one being different...  I'm afraid someone might try to readd it because it was off and no noqa was there to explain

src/ansible_navigator/actions/_actions.py
```diff
@@ -111,17 +159,32 @@ def run_interactive(package: str, action: str, *args: Any, **_kwargs: Any) -> An
...
+    """  # noqa: D402 # Refers to the actions's run_stdout, not this function
```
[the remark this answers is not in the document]
cidrblock: Clarfied

src/ansible_navigator/actions/_actions.py
webknjaz: Okay, I won't insist. Just don't like having too many noqas. Also, WPS has a rule with a per-module limit of noqas to prevent abuse.

webknjaz (approved):

PR #861, opened 2022-02-02
Reduce `pre-commit` verbosity in `tox.ini`

cidrblock:
- Reduce the verbosity of the pre-commit output, when needed it can be enabled locally with `tox -e lint -- --all-files -v`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,6 @@ description = Run pytest under {basepython} ({envpython})
 # one of it's layer
 allowlist_externals =
   cat
-  echo
   rm
   grep
 deps =
@@ -69,9 +68,6 @@ allowlist_externals = ansible-playbook
 [testenv:lint]
 description = Enforce quality standards under {basepython} ({envpython})
 install_command = pip install {opts} {packages}
-# NOTE: The following silences the list of installed python packages during the tox run
-# NOTE: it can be commented out as needed for troubleshooting
-list_dependencies_command = echo
 commands =
   pre-commit run {posargs:--show-diff-on-failure \
     --hook-stage manual \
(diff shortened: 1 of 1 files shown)


webknjaz (changes requested):
Removing the dependency list in the CI is an exceptionally bad idea. When present, it allows comparing old CI runs with the new ones when something in the deptree breaks. Especially, since we don't have per-env lockfiles to know when something changes.
The pre-commit part is fine, I guess. Although, timing is often useful, as well as hook ids/aliases.

cidrblock:
> Removing the dependency list in the CI is an exceptionally bad idea. When present, it allows comparing old CI runs with the new ones when something in the deptree breaks. Especially, since we don't have per-env lockfiles to know when something changes. The pre-commit part is fine, I guess. Although, timing is often useful, as well as hook ids/aliases.

I figure if you are going to have bad idea, Go big or go home! ;)

Great call out about the dep version tracking in CI, we have had that problem before.

all that is removed,  I can't think of an easy way for GHA to run precommit with -v but have it without locally unless requested.....  (other than a whole new tox entry which isn't worth it)

webknjaz (approved):

ssbarnea (approved):

ssbarnea:
> > Removing the dependency list in the CI is an exceptionally bad idea. When present, it allows comparing old CI runs with the new ones when something in the deptree breaks. Especially, since we don't have per-env lockfiles to know when something changes. The pre-commit part is fine, I guess. Although, timing is often useful, as well as hook ids/aliases.
> 
> I figure if you are going to have a bad idea, Go big or go home! ;)
> 
> Great call out about the dep version tracking in CI, we have had that problem before.
> 
> all that is removed, I can't think of an easy way for GHA to run precommit with -v but have it without locally unless requested..... (other than a whole new tox entry which isn't worth it)
> 
> Thanks for `tox -q` or `-qq` suggesting, _that_ does help

I know how to enable verbosity only for CI but I would not bother to do it as I find it annoying even in CI.

PR #862, opened 2022-02-02
Reduce black verbosity

cidrblock:
- Remove the `-v` flag from black since it was just noise and will by default report individual files


ssbarnea (approved):

ssbarnea:
I wanted to propose that for few good weeks, thank for addressing it.

PR #864, opened 2022-02-02
Disable anchors and aliases when serializing to yaml

cidrblock:
This will ensure anchors and aliases are not used when we show content as yaml on the screen.

This was found during work for the `:settings` subcommand stdout ouput because several settings had some key-values in common and they were anchored and aliased.

Tests to confirm.

### DIFF
--- a/src/ansible_navigator/_yaml.py
+++ b/src/ansible_navigator/_yaml.py
@@ -56,8 +56,13 @@ class HumanDumper(Dumper):
     # pylint: disable=too-many-ancestors
     """An instance of a pyyaml Dumper.
 
-    This deviates from the base to dump a multiline string in a human readable format.
+    This deviates from the base to dump a multiline string in a human readable format
+    and disables the use of anchors and aliases.
     """
+
+    def ignore_aliases(self, _data: Any) -> bool:
+        # Return true to disable use of anchors and aliases
+        return True
 
     def represent_scalar(
         self,
--- a/src/ansible_navigator/_yaml.py
+++ b/src/ansible_navigator/_yaml.py
@@ -61,7 +61,10 @@ class HumanDumper(Dumper):
     """
 
     def ignore_aliases(self, _data: Any) -> bool:
-        # Return true to disable use of anchors and aliases
+        """Return true to disable use of anchors and aliases.
+
+        :param _data: The data used to make the determination
+        """
         return True
 
     def represent_scalar(
--- a/src/ansible_navigator/_yaml.py
+++ b/src/ansible_navigator/_yaml.py
@@ -64,6 +64,7 @@ class HumanDumper(Dumper):
         """Return true to disable use of anchors and aliases.
 
         :param _data: The data used to make the determination
+        :returns: True, indicating aliases and anchors should not be used
         """
         return True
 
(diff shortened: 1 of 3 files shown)


tests/unit/test_yaml.py
```diff
@@ -1,6 +1,10 @@
 """Tests related to the yaml abstraction layer.
 """
 
+from unittest.mock import patch
```
webknjaz, line 4: Please avoid importing `unittest`. Use pytest-native fixtures like `monkeypatch` instead.

tests/unit/test_yaml.py
```diff
@@ -9,3 +13,22 @@ def test_check_yaml_imports():
...
+
+
+@patch("ansible_navigator._yaml.HumanDumper.ignore_aliases", Dumper.ignore_aliases)
+def test_anchor_alias():
+    """Ensure anchors and aliases are present with ignore_aliases restored."""
+    data = {"integer": 42}
+    many_data = [data, data, data]
+    result = yaml_import.human_dump(many_data)
+    assert "&id" in result
+    assert "*id" in result
```
webknjaz, line 34:
```suggestion
    yaml_string = yaml_import.human_dump(many_data)
    assert "&id" in yaml_string
    assert "*id" in yaml_string
```

tests/unit/test_yaml.py
```diff
@@ -9,3 +13,22 @@ def test_check_yaml_imports():
     assert yaml_import.yaml is not None
     assert yaml_import.Dumper is not None
     assert yaml_import.Loader is not None
+
+
+def test_no_anchor_alias():
```
webknjaz, line 18: Looks like it's just a duplicated test that could be a part of a parametrized implementation.

src/ansible_navigator/_yaml.py
```diff
@@ -56,9 +56,18 @@ class HumanDumper(Dumper):
     # pylint: disable=too-many-ancestors
     """An instance of a pyyaml Dumper.
 
-    This deviates from the base to dump a multiline string in a human readable format.
+    This deviates from the base to dump a multiline string in a human readable format
+    and disables the use of anchors and aliases.
     """
 
+    def ignore_aliases(self, _data: Any) -> bool:
+        """Return true to disable use of anchors and aliases.
```
webknjaz, line 64: "true" is an implementation detail. Why should the caller care about it?
```suggestion
        """Disable the use of anchors and aliases in the given data.
```

webknjaz:
@cidrblock is the fact of improving the data representation significant to the end-users? Maybe mention this in a change note?

cidrblock:
> @cidrblock is the fact of improving the data representation significant to the end-users? Maybe mention this in a change note?

In this case the problem doesn't exist yet that I'm aware of.  It was found while working on the stdout for `:settings` which is WIP

tests/unit/test_yaml.py
cidrblock: ok

tests/unit/test_yaml.py
cidrblock: done

webknjaz (approved):

Issue #610, opened 2021-11-16
Image introspection failed (WSL2 Ubuntu 20.04)


watsonb:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
I'm trying to use `ansible-navigator` on Ubuntu 20.04 via WSL 2 with `podman`.  I followed podman install instructions here: https://www.redhat.com/sysadmin/podman-windows-wsl2

Yes, I recognize WSL2 support may be a stretch goal, but it is one I'm working toward nonetheless.

Using my `ansible-navigator.yml` config, when I run `ansible-navigator`, the specified EE image is pulled from our Azure Container Registry and I am greeted by the "welcome" page.  When I list images, I see my image.  When I try to inspect the version of Ansible and collections contained in the image, I get an error message from `ansible-navigator` stating that it failed to introspect the image and that this is probably a bug that I should report.

##### ANSIBLE-NAVIGATOR VERSION

```
ansible-navigator 1.1.0
```

Other pertinent environment information:
```bash
wsl --status
Default Distribution: Ubuntu-20.04
Default Version: 2

Windows Subsystem for Linux was last updated on 10/14/2021
WSL automatic updates are on.

Kernel version: 5.10.60.1

wsl --list --verbose
  NAME            STATE           VERSION
* Ubuntu-20.04    Running         2
```

Ubuntu 20.04 is a completely fresh (apt update) install.  Inside Ubuntu 20.04:

```bash
$ ansible-navigator --version
ansible-navigator 1.1.0

$ podman --version
podman version 3.3.1

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal

$ python --version
Python 3.8.10

$ pip --version
pip 21.3.1 from /home/ben/venv_ansible_navigator/lib/python3.8/site-packages/pip (python 3.8)
```

##### CONFIGURATION

```yml
---
ansible-navigator:
  ansible:
    config: $PWD/ansible.cfg

  ansible-runner:
    artifact-dir: $PWD/navigator_artifacts
    rotate-artifacts-count: 10
    timeout: 300

  app: welcome  # ‘collections’, ‘config’, ‘doc’, ‘images’, ‘inventory’, ‘replay’, ‘run’ or ‘welcome’

  collection-doc-cache-path: $PWD/cache.db

  color:
    enable: True
    osc4: True

  editor:
[...]

PR #866, opened 2022-02-02
Error out when `/dev/mqueue/` doesn't exist and container engine is `podman` with EE support enabled

cidrblock:
In the case of WSL2 + Ubuntu, no `/dev/mqueue` directory is present, but runner provides `ipc=host` to podman which requires POSIX messaage queue support.

Related: #610 
Related: https://github.com/ansible/ansible-runner/issues/984

Tests added.

### DIFF
--- a/docs/changelog-fragments.d/866.bugfix.md
+++ b/docs/changelog-fragments.d/866.bugfix.md
@@ -1,0 +1,5 @@
+Detect and error for missing `/dev/mqueue` directory when
+execution-environment support is enabled and the container
+engine is set to `podman`. This is a condition commonly
+found using Ubuntu with WSL2.
+-- by {user}`cidrblock`
--- a/docs/changelog-fragments.d/866.bugfix.md
+++ b/docs/changelog-fragments.d/866.bugfix.md
@@ -1,5 +1,7 @@
-Detect and error for missing `/dev/mqueue` directory when
-execution environment support is enabled and the container
-engine is set to `podman`. This is a condition commonly
-found using Ubuntu with WSL2.
+Fixed a condition where image introspection failed because
+of a missing `/dev/mqueue/` directory when execution environment
+support was enabled and the container engine was set to `podman` by
+checking early and providing an error message and suggested fix. This
+is a condition commonly found using Ubuntu with WSL2.
+
 -- by {user}`cidrblock`
--- a/docs/changelog-fragments.d/866.bugfix.md
+++ b/docs/changelog-fragments.d/866.bugfix.md
@@ -1,5 +1,5 @@
 Detect and error for missing `/dev/mqueue` directory when
-execution-environment support is enabled and the container
+execution environment support is enabled and the container
 engine is set to `podman`. This is a condition commonly
 found using Ubuntu with WSL2.
 -- by {user}`cidrblock`
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -282,19 +282,21 @@ class NavigatorPostProcessor:
                     if hint:
                         exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
 
-            # check for /dev/mqueue when using podman because runner passes ipc=host
+            # check for /dev/mqueue/ when using podman because runner passes ipc=host
             # https://github.com/ansible/ansible-navigator/issues/610
-            if config.container_engine == "podman" and not Path("/dev/mqueue").exists():
-                exit_msg = (
-                    "Execution environment support while using podman requires a"
-                    " '/dev/mqueue' directory."
-                )
-                exit_messages.append(ExitMessage(message=exit_msg))
-                hint = (
-                    "Try creating it with 'mkdir /dev/mqueue' or reference the documentation for"
-                    " your operating system related to POSIX message queues."
-                )
-                exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
+            mqueue_path = "/dev/mqueue/"
+            if config.container_engine == "podman":
+                if not Path(mqueue_path).exists() or not Path(mqueue_path).is_dir():
+                    exit_msg = (
+                        "Execution environment support while using podman requires a"
+                        f" '{mqueue_path}' directory."
+                    )
+                    exit_messages.append(ExitMessage(message=exit_msg))
+                    hint = (
+                        "Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
+                        " for your operating system related to POSIX message queues."
+                    )
+                    exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
         else:
             new_messages, new_exit_messages = check_for_ansible()
             messages.extend(new_messages)
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -281,6 +281,20 @@ class NavigatorPostProcessor:
                         )
                     if hint:
                         exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
+
+            # check for /dev/mqueue when using podman because runner passes ipc=host
+            # https://github.com/ansible/ansible-navigator/issues/610
+            if config.container_engine == "podman" and not Path("/dev/mqueue").exists():
+                exit_msg = (
+                    "Execution environment support while using podman requires a"
+                    " '/dev/mqueue' directory."
+                )
+                exit_messages.append(ExitMessage(message=exit_msg))
+                hint = (
+                    "Try creating it with 'mkdir /dev/mqueue' or reference the documentation for"
+                    " your operating system related to POSIX message queues."
+                )
+                exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
         else:
             new_messages, new_exit_messages = check_for_ansible()
             messages.extend(new_messages)
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -293,7 +293,7 @@ class NavigatorPostProcessor:
                     )
                     exit_messages.append(ExitMessage(message=exit_msg))
                     hint = (
-                        "Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
+                        f"Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
                         " for your operating system related to POSIX message queues."
                     )
                     exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -285,18 +285,20 @@ class NavigatorPostProcessor:
             # check for /dev/mqueue/ when using podman because runner passes ipc=host
             # https://github.com/ansible/ansible-navigator/issues/610
             mqueue_path = "/dev/mqueue/"
-            if config.container_engine == "podman":
-                if not Path(mqueue_path).exists() or not Path(mqueue_path).is_dir():
-                    exit_msg = (
-                        "Execution environment support while using podman requires a"
-                        f" '{mqueue_path}' directory."
-                    )
-                    exit_messages.append(ExitMessage(message=exit_msg))
-                    hint = (
-                        f"Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
-                        " for your operating system related to POSIX message queues."
-                    )
-                    exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
+            podman_no_mqueue_dir = (
+                config.container_engine == "podman" and not Path(mqueue_path).is_dir()
+            )
+            if podman_no_mqueue_dir:
+                exit_msg = (
+                    "Execution environment support while using podman requires a"
+                    f" '{mqueue_path}' directory."
+                )
+                exit_messages.append(ExitMessage(message=exit_msg))
+                hint = (
+                    f"Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
+                    " for your operating system related to POSIX message queues."
+                )
+                exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
         else:
             new_messages, new_exit_messages = check_for_ansible()
             messages.extend(new_messages)
--- a/tests/unit/configuration_subsystem/test_posix_message_queue.py
+++ b/tests/unit/configuration_subsystem/test_posix_message_queue.py
@@ -1,10 +1,10 @@
-"""Tests related to the auto detection of the container engine.
+"""Tests related to a missing ``/dev/mqueue`` directory when using ``podman``.
 """
 from pathlib import Path
 
 
 def test_posix_message_queue(monkeypatch, generate_config):
-    """Check for error messages related to missing /dev/mqueue and podman.
+    """Check for error messages related to missing ``/dev/mqueue`` and ``podman``.
 
     :param monkeypatch: The monkeypatch fixture
     :param generate_config: The configuration generator fixture
--- a/tests/unit/configuration_subsystem/test_posix_message_queue.py
+++ b/tests/unit/configuration_subsystem/test_posix_message_queue.py
@@ -1,12 +1,11 @@
-"""Tests related to a missing ``/dev/mqueue`` directory when using ``podman``.
-"""
+"""Tests related to a missing ``/dev/mqueue`` directory when using ``podman``."""
 from pathlib import Path
 
 
 def exists(_arg):
     """Return false when checking for a file or directory path.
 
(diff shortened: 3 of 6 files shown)


docs/changelog-fragments.d/866.bugfix.md
```diff
@@ -0,0 +1,5 @@
+Detect and error for missing `/dev/mqueue` directory when
```
webknjaz, line 1:
```suggestion
Started erroring out loudly on missing `/dev/mqueue/` directory when
```
(past tense for consistency —see the RTD preview + trailing slash to emphasize that it's a dir)

docs/changelog-fragments.d/866.bugfix.md
```diff
@@ -0,0 +1,5 @@
...
+execution environment support is enabled and the container
+engine is set to `podman`. This is a condition commonly
+found using Ubuntu with WSL2.
+-- by {user}`cidrblock`
```
webknjaz, line 5: IIRC Markdown would inline this into the paragraph in there's no empty line
```suggestion

-- by {user}`cidrblock`
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -281,6 +281,20 @@ def execution_environment(self, entry: SettingsEntry, config) -> PostProcessorRe
                         )
                     if hint:
                         exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
+
+            # check for /dev/mqueue when using podman because runner passes ipc=host
+            # https://github.com/ansible/ansible-navigator/issues/610
+            if config.container_engine == "podman" and not Path("/dev/mqueue").exists():
```
webknjaz, line 287: Maybe use is_dir since the error message claims exactly that?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -281,6 +281,20 @@ def execution_environment(self, entry: SettingsEntry, config) -> PostProcessorRe
...
+                exit_msg = (
+                    "Execution environment support while using podman requires a"
+                    " '/dev/mqueue' directory."
```
webknjaz, line 290: Same as with the change note
```suggestion
                    " '/dev/mqueue/' directory."
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -281,6 +281,20 @@ def execution_environment(self, entry: SettingsEntry, config) -> PostProcessorRe
...
+                )
+                exit_messages.append(ExitMessage(message=exit_msg))
+                hint = (
+                    "Try creating it with 'mkdir /dev/mqueue' or reference the documentation for"
```
webknjaz, line 294: I think it may be reasonable to put this path into a module-level constant
```suggestion
                    "Try creating it with 'mkdir /dev/mqueue/' or reference the documentation for"
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,36 @@
+"""Tests related to a missing ``/dev/mqueue`` directory when using ``podman``."""
```
webknjaz, line 1:
```suggestion
"""Tests related to a missing ``/dev/mqueue/`` directory when using ``podman``."""
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,36 @@
...
+    """Return false when checking for a file or directory path.
+
+    :param _arg: The file or directory path to look for
+    :returns: False, indicating it does not exist
+    """
+    return False
+
+
+def test_posix_message_queue_ee(monkeypatch, generate_config):
+    """Check for error messages related to missing ``/dev/mqueue`` and ``podman`` when ee true.
```
webknjaz, line 15:
```suggestion
    """Check for error messages related to missing ``/dev/mqueue/`` and ``podman`` when ee true.
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,36 @@
...
+
+    :param monkeypatch: The monkeypatch fixture
+    :param generate_config: The configuration generator fixture
+    """
+    monkeypatch.setattr(Path, "exists", exists)
```
webknjaz, line 20: Since this is copy-pasted in all the tests, I recommend putting it into a local fixture.

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,36 @@
+"""Tests related to a missing ``/dev/mqueue`` directory when using ``podman``."""
+from pathlib import Path
+
+
+def exists(_arg):
```
webknjaz, line 5: maybe rename this to

```suggestion
def _never_exists(_arg):
```

webknjaz (commented):
Maybe also add a note to the FAQ document?

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: tests were completely rewritten....

cidrblock: got it! ty

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: tests were completely rewritten....

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: tests were completely rewritten....

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: got it TY!

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: used a var so it's the same everywhere, good idea

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: used a var so it's consistent, thx

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: added it, good idea

docs/changelog-fragments.d/866.bugfix.md
cidrblock: added, looks better

docs/changelog-fragments.d/866.bugfix.md
cidrblock: rephrased. lmk

ganeshrn (approved):
I think the work that @webknjaz is doing with enabling windows + WSL2 in CI for `ansible-language-server` can be reused in ansible-navigator project to test WSL2 environments in CI.

docs/changelog-fragments.d/866.bugfix.md
webknjaz: Looks good, resolving.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -281,6 +281,22 @@ def execution_environment(self, entry: SettingsEntry, config) -> PostProcessorRe
                         )
                     if hint:
                         exit_messages.append(ExitMessage(message=hint, prefix=ExitPrefix.HINT))
+
+            # check for /dev/mqueue/ when using podman because runner passes ipc=host
+            # https://github.com/ansible/ansible-navigator/issues/610
+            mqueue_path = "/dev/mqueue/"
+            if config.container_engine == "podman":
+                if not Path(mqueue_path).exists() or not Path(mqueue_path).is_dir():
```
webknjaz, line 289: Plese combine these clauses to dedent the whole conditional block:
```suggestion
            podman_engine_with_no_mqueue_path = (
                config.container_engine == "podman" and
                not Path(mqueue_path).exists() or not Path(mqueue_path).is_dir()
            (
            if podman_engine_with_no_mqueue_path:
```
This way, there's a named conditional with a clearer purpose/semantics + less nesting.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -281,6 +281,22 @@ def execution_environment(self, entry: SettingsEntry, config) -> PostProcessorRe
...
+                    exit_msg = (
+                        "Execution environment support while using podman requires a"
+                        f" '{mqueue_path}' directory."
+                    )
+                    exit_messages.append(ExitMessage(message=exit_msg))
+                    hint = (
+                        "Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
```
webknjaz, line 296: Looks like this is missing an f-string prefix.
```suggestion
                        f"Try creating it with 'mkdir {mqueue_path}' or reference the documentation"
```
Is this not covered with tests?

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,33 @@
...
+    :param is_dir: The return value to set for ``pathlib.Path.is_dir``
+    :param ee_support: The value to set for ``--ee``
+    :param engine: The value to set for ``--ce``
+    :param generate_config: The configuration generator fixture
+    """
+    mqueue_msg = (
+        "Execution environment support while using podman requires a '/dev/mqueue/' directory."
+    )
+    with patch("pathlib.Path.exists", return_value=exists):
+        with patch("pathlib.Path.is_dir", return_value=is_dir):
```
webknjaz, line 28: Please use the built-in `monkeypatch` fixture — it's a canonical way of doing the same in the pytest ecosystem. FWIW importing `unittest` should be avoided to prevent accidentally limiting pytest's capabilities. Why is it not forbidden via pylint yet?

webknjaz: Besides, it'd allow for dropping the nesting.

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,33 @@
...
+
+from unittest.mock import patch
+
+import pytest
+
+
+@pytest.mark.parametrize("exists,", [True, False], ids=["exists-True", "exists-False"])
+@pytest.mark.parametrize("is_dir,", [True, False], ids=["is_dir-True", "is_dir-False"])
+@pytest.mark.parametrize("ee_support,", [True, False], ids=["ee_support-True", "ee_support-False"])
+@pytest.mark.parametrize("engine,", ["podman", "docker"], ids=["engine-podman", "engine-docker"])
```
webknjaz, line 11:
```suggestion
@pytest.mark.parametrize("is_dir", [True, False], ids=["is_dir-True", "is_dir-False"])
@pytest.mark.parametrize("ee_support", [True, False], ids=["ee_support-True", "ee_support-False"])
@pytest.mark.parametrize("engine", ["podman", "docker"], ids=["engine-podman", "engine-docker"])
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,33 @@
...
+    :param engine: The value to set for ``--ce``
+    :param generate_config: The configuration generator fixture
+    """
+    mqueue_msg = (
+        "Execution environment support while using podman requires a '/dev/mqueue/' directory."
+    )
+    with patch("pathlib.Path.exists", return_value=exists):
+        with patch("pathlib.Path.is_dir", return_value=is_dir):
+            response = generate_config(params=["--ce", engine, "--ee", str(ee_support)])
+            if ee_support is False or engine == "docker" or (exists is True and is_dir is True):
```
webknjaz, line 30: This is quite complex. It's best not to have too much logic in tests as it makes them fragile and hard to follow.
Besides, this complex condition probably has some meaning so I'd put it into a var.

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,33 @@
...
+                assert mqueue_msg not in [exit_msg.message for exit_msg in response.exit_messages]
+            else:
+                assert mqueue_msg in [exit_msg.message for exit_msg in response.exit_messages]
```
webknjaz, line 33: This is basically copy-paste. It could be optimized so that it's not as complex.

tests/unit/configuration_subsystem/test_posix_message_queue.py
```diff
@@ -0,0 +1,33 @@
...
+
+    :param exists: The return value to set for ``pathlib.Path.exists``
+    :param is_dir: The return value to set for ``pathlib.Path.is_dir``
+    :param ee_support: The value to set for ``--ee``
+    :param engine: The value to set for ``--ce``
+    :param generate_config: The configuration generator fixture
+    """
+    mqueue_msg = (
+        "Execution environment support while using podman requires a '/dev/mqueue/' directory."
+    )
```
webknjaz, line 26:
```suggestion
    )
    mqueue_msg_should_exist = ee_support is False or engine == "docker" or (exists is True and is_dir is True)
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
webknjaz, line 33: This could easily become
```suggestion
            mqueue_msg_exists = mqueue_msg in [exit_msg.message for exit_msg in response.exit_messages]
            assert mqueue_msg_exists is mqueue_msg_should_exist
```

webknjaz: Or even better:
```suggestion
            mqueue_msg_exists = any(exit_msg.message == mqueue_msg for exit_msg in response.exit_messages)
            assert mqueue_msg_exists is mqueue_msg_should_exist
```

tests/unit/configuration_subsystem/test_posix_message_queue.py
webknjaz: I'd stick with this function and not fall into the trap of stdlib unittest.

tests/unit/configuration_subsystem/test_posix_message_queue.py
webknjaz: If that patch is really needed (I don't think it is), there's a package called `pytest-mocker` that exposes all of this as a pytest fixture.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
relrod: I actually like the original here better because it's easier than having to guess which way Python will associate the set of conditionals. If you do take @webknjaz's suggestion, please wrap the second set of conditionals in parens so it's easy to see at a glance what the actual intent is for this to be parsed.

webknjaz: Also, could just name that second conditional too.

webknjaz: Looking closer, I realized that there's `Path(mqueue_path).exists()` before `is_dir()` — why is it even necessary? Wouldn't `is_dir()` only be true if it exists?

cidrblock: yep! kinda funny :)

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: did something similar

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: @relrod  let me know if I captured your feedback. A lot changed, but I think I got it...

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: Didn't need it :)

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: Simpler now I think, LMK

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: Agree. Done I think

cidrblock: Did something similar, lmk. Good stuff

tests/unit/configuration_subsystem/test_posix_message_queue.py
cidrblock: This isn't really relevant after the last push. Let me know

webknjaz (approved):

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
ssbarnea: Be warned that is_dir will returns false for symlinked folders. They act as folders but is_dir reports false.

PR #870, opened 2022-02-04
Documentation improvements in one `test_direct_interactive_ee_with_volmount.py` (D/DAR)

cidrblock:
This puts the one-line doc string on one-line.

This file was not in the flake8 files, so no changes necessary there.

### DIFF
--- a/tests/integration/actions/collections/test_direct_interactive_ee_with_volmount.py
+++ b/tests/integration/actions/collections/test_direct_interactive_ee_with_volmount.py
@@ -1,5 +1,4 @@
-"""Tests for collections from CLI, interactive, with an EE and volmount.
-"""
+"""Tests for collections from CLI, interactive, with an EE and volume mount."""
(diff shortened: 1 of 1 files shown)


webknjaz (approved):

webknjaz:
> (D/DAR)

Are you sure it's actually related to these? It seems like a spelling improvement, no?

cidrblock:
> > (D/DAR)
> 
> Are you sure it's actually related to these? It seems like a spelling improvement, no? It would be helpful if the precise rules were mentioned.

Yeah, you can see it here:  https://github.com/ansible/ansible-navigator/runs/5058950003?check_suite_focus=true#step:7:38

My goal is to get consistency and better docs, yeah, that satisfies the linters but that's not the motivation.  The linters are just a tool, not the reason.

When there isn't a flake8 changes, I'll try to remember to add the err code for you in the future.

PR #871, opened 2022-02-04
Reduce false fixture changes for `config` integration tests

### LINKED ISSUE #346: [tests] Sanitize DEFAULT_LOCAL_TMP in fixtures
See #345 for some examples

cidrblock:
Resolves #346 

This should reduce the number of fixture changes due to changes in developer environments and per-run values. 

This changes the way we mask in the fixtures.  Rather than masking the whole line, we'll just mask the last column.

- Additionally add a couple new ansible settings that change per environment.
- Allow a couple other step's output to be masked

The masking change is in the base.py file, the fixtures changes are related to the way we'll mask now.


tests/integration/actions/config/base.py
```diff
@@ -74,11 +77,21 @@ def test(self, request, tmux_session, step):
...
+                "CACHE_PLUGIN_CONNECTION",
+                "COLLECTIONS_PATHS",
+                "DEFAULT_CALLBACK_PLUGIN_PATH",
+                "DEFAULT_LOCAL_TMP",
+            ]
+            # Determine if a menu if showing
+            match = re.search("CURRENT VALUE", received_output[0])
+            if match:
+                mask_start, mask_end = match.span()
+                mask = (mask_end - mask_start) * "X"
```
webknjaz, line 91: Isn't this just
```suggestion
                mask = len("CURRENT VALUE") * "X"
```
?

webknjaz: (which then could be moved outside the if-block?)

cidrblock: I prefer using the results of the span in this case rather than repeating the string..............

cidrblock: (probably b/c it's a reusable pattern in case the search was done against a regex)

tests/integration/actions/config/base.py
```diff
@@ -74,11 +77,21 @@ def test(self, request, tmux_session, step):
...
-                if any(m in line for m in maskables):
-                    received_output[idx] = mask
+            maskables = [
+                "BECOME_PLUGIN_PATH",
+                "CACHE_PLUGIN_CONNECTION",
+                "COLLECTIONS_PATHS",
+                "DEFAULT_CALLBACK_PLUGIN_PATH",
+                "DEFAULT_LOCAL_TMP",
+            ]
+            # Determine if a menu if showing
```
relrod, line 87:
```suggestion
            # Determine if a menu is showing
```

tests/integration/actions/config/base.py
relrod: I think I'd opt to not use regex here at all, it's quite a hammer, just to search for a simple string and roughly an order of magnitude slower:

```

In [8]: %timeit "CURRENT VALUE" in "FOO     BAR      CURRENT VALUE     BAZ"
38.6 ns ± 0.169 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [9]: %timeit re.search("CURRENT VALUE", "FOO     BAR      CURRENT VALUE     BAZ")
360 ns ± 0.847 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

relrod: Ah, nevermind, you use `mask_start` below.

cidrblock: Since it was a test, I wasn't all too concerned about the re, we do need to pay attention to those in the user facing code....  But now that you mention it, I think str.find would work.  

let's see how it looks

tests/integration/actions/config/base.py
cidrblock: ha :) thx @relrod

tests/integration/actions/config/base.py
relrod: Yeah, I was just going to come back and suggest that.

```
In [10]: %timeit "FOO     BAR      CURRENT VALUE     BAZ".find("CURRENT VALUE")
87.2 ns ± 0.126 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

relrod: And I see @webknjaz beat me to the suggestion anyway :) I'll show myself out.

tests/integration/actions/config/base.py
```diff
@@ -74,11 +77,21 @@ def test(self, request, tmux_session, step):
...
+            match = re.search("CURRENT VALUE", received_output[0])
+            if match:
+                mask_start, mask_end = match.span()
+                mask = (mask_end - mask_start) * "X"
+                for idx, line in enumerate(received_output):
+                    if any(f"│{m}" in line for m in maskables):
+                        received_output[idx] = received_output[idx][0:mask_start] + mask
```
webknjaz, line 94: I've taken a stub at dropping `re` and it seems cleaner with
```suggestion
            def _line_needs_masking(line):
                return any(f"│{m}" in line for m in maskables)

            mask_marker = "CURRENT VALUE"
            mask = len(mask_marker) * "X"

            try:
                mask_start = received_output[0].index(mask_marker)
            except ValueError:
                mask_start = None

            if mask_start is not None:
                received_output = [
                    line[:mask_start] + mask if _line_needs_masking(line)
                    else line
                    for line in received_output
                ]
```

webknjaz: @cidrblock the only thing that is unclear is why the replaced thing is not just the masked value but the rest of the line too?

tests/integration/actions/config/base.py
webknjaz: I've added an implementation suggestion not using `re` below: https://github.com/ansible/ansible-navigator/pull/871/files#r799593279

tests/integration/actions/config/base.py
```diff
@@ -74,11 +77,22 @@ def test(self, request, tmux_session, step):
...
+            ]
+            # Determine if a menu is showing
+            mask_column_name = "CURRENT VALUE"
+            column_start = received_output[0].find(mask_column_name)
+            column_exists = column_start != -1
+            if column_exists:
+                mask = len(mask_column_name) * "X"
+                for idx, line in enumerate(received_output):
+                    if any(f"│{m}" in line for m in maskables):
+                        received_output[idx] = received_output[idx][0:column_start] + mask
```
webknjaz, line 95: Since the previous conversation got hidden — could you explain why it is necessary to cut the end of the line together with the masked thing?

tests/integration/actions/config/base.py
cidrblock: The length of the value to mask can change, so I used the minimum possible column width which is always the length of the column header.  If we masked the the actual of the value with X * length of the value, we might still changes between developer system where the fixtures were generated.

tests/integration/actions/config/base.py
```diff
@@ -74,11 +77,22 @@ def test(self, request, tmux_session, step):
...
+                "BECOME_PLUGIN_PATH",
+                "CACHE_PLUGIN_CONNECTION",
+                "COLLECTIONS_PATHS",
+                "DEFAULT_CALLBACK_PLUGIN_PATH",
+                "DEFAULT_LOCAL_TMP",
+            ]
+            # Determine if a menu is showing
+            mask_column_name = "CURRENT VALUE"
+            column_start = received_output[0].find(mask_column_name)
+            column_exists = column_start != -1
```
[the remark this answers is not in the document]
cidrblock: help(str.find) returns the following:

```Help on method_descriptor:

find(...)
    S.find(sub[, start[, end]]) -> int
    
    Return the lowest index in S where substring sub is found,
    such that sub is contained within S[start:end].  Optional
    arguments start and end are interpreted as in slice notation.
    
    Return -1 on failure.
```

because it would take a doc look up I did added the named variable.

tests/integration/actions/config/base.py
cidrblock: I think I answered this above, at least what I think you are asking:

The length of the value to mask can change based on the developers workstation, so I used the minimum possible column width which is always the length of the column header. If we masked the the actual of the value with X * length of the value, we might still changes between developer system where the fixtures were generated.

tests/integration/actions/config/base.py
webknjaz: Ah, okay. Although, the second question would be — why not set a fixed TTY size in tests so that it would not depend on the developer's setup?

tests/integration/actions/config/base.py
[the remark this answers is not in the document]
webknjaz: I know that, it just looks ugly :)

tests/integration/actions/config/base.py
cidrblock: we do, but that wouldn't cover a case where the value didn't go to the far right side of the screen.

tests/integration/actions/config/base.py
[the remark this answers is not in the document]
cidrblock: It looks magic (number) to me!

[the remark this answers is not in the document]
webknjaz: Yeah. Even though I _know_ that it's likely coming from C with its limitations of the ancient times and missing exception-raising mechanism, I'd at least make it named like `SEARCH_FAILURE_CONSTANT = 1`. It's nice that WPS is able to catch some of such cases and maybe eventually we'll be more protected against similar code smells...

tests/integration/actions/config/base.py
webknjaz: Interesting... FWIW I'm resolving this now.

tests/integration/actions/config/base.py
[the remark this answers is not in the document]
webknjaz: One way of making this piece of code clearer would be
```python
try:
    column_start = received_output[0].index(mask_column_name)
except ValueError:
    pass  # no match, nothing to do
else:
    ...  # replacement logic here
```

webknjaz (approved):

tests/integration/actions/config/base.py
cidrblock: thanks for the help here @relrod and @webknjaz

PR #872, opened 2022-02-04
Guard against using unittest ecosystem within pylint

webknjaz:
This patch implements recommendations to use `pytest`-native fixtures
instead of importing `unittest`-world helper untilities directly using
a built-in `pylint` rule `preferred-module`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,14 @@ profile = "black" # Avoid conflict with black
 skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to ansible-test sanity ruleset
 
 [tool.pylint]
+    [tool.pylint.imports]
+    preferred-modules = [
+      # NOTE: The unittest replacements below help keep
+      # NOTE: the tests pytest ecosystem-oriented.
+      "unittest:pytest",
+      "unittest.mock:pytest-mock",
+    ]
+
     [tool.pylint.master]
     # tm_tokenize is virtually vendored and shouldn't be linted as such
     ignore = "tm_tokenize"
--- a/tests/integration/_cli2runner.py
+++ b/tests/integration/_cli2runner.py
@@ -1,5 +1,6 @@
 """test from CLI up to runner
 """
+# pylint: disable=preferred-module  # FIXME: remove once migrated per GH-872
 from unittest import mock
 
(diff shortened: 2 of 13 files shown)


tests/integration/_cli2runner.py
```diff
@@ -1,5 +1,6 @@
 """test from CLI up to runner
 """
+# pylint: disable=preferred-module  # FIXME: remove once migrated per GH-872
```
cidrblock, line 3: Should we renable the check after the import so nothing else sneaks through in the future

cidrblock (approved):
Approving to avoid mixing the 2.

But..........  Some clean up help here would be appreciated, if anyone wants to pick some off.

Also- Let's not push too hard in the future if another comes through until these are remediated and there are repeatable examples in the code base that we can point to.

tests/integration/_cli2runner.py
webknjaz: Not sure. But given that nothing else is blocked, it shouldn't matter now.

pyproject.toml
```diff
@@ -67,6 +67,14 @@ profile = "black" # Avoid conflict with black
 skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to ansible-test sanity ruleset
 
 [tool.pylint]
+    [tool.pylint.imports]
+    preferred-modules = [
+      # NOTE: The unittest replacements belog help keep
```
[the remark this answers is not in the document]
webknjaz: *needs to learn to type in words*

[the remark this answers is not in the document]
webknjaz: Oh, that's "below"!

tests/integration/_cli2runner.py
webknjaz: @cidrblock resolve this if you agree.

cidrblock: I would prefer it get added, but you are right, it doesn't matter today, but this is one of those things that I remember because I've paid the price for a single, global pylint disable in the past which let a growing file get more and more errors.

I also think it's a good pattern to get into, as an example in the code, because I was suprised to learn about this and I often wonder if others would be too.

That said, I will leave it up to you to decide and see what you do :)

PR #873, opened 2022-02-05
Remove duplicate pydocstyle check from precommit configuration

cidrblock:
This removes the duplicate entry for pydocstyle from the pre-commit configuration.

Currently it is run via flake, included through the installation of the we make style guide.

This should reduce confusion and complexity for those cleaning up docstring violations about where thing should be removed and how doc string improvement suggestions are generated.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -151,7 +151,7 @@ repos:
     language_version: python3
     additional_dependencies:
     - darglint
-    - flake8-docstrings
+    - flake8-docstrings # uses pydocstyle
     stages:
     - manual
   - id: flake8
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

.pre-commit-config.yaml
```diff
@@ -193,7 +151,7 @@ repos:
     language_version: python3
     additional_dependencies:
     - darglint
-    - flake8-docstrings
+    - flake8-docstrings # uses pydocstyle
```
webknjaz, line 154: That's sorta the only thing it does. Why document it?

ssbarnea: Because not everyone knows that pydocstyle is use in that project if you cannot even find any matches using `pydocstyle` as a search string. I would have not added the comment if the plugin name was named `flake8-pydocstyle`.

PR #876, opened 2022-02-05
Format the precommit file using RH vscode plugin defaults

cidrblock:
This simply reformats the `pre-commit` config file to the defaults used by the RedHat yaml vscode plugin.

This was done to allow users of the RH yaml vscode plugin to enable "format on save" without needing to needing to be concerned about scope creep in a PR intended to change it's functionality.

e.g. https://github.com/ansible/ansible-navigator/pull/873/commits/f5d85ba6e1fbfcdad834242223b489877ede99ff

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,307 +1,305 @@
 ---
-
 default_language_version:
   # ensures that we get same behavior on CI(s) as on local machines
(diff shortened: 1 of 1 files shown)


webknjaz (commented):
I would've preferred a denser formatting style with yamlfmt

PR #877, opened 2022-02-05
Format the precommit file using RH vscode plugin defaults

cidrblock:
This simply reformats the pre-commit config file to the defaults used by the RedHat yaml vscode plugin.

This was done to allow users of the RH yaml vscode plugin to enable "format on save" without needing to needing to be concerned about formatting scope creep in a PR intended to simply modify it's functionality.

Using the Redhat vscode plugin without overriding the defaults will also provide us insight to what many of our user's experience which is really valuable.

e.g. https://github.com/ansible/ansible-navigator/commit/f5d85ba6e1fbfcdad834242223b489877ede99ff

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,265 +1,263 @@
 ---
-
 default_language_version:
   # ensures that we get same behavior on CI(s) as on local machines
   python: python3.10
 repos:
-- repo: https://github.com/asottile/add-trailing-comma.git
-  rev: v2.2.1
-  hooks:
-  - id: add-trailing-comma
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):
Glad to see that someone else shares my pain and is trying to find a cure. My only question is if that is the only file causing this in the repo?

cidrblock:
> Glad to see that someone else shares my pain and is trying to find a cure. My only question is if that is the only file causing this in the repo?

There's some yaml fixtures, but they are rarely touched. IMO, we can deal with those when they become an issue.

PR #888, opened 2022-02-06
Correct spelling of `collecion_cache` to `collection_cache`

cidrblock:
This also triggered the need to correctly type it.

The ignore is a result of 

```
src/ansible_navigator/initialization.py:30:5: error: Relative import climbs too
many namespaces  [misc]
        from ...share.ansible_navigator.utils import KeyValueStore
```

Since the `KeyValueStore` def is in the share directory.

### DIFF
--- a/src/ansible_navigator/initialization.py
+++ b/src/ansible_navigator/initialization.py
@@ -6,6 +6,7 @@ import logging
 import os
 import sys
 
+from typing import TYPE_CHECKING
 from typing import Dict
 from typing import List
 from typing import NoReturn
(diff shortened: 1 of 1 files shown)


src/ansible_navigator/initialization.py
```diff
@@ -23,6 +24,11 @@
 from .utils import find_settings_file
 
 
+if TYPE_CHECKING:
+
+    from ...share.ansible_navigator.utils import key_value_store  # type: ignore[misc]
```
relrod, line 29: This import, and the required ignore for it, seem odd to me. Time to move `key_value_store` into `src/ansible_navigator`? This would let us kill the (also weird to me) `_get_kvs()` below, too.

ssbarnea (approved):

src/ansible_navigator/initialization.py
cidrblock: Maybe? Let's talk more about this...........  It just feels weird to me if we were pulled navigator internals into into the EE to introspect it. Maybe that's not an issue.  We could maybe talk about a new parameter or flag to do the introspection and maybe pick that off early in cli.py?

BTW, when I did that TYPING ...  I was thinking about the fact it was odd and there should be a better way.

cidrblock: @relrod ping so we remember to talk more about this.

PR #896, opened 2022-02-07
Integrate gitlint into pre-commit vetting

webknjaz:
This adds a small check for the last commit message per best practices to vetting. It is not expected to be integrated as enforced in the current state but would be useful for some folks to try it out.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,17 @@ default_language_version:
   # ensures that we get same behavior on CI(s) as on local machines
   python: python3.10
 repos:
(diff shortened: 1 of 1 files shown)


cidrblock:
This is cool, let's keep on the list for some time in the future and see what the team thinks about vetting it...

With all the changes here over the past month and the amount of cleanup work in progress, I really think we should wait.

cidrblock:
converting to draft for now

cidrblock:
This has gotten a little stale.  Closing, feel free to reopnen as needed in the future.

PR #899, opened 2022-02-07
Remove a few abbreviations in tm_tokenize

cidrblock:
This removes a few abbreviations in the tm_tokenize subsystem.

Normally hesitant to make significant changes here, but the scope in which these vars are used is small.

Tested locally in addition to coverage provided by GHA which gives this a good run as well.


cidrblock:
yeah.............  we can totally ignore it.  Good call.

PR #901, opened 2022-02-08
Normalize references to screen/viewport height and width

cidrblock:
This normalizes the way we refer to the height and width of both the screen and viewport.

There was a mix of x,y,_h,_w.  It is simply width and height now throughout.

Testing of resize, scrolling and refresh were done locally, in addition to the coverage GHAs will provide for rendering.

### DIFF
--- a/src/ansible_navigator/ui_framework/curses_window.py
+++ b/src/ansible_navigator/ui_framework/curses_window.py
@@ -56,7 +56,7 @@ class CursesWindow:
 
         self._screen: Window
         self.win: Window
-        self._screen_miny = 3
+        self._screen_min_height = 3
         self._prefix_color = 8
         self._theme_dir: str
         self._term_osc4_support: bool
@@ -65,7 +65,7 @@ class CursesWindow:
         self._set_colors()
 
     @property
-    def _screen_w(self) -> int:
+    def _screen_width(self) -> int:
         """return the screen width
 
         :return: the current screen width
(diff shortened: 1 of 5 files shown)


src/ansible_navigator/ui_framework/curses_window.py
```diff
@@ -125,39 +125,45 @@ def _add_line(
         :param prefix: The prefix for the line
         :type prefix: str or None
         """
-        win = window
```
relrod, line 128: Given this change and the subsequent ones -- should we standardize on `window` instead of `win` everywhere?

```
rick@mba ansible-navigator % rg '\.win\b'
src/ansible_navigator/ui_framework/form_handler_text.py
36:        self._add_line(self.win, 0, cline)
37:        self.win.clrtoeol()
90:            self._add_line(self.win, 0, tuple([clp]))
92:            self.win.move(0, 0)
96:            char = self.win.getch()
105:            self.win.refresh()

src/ansible_navigator/ui_framework/form_handler_options.py
47:            self._add_line(self.win, idx, ([clp_option_code, clp_text]))
60:            char = self.win.getch()

src/ansible_navigator/ui_framework/form_handler_button.py
45:        self._add_line(self.win, 0, ([clp_button]))
54:            char = self.win.getch()

src/ansible_navigator/ui_framework/ui.py
330:        self._one_line_input.win = curses.newwin(1, self._screen_width, input_at, 1)
331:        self._one_line_input.win.keypad(True)

src/ansible_navigator/ui_framework/curses_window.py
58:        self.win: Window

src/ansible_navigator/ui_framework/form_presenter.py
165:            form_field.win = window
194:        form_field.win = window
206:        form_field.win = window
280:            form_field.window_handler.win = form_field.win
```

relrod (approved):

src/ansible_navigator/ui_framework/curses_window.py
cidrblock: probably.........  in this case, I got rid of it because it made no sense...

The rest of the ones above all subclass the curses_window which has the self.win.... probably no resaon not to do it.  let me see how long it takes.

Issue #697, opened 2021-12-10
Switch to the use of `Pathlib` where possible


cidrblock:
as noted in #624

Switch to the use of `Pathlib` where possible away from `os.path`

webknjaz:
https://gitlab.com/RoPP/flake8-use-pathlib

tinvaan:
Fwiw, https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

PR #906, opened 2022-02-08
Migrate to pathlib in `catalog_collections`

cidrblock:
This migrates the catalog collection script to pathlib away from file and directory paths created from strings and using os.path.xxx.

Additional testing was done locally:
- With the default image
- With ee-minimal
- With ee-supported

Before and after deleting the collection cache.

Pulling this one down and trying it out would be a great help.  Admittedly I am not too familiar with pathlib, although more now :)

Fixtures were updates as a result of pathlib removing the trailing slash: https://bugs.python.org/issue21039

Partial #697


share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -112,6 +112,29 @@ def _generate_checksum(file_path: str, relative_path: str) -> Dict:
...
+
+        :param top: The root directory
+        :param top_down: Scan top-down or bottom-up
+        :param follow_links: Visit directories pointed to by symlinks
+        :yields: The top directory, subdirectories and non directories
+        """
+        names = list(top.iterdir())
+
+        dirs = (node for node in names if node.is_dir() is True)
+        nondirs = (node for node in names if node.is_dir() is False)
```
relrod, line 126: Consider only iterating once here. `is_dir()` [calls `stat()`](https://github.com/python/cpython/blob/80e4f262aa27a39abf3fadc19a6323fea4607a8f/Lib/pathlib.py#L1242) and I suspect in a directory with a lot of nodes (and potentially on NFS or something), this can be slow to have to do multiple times. 

Something like this (feel free to pretty it up and expand the if expression):

```python
dirs, nondirs = [], []
for node in names:
    dirs.append(node) if node.is_dir() else nondirs.append(node)
```

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -112,6 +112,29 @@ def _generate_checksum(file_path: str, relative_path: str) -> Dict:
         }
         return res
 
+    def _path_walk(self, top: Path, top_down: bool = False, follow_links: bool = False):
```
relrod, line 115: Prefer moving this out to a util function. It doesn't seem to use the class instance anywhere and it seems more general.

ssbarnea (approved):

relrod (changes requested):
I want to review this more before it merges, I think `_path_walk()` has some issues in my quick testing, but I need to test some more.

relrod:
I'll work on this some more probably this weekend, but here's what I'm seeing (mostly as a note to myself for later):

First, this is my directory structure for testing:

```
In [22]: !tree baz
baz
├── a.txt
├── b.txt
├── dir1
│   └── c.txt
└── dir2
    ├── d.txt
    └── dir3
        └── e.txt

3 directories, 5 files
```

I have your `_path_walk` pulled out into a function called `path_walk` with no modifications other than changing its name and recursive call to point to the right thing and removing `self` as an argument. There are a fair number of differences (yours never seems to show directories - the baz entry should have `['dir2', 'dir3']` as the second element of the tuple, for example).

```
In [16]: for a, b, c in list(path_walk(Path("baz"), top_down=True)):
    ...:     print((a, list(b), list(c)))
    ...:
(PosixPath('baz'), [], [PosixPath('baz/b.txt'), PosixPath('baz/a.txt')])
(PosixPath('baz/dir2'), [], [PosixPath('baz/dir2/d.txt')])
(PosixPath('baz/dir2/dir3'), [], [PosixPath('baz/dir2/dir3/e.txt')])
(PosixPath('baz/dir1'), [], [PosixPath('baz/dir1/c.txt')])

In [17]: for a, b, c in list(path_walk(Path("baz"), top_down=False)):
    ...:     print((a, list(b), list(c)))
    ...:
(PosixPath('baz'), [], [PosixPath('baz/b.txt'), PosixPath('baz/a.txt')])
(PosixPath('baz/dir2'), [], [PosixPath('baz/dir2/d.txt')])
(PosixPath('baz/dir2/dir3'), [], [PosixPath('baz/dir2/dir3/e.txt')])
(PosixPath('baz/dir2/dir3'), [], [])
(PosixPath('baz/dir2'), [], [])
(PosixPath('baz/dir1'), [], [PosixPath('baz/dir1/c.txt')])
(PosixPath('baz/dir1'), [], [])
(PosixPath('baz'), [], [])

In [18]: import os

In [19]: for a, b, c in list(os.walk("baz", topdown=True)):
    ...:     print((a, b, c))
    ...:
('baz', ['dir2', 'dir1'], ['b.txt', 'a.txt'])
('baz/dir2', ['dir3'], ['d.txt'])
('baz/dir2/dir3', [], ['e.txt'])
('baz/dir1', [], ['c.txt'])

In [20]: for a, b, c in list(os.walk("baz", topdown=False)):
    ...:     print((a, b, c))
    ...:
('baz/dir2/dir3', [], ['e.txt'])
('baz/dir2', ['dir3'], ['d.txt'])
('baz/dir1', [], ['c.txt'])
('baz', ['dir2', 'dir1'], ['b.txt', 'a.txt'])
```

cidrblock:
@relrod Switched to a glob per plugin directory, cleaner and smaller.

share/ansible_navigator/utils/catalog_collections.py
cidrblock: The path_walk function was removed entirely

share/ansible_navigator/utils/catalog_collections.py
cidrblock: This function replaced with a glob

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -116,62 +115,57 @@ def _generate_checksum(file_path: str, relative_path: str) -> Dict:
     def _process_plugin_dir(
         self,
         plugin_type: str,
-        filenames: List,
-        file_checksums: Dict,
-        dirpath: str,
+        filenames: Generator[Path, None, None],
```
webknjaz, line 118: Is it wise to only accept generators and not generic iterables here? The code below only relies on the iteration protocol so why limit this to a very narrow iterator subtype?

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -116,62 +115,57 @@ def _generate_checksum(file_path: str, relative_path: str) -> Dict:
...
         """Process the contents of an <...>/ansible_collections/ directory.
 
         :param directory: The path to collections directory to walk and load
         """
-        for directory_path in glob(f"{directory}/*/*/"):
-            manifest_file = f"{directory_path}/MANIFEST.json"
-            galaxy_file = f"{directory_path}/galaxy.yml"
+        for directory_path in directory.glob("*/*/"):
+            manifest_file = Path(directory_path, "MANIFEST.json")
+            galaxy_file = Path(directory_path, "galaxy.yml")
```
webknjaz, line 148: `directory_path` is already a `Path` object, no need to reconstruct it:
```suggestion
            manifest_file = directory_path / "MANIFEST.json"
            galaxy_file = directory_path / "galaxy.yml"
```

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -222,8 +216,8 @@ def process_directories(self) -> Tuple[Dict, List]:
         :returns: All collections found and any errors
         """
         for directory in self._directories:
```
webknjaz, line 218: Should the class initializer accept an iterable of `Path` instead of strings for `self._directories`?

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -222,8 +216,8 @@ def process_directories(self) -> Tuple[Dict, List]:
...
-            collection_directory = f"{directory}/ansible_collections"
-            if os.path.exists(collection_directory):
+            collection_directory = Path(directory, "ansible_collections")
```
webknjaz, line 219: With the comment above, this would become:
```suggestion
            collection_directory = directory / "ansible_collections"
```

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -250,7 +244,7 @@ def worker(pending_queue: multiprocessing.Queue, completed_queue: multiprocessin
 
         try:
             (doc, examples, returndocs, metadata) = get_docstring(
```
webknjaz, line 246: Why doesn't this function accept pathlib objects? Even if we want strings for some reason, I'd move `str(path)` inside of the function.

relrod: @webknjaz note that this function is from ansible/ansible (`ansible.utils.plugin_docs.get_docstring`).  We could change it there, but it's not just a simple function update in this same PR.

share/ansible_navigator/utils/catalog_collections.py
cidrblock:
```
import itertools

numbers = (i for i in range(0, 10))

numbers_1, numbers_2 = itertools.tee(numbers, 2)

assert list(numbers_1) == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
assert list(numbers_1) == []

assert list(numbers_2) == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
assert list(numbers_2) == []
```

that's neat :)

share/ansible_navigator/utils/catalog_collections.py
cidrblock: Cool, although odd.

```
>>> directory_path = pathlib.Path("/home/bthornto")
>>> directory_path.__truediv__("file.txt")
PosixPath('/home/bthornto/file.txt')
```

share/ansible_navigator/utils/catalog_collections.py
webknjaz: Dunno. I think the rule of thumb is to have a specific return type but broad argument types. But I won't insist. Resolving.

share/ansible_navigator/utils/catalog_collections.py
webknjaz: Well, that's one of the main selling points of the pathlib interface. It's supposed to be the canonical way of using it.

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -115,62 +114,57 @@ def _generate_checksum(file_path: str, relative_path: str) -> Dict:
...
         :param collection: The details of the collection
         """
         for filename in filenames:
-            file_path = f"{dirpath}/{filename}"
-            relative_path = file_path.replace(collection["path"], "")
-            _basename, extension = os.path.splitext(filename)
-            if not filename.startswith("__") and extension == ".py":
-                checksum_dict = file_checksums.get(relative_path)
+            relative_path = Path(filename).relative_to(collection["path"])
+            if not str(filename).startswith("__"):
```
relrod, line 130: Readability nit, invert this conditional and continue early

```python
if str(filename).startswith("__"):
    continue
```

Then the rest of the loop body can be outdented one level.

cidrblock:
This PR is being closed.

relrod (approved):

PR #909, opened 2022-02-09
Replace `yahmool` in tests, update fixtures.

cidrblock:
This repalce the bogus file extsion `.yahmool` with `.os2` in the tests.

Fixture updates were required as well.

When OS/2 is supported, I hope to be long gone.

### DIFF
--- a/tests/fixtures/integration/actions/config/ansible.cfg
+++ b/tests/fixtures/integration/actions/config/ansible.cfg
@@ -1,2 +1,2 @@
 [defaults]
-yaml_valid_extensions = .yahmool
+yaml_valid_extensions = .os2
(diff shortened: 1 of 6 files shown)


ssbarnea (commented):
If I understand correctly the purpose of this is to check if we are able to load YAML files non standard extensions.

Why replacing a confusing extension with another confusing extension? Wasn't `.yaml-too` a good enough self-explanatory alternative?

cidrblock:
> If I understand correctly the purpose of this is to check if we are able to load YAML files non standard extensions.
> 
> Why replacing a confusing extension with another confusing extension? Wasn't `.yaml-too` a good enough self-explanatory alternative?

The test is simply ensuring we can use a custom ansible.cfg file with navigator and the output of ansible-config as shown in `:config` reflect the entries in that ansible.cfg file.  The extension is never used, it is only str matched against the output on the screen.

ssbarnea (approved):

PR #910, opened 2022-02-09
Remove formatter from the checks

webknjaz:
This is necessary because it was never accepted by all DevTools
contributors and imposing it on unsuspecting victims is not fair.
The contribution guide says that it's not enforced but it's not
entirely correct -- it is not possible to contribute without having
to reproduce the same result as black emits.

This is undesirable because it puts extra burden on contributors who
want to see their code structured in a meaningful way. While black
helps with the style consistency, it's not the only way of achieving
it. And in its current state, it sometimes makes formatting changes
that don't make sense and work against readability. Unfortunately,
because of it's repressive nature, it's not possible to fix it while
having it enabled.

Black also hides a number of useful linting violations that, when
actually shown, can help the coders identify structural/architectural
problems.

As for alternatives, it's possible to maintain consistent code style
in a friendlier way by combining several other smaller formatters
(like autopep8, some configurations of yapf and pre-commit.com hooks)
that are as acceptable PEP 8-wise but give a bit more flexibility in
preserving structural readability of contributions. This is not to
say that people are now allowed to nitpick style in reviews but it
should be noted that sometimes important structural improvement
requests get mistaken for style change. When getting reviews, the
motivation of improving structure should be assumed, if the
suggested code snippets contains style updates, it can be disgregarded
for as long as the underlying idea is well understood and considered
by the PR submitter.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -14,14 +14,6 @@ maintain it and you contribute.
 
 4. isort (sorts import statements)
 
-```{note}
-In early development cycles, a decision was made to use black as a formatter
-which is why current pull-requests are required to pass a
-`black --diff` check of the source tree. The decision to use `black` is
-left to individual developers as the formatting changes it makes can be
-achieved without it.
-```
-
 Details can be found below on how to run these manually, our CI will also
 check them for you.
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,16 +17,6 @@ repos:
       - id: isort
         name: Sort import statements using isort
 
-  - repo: https://github.com/psf/black.git
-    rev: 22.1.0
-    hooks:
-      - id: black
-        language_version: python3
-        args:
-          - --diff
-          - --check
-          - .
-
   - repo: https://github.com/Lucas-C/pre-commit-hooks.git
     rev: v1.1.11
     hooks:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,9 +46,6 @@
     showcontent = true
 
 
-[tool.black]
-line-length = 100
-
 [tool.coverage.run]
     branch = true
     source_pkgs = ["ansible_navigator"]
@@ -63,7 +60,7 @@ known_first_party = "ansible_navigator, key_value_store" # No effect at implemen
 lines_after_imports = 2 # Ensures consistency for cases when there's variable vs function/class definitions after imports
 lines_between_types = 1 # Separate import/from with 1 line, minimizing changes at time of implementation
 no_lines_before = "LOCALFOLDER"  # Keeps local imports bundled with first-party
-profile = "black" # Avoid conflict with black
+profile = "black"
 skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to ansible-test sanity ruleset
 
 [tool.pylint]


cidrblock:
This is a good conversation to have, thank you for taking the time to raise this.

At this time I'm -1 here as I would prefer to address this at a later date.  There has been a tremendous amount of CI work, linters, formatters, test changes and clean-up in an effort to find stability and consistency in the repo and I would prefer to see everything just settle down for a while.

This is not a + or - for or against black, it is a -1 for changes to the that would change the current workflow to it's contributors given complexity of what is currently in place.

relrod:
I've been -1 on black since the very beginning of navigator, for many of the reasons @webknjaz outlined in the summary of this PR. Many times, black serves to get in the way of understanding. Often times when I format code in a particular way, it's because it makes it easier to see at a glance what is happening, either because the situation calls for it, or because it's conventional and most developers would format things in a particular way. Black undoes that and does not contextualize its changes. It makes them en masse, without regard for _why_ the code was formatted the way it was. When you do that, you lose a _lot_ of metainformation about the code you're reading.

Further, I often found myself formatting things in _unconventional_ ways to try to get black to approximate how I want something to look. An example of this was injecting random commas (before we started using them everywhere) to get black to format argument lists and regular containers a certain way (one entry per line) in an attempt to, among either things, make diffs easier to read.

It is my opinion that because of the above, black serves to harm - not help - the readability and maintainability of the code. I was -1 on black at the start (but was outvoted), and I'm still -1 on it now.

I'm +1 to getting rid of it.

I don't think this will hamper or impede workflows, it is strictly a reduction of workflow to have to care about. Which, with all of the other linters and tools we have enabled right now, is purely a win.

ganeshrn:
> I've been -1 on black since the very beginning of navigator, for many of the reasons @webknjaz outlined in the summary of this PR. Many times, black serves to [get in the way](https://github.com/ansible/ansible-navigator/blob/a2ca04202f3e136fb3b484aeb6d683b55d44a22a/tests/unit/configuration_subsystem/utils.py#L53-L55) of understanding. 

I would argue that three levels of dereferencing  in a single line in itself is not readable code, you can't blame black for making it unreadable :-) 

Often times when I format code in a particular way, it's because it makes it easier to see at a glance what is happening, either because the situation calls for it, or because it's conventional and most developers would format things in a particular way. Black undoes that and does not contextualize its changes. It makes them en masse, without regard for _why_ the code was formatted the way it was. When you do that, you lose a _lot_ of metainformation about the code you're reading.
> 
> Further, I often found myself formatting things in _unconventional_ ways to try to get black to approximate how I want something to look. An example of this was injecting random commas (before we started using them everywhere) to get black to format argument lists and regular containers a certain way (one entry per line) in an attempt to, among either things, make diffs easier to read.
> 

In addition to making code readable for author, it is equally important to make code readable for other contributors to the project. Code readability is relative and varies from person to person. Having uniformity in the way code is structured across different sub-systems of navigator (which might be implemented by different contributors) does improve overall readability and black allows that uniformity.  If most of the active contributors to the project are in favour of adding black and think that it doesn't hamper code readability I don’t see a reason why it should be removed.

I am -1 to remove black.

P.S: This is my last comment on the thread as I don't want to spend any more of my time, focus and energy on code formatting discussions.

cidrblock:
Converting to draft, priyam is on PTO, we'll make sure he can see this and his voice is heard when he is back

cidrblock:
Title changed.

priyamsahoo:
As a developer, it has always been a hard time formating the code to make it look ** pretty ** for me. And I sometimes spend more time formatting the code properly (since formatting certain part of the code sometimes doesn't make sense), than spending time to actually write it.

I am always in favour of having something to automate this stuff but at the same time, I do have bitter memories of working with black. That is the reason I am **voting +1** for removing it and also to see how a combination of other tools (as @webknjaz suggested) would make it better for both, the devs and the code readers.

cidrblock:
Since there is no consensus here, closing this.

It would seem the next step would be for an alternative set of formatters be presented or black adopted, or things left as is.

webknjaz:
> It would seem the next step would be for an alternative set of formatters be presented

Already done via https://github.com/ansible/ansible-navigator/pull/855.

> or black adopted, or things left as is.

Which is already forced on people without giving anybody a chance to have an adoption discussion :man_shrugging:. I'm so sad that almost nobody is open to a productive dialogue and is either shutting down the conversations or is silently voting without explanation, turning a blind eye on the issues without caring to understand and/or acknowledge the problems.

It's funny that Ganesh's post actually demonstrates some of the problems I've been pointing out claiming that they are unrelated. Maybe Rick's example wasn't explicit enough but it clearly shows the result of a formatted influence on the attitude towards ignoring the linting alerts that could expose structural problems: the goal of using linters is to point out possible problems but formatting takes that away in some (very specific) cases putting this burden back on humans so basically it is a contributor's responsibility now to see such problems and correct them instead of merging things right away. This is not what happened in the example, though, which is rather alarming. If a linter had a chance to point at the issue, it'd unburden the reviewers and would hint the author to correct the structure.

PR #912, opened 2022-02-09
Spelling and variable name improvements in several files

cidrblock:
- Changes to several file to use a more descriptive variable name
- ignore sentinals in one file so it doesn't get reintroduced
- correct kegexs (plural)


src/ansible_navigator/ui_framework/form_utils.py
```diff
@@ -96,8 +96,8 @@ def form_to_dict(form: Form, key_on_name: bool = False) -> Dict:
             res["fields"][idx]["response"] = copy.copy(field.response)
             res["fields"][idx]["value"] = copy.copy(field.value)
         elif isinstance(field, (FieldChecks, FieldRadio)):
-            for oidx, option in enumerate(field.options):
-                res["fields"][idx]["options"][oidx]["checked"] = option.checked
+            for option_idx, option in enumerate(field.options):
+                res["fields"][idx]["options"][option_idx]["checked"] = option.checked
```
ganeshrn, line 100: This line can be simplified. Instead of 4 level dereferencing in single line it can in multiple lines.
Maybe something like:
```
selected_option =  res["fields"][idx]["options"] 
selected_option[option_idx]["checked"] = option.checked 
```

cidrblock: great suggestion, TY

ganeshrn (approved):

PR #916, opened 2022-02-10
Reduce lint noise (duplicate checks)

cidrblock:
Adds exclusion for those flake8 checks that overlap with pylint


webknjaz:
It would be better to add ignores on the pylint side because it's known to be slower in general and it could result in a higher performance/responsiveness benefit.

PR #918, opened 2022-02-10
Replace dyct with dictionary in one test file

cidrblock:
1 function in tests.


ssbarnea (approved):
I would have prefered to also see types added to this function if you also edit it, so we have less work do to later.

cidrblock:
> I would have prefered to also see types added to this function if you also edit it, so we have less work do to later.

funny, I almost did and update the docstring....... but chose not to

PR #936, opened 2022-02-11
Add spell checking (cspell) to pre-commit (as optional)

cidrblock:
This is not required for anyone to use.

but it works really well with https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

This adds:
- new precommit entry for cspell
- Add a cspell dictionary
- Add a cspell configuration file

To use it (in addition to the vscode plugin)
```
pre-commit run cspell --hook-stage manual --all-files
```

Future PR(s) will give better access to it via tox without running the entirety of lint-vetting, which can be rather noisy.

(will need to keep an eye out for cspell releases as all the PRs aren't in the dictionaries yet)


ssbarnea:
TBH, i would prefer a tox env called just `lint2` as it would be easier to type an a good way to indicate that the goal is for next/future lint. Other names are either too long to type or not clearly indicating the purpose.

I would propose switching to `lint2` in a different change or not touching the name of the job for now.

Adding spellcheck to the current list of vetting is perfectly fine, does not need many explanations AFAIK. In fact I am so open that I would not mind making it directly an active check (lint).

cidrblock:
> TBH, i would prefer a tox env called just `lint2` as it would be easier to type an a good way to indicate that the goal is for next/future lint. Other names are either too long to type or not clearly indicating the purpose.
> 
> I would propose switching to `lint2` in a different change or not touching the name of the job for now.
> 
> Adding spellcheck to the current list of vetting is perfectly fine, does not need many explanations AFAIK. In fact I am so open that I would not mind making it directly an active check (lint).

Thanks @ssbarnea 

The PR just adds it, no modification of tox, GH etc....

ssbarnea (approved):

relrod (approved):

PR #939, opened 2022-02-12
Add types to `colorize.py`

cidrblock:
Noticed during #936.

The file was poorly typed and a dictionary was used as a return value in many cases to the UI.

- Replaced the dictionary with a simple namespace
- Eliminated the use of `type: Any` by rescoping the to_list function
- Added, corrected types throughout the file
- Updated docstrings to reflect changes and correct errors
- Broke out ANSI functionality into it's own function because the return value was so different
- Update tests as needed

Minimal changes to `ui.py` only as required:
- Reflect the SimpleLinePart being returned vs. the dictionaries
- Call the attributes of the namespace vs. the dictionary keys
- Redirect the ANSI call to the new function

Additionally removed flake8 entries that were now incorrect.


src/ansible_navigator/ui_framework/colorize.py
```diff
@@ -32,32 +39,33 @@
...
 
 class ColorSchema:
-    """Simple holder for the schema (theme)"""
+    """A storage mechanism for the schema (theme)."""
 
     # pylint: disable=too-few-public-methods
 
-    def __init__(self, schema):
-        """start
+    def __init__(self, schema: dict):
```
ssbarnea, line 46: This is not the correct type for dictionaries. It should be something like `Dict[str, str]` or whatever makes sense for our specific case. Do not use `dict` class as a type, using `Dict`.

cidrblock: in this case the schema is a theme loaded from a file, so it would be really tricky to go deeper than one layer....  But based on the only theme we have and it's structure, I think this will work,

If and when we ever allow a user specified color theme, the error checking will need to improved...... alot.  Maybe even with a json schema

src/ansible_navigator/ui_framework/curses_defs.py
```diff
@@ -22,3 +23,18 @@ class CursesLinePart(NamedTuple):
 
 CursesLine = Tuple[CursesLinePart, ...]
 CursesLines = Tuple[CursesLine, ...]
+
+
+RgbTuple = Tuple[int, ...]
+
+
+class SimpleLinePart(SimpleNamespace):
```
relrod, line 31: I think it's better to do this as a `NamedTuple` or dataclass. I'm not able to find any examples (outside of navigator) of subclassing `SimpleNamespace` and using it in this way. I think doing it this way averts all type safety:

```
[/tmp] $ cat f.py
#!/usr/bin/env python3

from types import SimpleNamespace

class Foo(SimpleNamespace):
    bar: int

hello = Foo(bar="not an int")
print(hello.bar)

[/tmp] $ mypy f.py
Success: no issues found in 1 source file
```

whereas with `NamedTuple`:

```
[/tmp] $ cat f.py
#!/usr/bin/env python3

from typing import NamedTuple

class Foo(NamedTuple):
    bar: int

hello = Foo(bar="not an int")
print(hello.bar)

[/tmp] $ mypy f.py
f.py:8: error: Argument "bar" to "Foo" has incompatible type "str"; expected "int"
Found 1 error in 1 file (checked 1 source file)
```

I think subclassing `SimpleNamespace` and introducing fields like this doesn't actually influence the behavior (or type safety) of a `SimpleNamespace`.

src/ansible_navigator/ui_framework/curses_defs.py
```diff
@@ -22,3 +23,18 @@ class CursesLinePart(NamedTuple):
 
 CursesLine = Tuple[CursesLinePart, ...]
 CursesLines = Tuple[CursesLine, ...]
+
+
+RgbTuple = Tuple[int, ...]
```
relrod, line 28: Can it really be arbitrarily many `int`s? Isn't it just a 3-tuple?

src/ansible_navigator/ui_framework/curses_defs.py
cidrblock: good catch, a dataclass is a better pick here, since we need it to be mutable...the columns are determined after the line is built right now

src/ansible_navigator/ui_framework/curses_defs.py
cidrblock: have a look....

ssbarnea (approved):
approving bit disappointed that this did not cover for transparency or HDR.

PR #940, opened 2022-02-12
Add spell checking (cspell) to pre-commit (as required)

ssbarnea:
This activates cspell directly in pre-commit, making it much easier to use than the alternative approach. I propose to take the hit on this one and see exactly how well it works for us. We can always disable the check if it proves not useful.

Closes: #936


cidrblock (approved):
Thx Sorin.  I didn't expect to see this, but I think we should try it out.

I've seen the benefits across along list of PRs fixing mistakes and improvements to made-up obscure variable names.  (99.9% of which I'm sure I added, I have an issue with spelling)

That said, a couple other approvals from the team would be great.  I would especially like @relrod to voice his opinion here, as his perspective is always good.

ganeshrn (approved):

priyamsahoo (approved):
Approving this but at the same time I want to share a weird fact that I like to see my own spelling mistakes in the code that I made in the past :)

Issue #620, opened 2021-11-18
RFE: subcommand to invoke ansible-vault


sugitk:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

The customer requests to invoke ansible-vault command with EE.

cidrblock:
@sugitk Ad-hoc command support will be coming soon via https://github.com/ansible/ansible-navigator/pull/624  

let me know fi this might work in this case,

thanks- Brad

sugitk:
@cidrblock
I haven't tested the PR yet though, I guess it will meet the requirement if it is implemented.  Thank you :)

cidrblock:
Related: https://github.com/ansible/ansible-navigator/pull/941

PR #941, opened 2022-02-13
Switch exec command to a postional

cidrblock:
As suggested by @webknjaz, this switches the command for `exec` at the command line to be a positional

Here are all the variations supported now
```
VAULT_COMMAND = (
    "ansible-vault encrypt_string --vault-password-file a_password_file 'foobar'"
    " --name 'the_secret'"
)

BASE_COMMAND = "ansible-navigator --ll debug --la false"

variations = (
    f"{BASE_COMMAND} --mode stdout exec {VAULT_COMMAND}",
    f"{BASE_COMMAND} --mode stdout exec --exec-shell false {VAULT_COMMAND}",
    f"{BASE_COMMAND} --mode stdout exec -- {VAULT_COMMAND}",
    f"{BASE_COMMAND} --mode stdout exec --exec-shell false -- {VAULT_COMMAND}",
    f'{BASE_COMMAND} exec "{VAULT_COMMAND}" --mode stdout',
    f'{BASE_COMMAND} exec "{VAULT_COMMAND}" --mode stdout --exec-shell false',
)
```
(the way this works is we use the "known args", the rest get passed to the EE, the `--` tells argparse to stop processing. So variations 0 and 1 work because the args passed to vault don't overlap with our own.)  So we can skip those when we document this a little better in the FAQ)


cidrblock:
Specifically looking for @webknjaz and/or @relrod review here

webknjaz:
My suggestion was somewhat different. I meant this:
```
ansible-navigator --mode stdout exec - - ansible-vault encrypt_string --vault-password-file a_password_file 'foobar' --name 'the_secret'
```

cidrblock:
> ```
> ansible-navigator --mode stdout exec - - ansible-vault encrypt_string --vault-password-file a_password_file 'foobar' --name 'the_secret'
> ```

I'll look into that.............. thanks

src/ansible_navigator/actions/exec.py
```diff
@@ -16,23 +17,43 @@
...
-    pass_through_args = None
+    logger.debug("exec_command: %s", exec_command)
+    logger.debug("exec_shell: %s", exec_shell)
+    logger.debug("extra_args: %s", extra_args)
     if exec_shell and exec_command:
         command = "/bin/bash"
-        pass_through_args = ["-c", exec_command]
+        # Determine if any extra args were picked up
+        if isinstance(extra_args, list):
+            pass_command = exec_command + " " + " ".join(extra_args)
```
webknjaz, line 42: Would this be better
```suggestion
            pass_command = " ".join((exec_command, *extra_args))
```
?

src/ansible_navigator/actions/exec.py
```diff
@@ -16,23 +17,43 @@
...
+        else:
+            pass_command = exec_command
```
webknjaz, line 44: Looking closer, it seems like we can make this a bit more flat:
```suggestion
        _extra_args = extra_args if isinstance(extra_args, list) else ()
        pass_command = " ".join((exec_command, *_extra_args))
```

src/ansible_navigator/actions/exec.py
cidrblock: way better.............. I didn't like it when I typed it.  ty

tests/integration/actions/exec/test_stdout_vault.py
```diff
@@ -0,0 +1,64 @@
...
+class ShellCommand(UiTestStep):
+    """A shell command."""
+
+    search_within_response = SearchFor.PROMPT
+
+
+VAULT_COMMAND = (
+    "ansible-vault encrypt_string --vault-password-file a_password_file 'foobar'"
+    " --name 'the_secret'"
+)
```
webknjaz, line 26: Why not use something like `echo` or similar/wrapper? I'd prefer an executable with low overhead for the test to be faster, plus running ansible internals is not essential for this test.

src/ansible_navigator/actions/exec.py
webknjaz: There's another, a little better, suggestion. Did you miss it?

webknjaz: Ah, I see you applied it already :)

tests/integration/actions/exec/test_stdout_vault.py
webknjaz: As an example, you could do something like printing the args back and then just compare the output in tests here. You could even make those assertions more precise:
```suggestion
VAULT_COMMAND = (
    "python -c 'import sys; print(sys.argv)' -- ansible-vault encrypt_string "
    "--vault-password-file a_password_file 'foobar' --name 'the_secret'"
)
```

cidrblock: The use of vault here was intentional, this will be _the_ documented approach for both galaxy and vault until such time navigator has entry points(if ever). I like the tests proving a suggested use case. Related: https://github.com/ansible/ansible-navigator/issues/620

cidrblock: that said........... that second example is really cool.  I may add another test file, with just that so as issues come in, if they do, there is a sane place to put them.

webknjaz: In that case, consider using [pytest-bdd](https://github.com/pytest-dev/pytest-bdd) (in the future) for documented behaviors.

cidrblock: will pass on the second suggestion here, only because asserting against the output of vault proves the params were passed all the way through to it

cidrblock:
BTW @webknjaz I love the `--` idea here................TY  We never really had a good suggestion for how to differentiate between navigator params and params that get passed to the underlying command, other than just dump them all on the command line.  Although we'll always try to not overlap with the underlying commands we support, this as a "recommended" way to do it would prevent breakage if that ever becomes impossible to maintain.

webknjaz (approved):

cidrblock:
This PR is being closed.

PR #943, opened 2022-02-13
Use full URL for documentation links

cidrblock:
Less room for misinterpretation.


webknjaz:
The main motivation for using short names was that they were less likely to exceed line length limits + they were easier to type manually.

cidrblock:
JSYK, I've already updated the URL in the "About" for the repo.

ssbarnea (approved):

PR #948, opened 2022-02-14
Notification in `run`: Preparing for automation, please hold on

### LINKED ISSUE #303: Add notification message in TUI mode for run action
In TUI with run action, the screen goes blank especially while running with EE enabled as in the background it takes some time to initialize the EE. While waiting send notification message to UI similar to collection action. (For example `Collecting collection content, this may take a minute the first time...` )

cidrblock:
Fixes #303 

This can last a second or 2, or less? depending on the development system and it's speed.

Users will see this until the first message is received from runner.

image

Tested locally with and with out an EE.  Intentionally not added to the tests because of risk of it being missed on the screen.

### DIFF
--- a/src/ansible_navigator/actions/run.py
+++ b/src/ansible_navigator/actions/run.py
@@ -284,7 +284,7 @@ class Action(App):
 
         # Show a notification until the first the first message from the queue is processed
         if self._subaction_type == "run":
-            messages = ["Preparing for automation, please hold on..."]
+            messages = ["Preparing for automation, please wait..."]
             notification = nonblocking_notification(messages=messages)
             interaction.ui.show(notification)
             while not self._first_message_received:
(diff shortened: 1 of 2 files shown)


src/ansible_navigator/actions/run.py
```diff
@@ -279,6 +282,14 @@ def run(self, interaction: Interaction, app: AppPublic) -> Union[Interaction, No
 
         self.steps.append(self._plays)
 
+        # Show a notification until the first the first message from the queue is processed
+        if self._subaction_type == "run":
+            messages = ["Preparing for automation, please hold on..."]
```
relrod, line 287: Nit: probably better to keep idioms out of user-facing strings, especially since we don't currently do any translations.

```suggestion
            messages = ["Preparing for automation, please wait..."]
```

relrod (approved):
One small inline suggestion. Approved assuming you will pull that in.

Issue #209, opened 2021-05-07
Consider renaming `app` to better reflect its use, varied uses 


cidrblock:
might be broad, but better now vs. later

PR #949, opened 2022-02-14
Rename App to ActionBase

cidrblock:
Partial #209 

This renames the `App` class to `ActionBase` as it better describes it's use.


relrod (approved):
lgtm!

PR #950, opened 2022-02-14
Add EE support status (enabled/disabled) to `:images` list output

### LINKED ISSUE #415: :images main page should indicate whether execution environment use is enabled
On this page there is not the information about the EE enabled.

there is an information about the **main EE** used on the `ansible-navigator config` but if this EE is unabled we don't know.

image

config example:
```
ansible-navigator:
#   app: run
#   collection-doc-cache-path: /tmp/cache.db
#   cmdline: "--forks 15"
#   editor:
#     command: vim_from_setting
#     console: False
#   documentation:
#     plugin:
#       name: shell
#       type: become
  execution-environment:
    image: "quay.io/ansible/awx-ee:devel"
    enabled: False

  log:
    level: debug
```

cidrblock:
Fixes #415 

This is what it looks like now:

image

Tests updated to confirm

### DIFF
--- a/src/ansible_navigator/actions/images.py
+++ b/src/ansible_navigator/actions/images.py
@@ -368,8 +368,10 @@ class Action(App):
             image["__full_name"] = f"{image['repository']}:{image['tag']}"
             image["__name"] = image["name"]
             if image["__full_name"] == self._args.execution_environment_image:
-                image["__name"] += " (primary)"
-                image["__name_tag"] += " (primary)"
+                ee_support = "enabled" if self._args.execution_environment else "disabled"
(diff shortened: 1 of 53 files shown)


relrod:
To me, this reads as though somehow that EE image, itself, is disabled somehow (like via docker/podman). i.e. it looks like the enabled/disabled is attached to this particular image when I think what you're trying to express here is that our ability to run it (via navigator) is disabled in settings.

Not sure if a better way to word/express/visualize that.

cidrblock:
With `:settings` coming, this doesn't seem necessary and might lead to confusion...... TY @relrod and Colin (@cloin)

Issue #951, opened 2022-02-14
`pull-policy` should be a dictionary in the settings (`pull: policy: ""`)


cidrblock:
Just like pull_arguments, see PR below

PR #952, opened 2022-02-14
Add pull_arguments

### LINKED ISSUE #505: Add an option to disable `tls-verify` for the EE pull.
This should default to True, as verification would be the desired behavior.

maybe something like:

```
ansible-navigator:
  execution-environment:
    tls-verification:
```

### LINKED ISSUE #632: Ability to specify the podman --cert-dir option in config
##### ISSUE TYPE
 - Add ability to specify additional podman command-line options in ansible-navigator.yml

##### SUMMARY
Our corporate environment uses a 3rd party SSL inspection tool that re-signs all traffic with the vendor's root certificate.  This obviously breaks things when trying to pull from, say, quay.io and the cert is signed by someone else.  It appears there is no good way to put the vendor's cert in a standard location that podman will search (e.g., /etc/containers/certs.d) (see https://github.com/containers/podman/issues/2726).  As such, I have to resort to passing the `--certs-dir` option to podman on the command-line each time I need to login to a container registry and pull from it.

However, if you set the pull-policy in your ansible-navigator.yml config, it will fail to pull the image on launch of `ansible-navigator` because it is not passing that `--certs-dir` option.  If podman can search known directories and use the certificates therein, this is a moot point, but I cannot manage to get that working.

cidrblock:
Fixes #505 
Fixes #632 

This add a `--pull-arguments` command line parameter which will pass additional args to the pull policy command.

- json schema updates WIP: https://github.com/ansible-community/schemas/issues/150
- test updates
- shortened `pull_policy` to `policy` in puller to match incoming `arguments`

This turns `pull` into a dictionary, `pull-policy` will need to be changes, and will cause a major release

See: https://github.com/ansible/ansible-navigator/issues/951

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -480,7 +480,7 @@ NavigatorConfiguration = ApplicationConfiguration(
             short_description=(
                 "Specify any additional parameters that should be added to the"
                 " pull command when pulling an execution environment from a container"
-                " registry. e.g. --pa \"'--tls-verify=false'\""
+                " registry. e.g. --pa='--tls-verify=false'"
             ),
             value=SettingsEntryValue(),
         ),
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -478,7 +478,7 @@ NavigatorConfiguration = ApplicationConfiguration(
             cli_parameters=CliParameters(action="append", nargs="+", short="--pa"),
             settings_file_path_override="execution-environment.pull.arguments",
             short_description=(
-                "Specify any additional parameters that should be appended to the"
+                "Specify any additional parameters that should be added to the"
                 " pull command when pulling an execution environment from a container"
                 " registry. e.g. --pa \"'--tls-verify=false'\""
             ),
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -474,6 +474,17 @@ NavigatorConfiguration = ApplicationConfiguration(
             value=SettingsEntryValue(default="module"),
         ),
         SettingsEntry(
+            name="pull_arguments",
+            cli_parameters=CliParameters(action="append", nargs="+", short="--pa"),
+            settings_file_path_override="execution-environment.pull.arguments",
+            short_description=(
+                "Specify any additional parameters that should be appended to the"
+                " pull command when pulling an execution environment from a container"
+                " registry. e.g. --pa \"'--tls-verify=false'\""
+            ),
+            value=SettingsEntryValue(),
+        ),
+        SettingsEntry(
             name="pull_policy",
             choices=["always", "missing", "never", "tag"],
             cli_parameters=CliParameters(short="--pp"),
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -850,7 +850,7 @@ class NavigatorPostProcessor:
         config: ApplicationConfiguration,
     ) -> PostProcessorReturn:
         # pylint: disable=unused-argument
-        """Post process ```pull_arguments```
+        """Post process ``pull_arguments``
 
         :param entry: The current settings entry
         :param config: The full application configuration
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -845,6 +845,27 @@ class NavigatorPostProcessor:
 
     @staticmethod
     @_post_processor
+    def pull_arguments(
+        entry: SettingsEntry,
+        config: ApplicationConfiguration,
+    ) -> PostProcessorReturn:
+        # pylint: disable=unused-argument
+        """Post process ```pull_arguments```
+
+        :param entry: The current settings entry
+        :param config: The full application configuration
+        :return: An instance of the standard post process return object
+        """
+        messages: List[LogMessage] = []
+        exit_messages: List[ExitMessage] = []
+        if entry.value.source == C.ENVIRONMENT_VARIABLE:
+            entry.value.current = to_list(entry.value.current)
+        if entry.value.current is not C.NOT_SET:
+            entry.value.current = flatten_list(entry.value.current)
+        return messages, exit_messages
+
+    @staticmethod
+    @_post_processor
     def set_environment_variable(
         entry: SettingsEntry,
         config: ApplicationConfiguration,
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -132,7 +132,7 @@ class ImagePuller:
         messages = [("Execution environment image name:", self._image)]
         messages.append(("Execution environment image tag:", self._image_tag))
         if isinstance(self._arguments, list):
-            arguments = shlex_join(arguments)
+            arguments = shlex_join(self._arguments)
         else:
             arguments = "None"
         messages.append(("Execution environment pull arguments:", arguments))
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -143,7 +143,7 @@ class ImagePuller:
         """print a little value added information"""
         messages = [("Execution environment image name:", self._image)]
         messages.append(("Execution environment image tag:", self._image_tag))
-        arguments = shlex_join(self._arguments) or "None"
+        arguments = shlex_join(self._arguments) or None
         messages.append(("Execution environment pull arguments:", arguments))
         messages.append(("Execution environment pull policy:", self._pull_policy))
         messages.append(("Execution environment pull needed:", self._pull_required))
@@ -154,8 +154,8 @@ class ImagePuller:
         print("\u002d" * width)
         column_width = max((len(m[0]) for m in messages))
         for msg, value in messages:
-            print(f"{msg.ljust(column_width)} {value}")
-            self._log_message(message=f"{msg}: {value}", level=logging.INFO)
+            print(f"{msg.ljust(column_width)} {value!s}")
+            self._log_message(message=f"{msg!s}: {value!s}", level=logging.INFO)
         print("\u002d" * width)
         print("Updating the execution environment")
         print("\u002d" * width)
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -32,11 +32,11 @@ class ImagePuller:
         :param arguments: Additional arguments to be appended to the pull policy
         :param pull_policy: The current pull policy from the settings
         """
-        self._arguments: List = arguments
+        self._arguments: List[str] = arguments
         self._assessment = ImageAssessment
-        self._container_engine = container_engine
+        self._container_engine: str = container_engine
         self._exit_messages: List[ExitMessage] = []
-        self._image = image
+        self._image: str = image
         self._image_present: bool
         self._image_tag: str
         self._logger = logging.getLogger(__name__)
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -5,7 +5,9 @@ import subprocess
 
 from types import SimpleNamespace
 from typing import List
+from typing import Union
 
+from ..configuration_subsystem import Constants
 from ..utils import ExitMessage
 from ..utils import ExitPrefix
 from ..utils import LogMessage
@@ -24,7 +26,13 @@ class ImagePuller:
     # pylint: disable=too-many-instance-attributes
     """Image puller"""
 
-    def __init__(self, container_engine: str, image: str, arguments: List[str], pull_policy: str):
+    def __init__(
+        self,
+        container_engine: str,
+        image: str,
+        arguments: Union[Constants, List[str]],
+        pull_policy: str,
+    ):
         """Initialize the container image puller.
 
         :param container_engine: The name of the container engine
@@ -32,7 +40,7 @@ class ImagePuller:
         :param arguments: Additional arguments to be appended to the pull policy
         :param pull_policy: The current pull policy from the settings
         """
-        self._arguments: List[str] = arguments
+        self._arguments: Union[Constants, List[str]] = arguments
         self._assessment = ImageAssessment
         self._container_engine: str = container_engine
         self._exit_messages: List[ExitMessage] = []
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -1,5 +1,6 @@
 """image puller"""
 import logging
+import shlex
 import subprocess
 
 from types import SimpleNamespace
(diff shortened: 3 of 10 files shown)


src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -475,6 +475,17 @@ class Internals(SimpleNamespace):
...
             value=SettingsEntryValue(default="module"),
         ),
+        SettingsEntry(
+            name="pull_arguments",
+            cli_parameters=CliParameters(action="append", nargs="+", short="--pa"),
+            settings_file_path_override="execution-environment.pull.arguments",
+            short_description=(
+                "Specify any additional parameters that should be added to the"
+                " pull command when pulling an execution environment from a container"
+                " registry. e.g. --pa \"'--tls-verify=false'\""
```
relrod, line 485: Are the doubled up quotes needed here? This looks odd.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -843,6 +843,27 @@ def playbook_artifact_replay(
...
         return messages, exit_messages
 
+    @staticmethod
+    @_post_processor
+    def pull_arguments(
+        entry: SettingsEntry,
+        config: ApplicationConfiguration,
+    ) -> PostProcessorReturn:
+        # pylint: disable=unused-argument
+        """Post process ```pull_arguments```
```
relrod, line 853: Guessing this should be wrapped in 2 backticks (for ReST) instead of 3.

src/ansible_navigator/image_manager/puller.py
```diff
@@ -38,15 +41,15 @@ def __init__(self, container_engine: str, image: str, pull_policy: str):
         self._image_tag: str
         self._logger = logging.getLogger(__name__)
         self._messages: List[LogMessage] = []
-        self._pull_policy = pull_policy
+        self._policy: str = policy
```
relrod, line 44: Maybe do this change in a separate PR. It's a good rename, but it's unrelated to adding this argument.

src/ansible_navigator/image_manager/puller.py
```diff
@@ -23,13 +24,15 @@ class ImagePuller:
...
+    def __init__(self, container_engine: str, image: str, arguments: List, policy: str):
         """Initialize the container image puller.
 
         :param container_engine: The name of the container engine
         :param image: The name of the image to pull
-        :param pull_policy: The current pull policy from the settings
+        :param arguments: Additional arguments to be appended to the pull policy
+        :param policy: The current pull policy from the settings
         """
+        self._arguments: List = arguments
```
relrod, line 35: Can you give a more specific type constructor parameter here and above? Is it `List[str]`?

src/ansible_navigator/image_manager/puller.py
```diff
@@ -128,8 +131,14 @@ def prologue_stdout(self):
         """print a little value added information"""
         messages = [("Execution environment image name:", self._image)]
         messages.append(("Execution environment image tag:", self._image_tag))
-        messages.append(("Execution environment pull policy:", self._pull_policy))
+        if isinstance(self._arguments, list):
+            arguments = " ".join(shlex.quote(arg) for arg in self._arguments)
```
relrod, line 135: `shlex.join()`

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,18 +151,33 @@ def prologue_stdout(self):
         print("Updating the execution environment")
         print("\u002d" * width)
 
+    def _generate_pull_command(self) -> List:
```
relrod, line 154: Can we give the type constructor its parameter? Is it a `List[str]`?

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,18 +151,33 @@ def prologue_stdout(self):
...
+        """Generate the pull command.
+
+        :returns: The list of command parts
+        """
+        command_line = [self._container_engine, "pull"]
+        if isinstance(self._arguments, list):
```
relrod, line 160: Would this `isinstance` check and the one above go away if we defaulted the new setting to `[]`?

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,18 +151,33 @@ def prologue_stdout(self):
...
+            for argument in self._arguments:
+                command_line.extend(shlex.split(argument))
```
relrod, line 162: `+=` might be more conventional than `extend()`.

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,18 +151,33 @@ def prologue_stdout(self):
...
+
     def pull_stdout(self):
         """pull the image, print to stdout
 
         podman writes to stderr
         docker writes to stdout
         """
         try:
+            command_line = self._generate_pull_command()
+            shlex_joined = " ".join(shlex.quote(arg) for arg in command_line)
```
relrod, line 174: `shlex.join()`

relrod: Also, is this necessary? I haven't followed the logic all the way so it might be fine, but we should make sure we're not double escaping.

relrod: Ah okay it's just for rendering, not for `subprocess.run()` below. Carry on, then. :)

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,18 +151,33 @@ def prologue_stdout(self):
...
             if self._container_engine == "podman":
-                subprocess.run([self._container_engine, "pull", self._image], check=True)
+                subprocess.run(command_line, check=True)
             elif self._container_engine == "docker":
                 subprocess.run(
-                    [self._container_engine, "pull", self._image],
+                    command_line,
                     check=True,
                     stderr=subprocess.PIPE,
                 )
```
relrod, line 183: Maybe collapse this whole block into

```python
stderr_pipe = subprocess.PIPE if self.container_engine == "docker" else None
subprocess.run(command_line, check=True, stderr=stderr_pipe)
```

to reduce some duplication. It took me a minute to realize that stderr is used here only for exception processing below. That's kind of clever/evil - because at first glance, it looks like we're discarding the result of `subprocess.run()` (and therefore wouldn't care about the stderr output).

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: thanks, got it.

src/ansible_navigator/image_manager/puller.py
cidrblock: ok.

src/ansible_navigator/image_manager/puller.py
cidrblock: yes, should be.

src/ansible_navigator/image_manager/puller.py
cidrblock: it's 3.8+ only, but I'll move this to utils so it's obvious.

src/ansible_navigator/image_manager/puller.py
cidrblock: should be yes, nice.

src/ansible_navigator/image_manager/puller.py
cidrblock: It would, but a default of `[]` gets really odd when you see the default in the `--help` because at the command line we use nargs + and append, and there's not really a way to express a list object at the CLI.

src/ansible_navigator/image_manager/puller.py
cidrblock: I switched to extend a while back and I don't recall why. Might be because of `append`.  

For the sake of research some interesting differences here which I was unaware of: https://stackoverflow.com/questions/3653298/concatenating-two-lists-difference-between-and-extend

This one I'll leave as is.

src/ansible_navigator/image_manager/puller.py
cidrblock: yeah, it looks like stderr is passed to popen which has a default of None, cleaner I like.

This subtle difference between podman and docker was tricky I recall with podman not using stderr.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: researching

cidrblock: ok............  this has to do with it being recognized as a unknown arg because it has a `--` in it....

```
import argparse
import shlex

parser = argparse.ArgumentParser(exit_on_error=False)
parser.add_argument("--pass-through", nargs="+")

# No quotes
command = "--pass-through --other=5"
try:
    known, unknown = parser.parse_known_args(shlex.split(command))
except argparse.ArgumentError as exc:
    assert "argument --pass-through: expected at least one argument" in str(exc)

# single quotes
command = "--pass-through '--other=5'"
try:
    known, unknown = parser.parse_known_args(shlex.split(command))
except argparse.ArgumentError as exc:
    assert "argument --pass-through: expected at least one argument" in str(exc)

# double quotes
command = '--pass-through "--other=5"'
try:
    known, unknown = parser.parse_known_args(shlex.split(command))
except argparse.ArgumentError as exc:
    assert "argument --pass-through: expected at least one argument" in str(exc)

# nested quotes
command = "--pass-through \"'--other=5'\""
known, unknown = parser.parse_known_args(shlex.split(command))
assert known.pass_through == ["'--other=5'"]

# equal raw
command = "--pass-through=--other=5"
known, unknown = parser.parse_known_args(shlex.split(command))
assert known.pass_through == ["--other=5"]

# equal quoted
command = "--pass-through='--other=5'"
known, unknown = parser.parse_known_args(shlex.split(command))
assert known.pass_through == ["--other=5"]
```

src/ansible_navigator/image_manager/puller.py
```diff
@@ -128,8 +132,14 @@ def prologue_stdout(self):
         """print a little value added information"""
         messages = [("Execution environment image name:", self._image)]
         messages.append(("Execution environment image tag:", self._image_tag))
+        if isinstance(self._arguments, list):
```
relrod, line 135: Missed this the first time... The type above (in the constructor) says this __has__ to be a `List[str]`, so this `isinstance` should always be true, unless I'm missing something.

src/ansible_navigator/image_manager/puller.py
```diff
@@ -142,24 +152,35 @@ def prologue_stdout(self):
         print("Updating the execution environment")
         print("\u002d" * width)
 
+    def _generate_pull_command(self) -> List[str]:
+        """Generate the pull command.
+
+        :returns: The list of command parts
+        """
+        command_line = [self._container_engine, "pull"]
+        if isinstance(self._arguments, list):
```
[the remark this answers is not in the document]
cidrblock: Good catch............ It can be Constants.NOT_SET, the default value for a setting not set

src/ansible_navigator/image_manager/puller.py
relrod: That's concerning to me. If the type was wrong, then why is it type-checking? Is this a result of #956 not being merged yet?

cidrblock: No, even with that one in now, mypy can't track the type.  but I think we talked about this and have an idea of where to go with SettingsEntry....

cidrblock:
This PR is being closed.

src/ansible_navigator/image_manager/puller.py
```diff
@@ -22,22 +26,30 @@ class ImagePuller:
...
+        pull_policy: str,
+    ):
         """Initialize the container image puller.
 
         :param container_engine: The name of the container engine
         :param image: The name of the image to pull
+        :param arguments: Additional arguments to be appended to the pull policy
         :param pull_policy: The current pull policy from the settings
         """
+        self._arguments: Union[Constants, List[str]] = arguments
```
ganeshrn, line 43: The self._arguments type can be set to list here itself instead of checking it later on?
Something like
```
self._arguments: List[str]
if not isinstance(arguments, list):
  self._arguments = []
else:
  self._arguments = arguments
```
```

File: src/ansible_navigator/image_manager/puller.py (hunk 11 above)
Comment (ganeshrn) [reply src/ansible_navigator/image_manager/puller.py:142]: With above suggested change this check is not required?

File: src/ansible_navigator/configuration_subsystem/navigator_post_processor.py [hunk 14]
```diff
@@ -843,6 +843,27 @@ def playbook_artifact_replay(
...
+
+        :param entry: The current settings entry
+        :param config: The full application configuration
+        :return: An instance of the standard post process return object
+        """
+        messages: List[LogMessage] = []
+        exit_messages: List[ExitMessage] = []
+        if entry.value.source == C.ENVIRONMENT_VARIABLE:
+            entry.value.current = to_list(entry.value.current)
+        if entry.value.current is not C.NOT_SET:
```
ganeshrn, line 863: If the value is not set can it be initialized to an empty list to avoid type checking later on in the code?

src/ansible_navigator/image_manager/puller.py
```diff
@@ -141,24 +159,35 @@ def prologue_stdout(self):
...
-            elif self._container_engine == "docker":
-                subprocess.run(
-                    [self._container_engine, "pull", self._image],
-                    check=True,
-                    stderr=subprocess.PIPE,
-                )
-            else:
-                raise ValueError("Unknown container engine")
-
+            command_line = self._generate_pull_command()
```
ganeshrn, line 186: In the post processor we check if either podman or docker is installed https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py#L213

What if both are not installed, based on the logic at the above line looks like the value will be still set to `auto`?
If yes maybe we can handle the error at L160 here in post processor.

src/ansible_navigator/utils/functions.py
```diff
@@ -443,6 +444,13 @@ def round_half_up(number: Union[float, int]) -> int:
     return int(rounded)
 
 
+def shlex_join(tokens) -> str:
```
ganeshrn, line 447: Maybe good to have type information for `tokens`

ganeshrn: Also bit of explanation as to why is it handled differently based on python version might help

src/ansible_navigator/image_manager/puller.py
cidrblock: @relrod Decided to keep it extend, resolving this

src/ansible_navigator/image_manager/puller.py
cidrblock: Yeah, the type of each settings parameter is lost along the way:

```
def pull_image(args: ApplicationConfiguration):
    """pull the image if required"""
    reveal_type(args.pull_arguments)
    image_puller = ImagePuller(
        container_engine=args.container_engine,
        image=args.execution_environment_image,
        arguments=args.pull_arguments,
        pull_policy=args.pull_policy,
    )
```
```
(venv) x1 ➜  ansible-navigator git:(podman_pull) ✗ mypy
src/ansible_navigator/cli.py: note: In function "pull_image":
src/ansible_navigator/cli.py:49:17: note: Revealed type is "Any"
```

Rick is working on a refactor of the settings so the type information is preserved...

src/ansible_navigator/image_manager/puller.py
cidrblock: yeah, moving it

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: We don't currently have a good way to do this, the default would have to be set to `[]` which doesn't translate well to command line parameters, if we didn't use the default, it would be odd in the `:settings` subcommand because the current would be different from the default, would show as "changed" but we don't have a Constant to describe "why" it was changed.

src/ansible_navigator/image_manager/puller.py
cidrblock: 160 is gone now, the only logic here now is 

```
            stderr_pipe = subprocess.PIPE if self._container_engine == "docker" else None
```
So we can capture it's stderr and log it.  If we introduce a 3rd container engine, we will treat it like we do podman.

src/ansible_navigator/utils/functions.py
cidrblock: got it

ganeshrn (approved):

PR #956, opened 2022-02-15
Convert SettingsEntry to a dataclass

cidrblock:
- it's 2 child classes were changed as well
- No effort was made to further clarify the types for EntryValue as that will be a significant under taking
- Union[None,  was replace with Optional

One change in the post process for which, because choices could be a bool.  This brings attention to the possiblilty we may want different "types" of choices. e.g. bools or strings.

This should at least ensure SettingsEntry does not get misused, at least.


src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -43,43 +46,46 @@ class Constants(Enum):
...
 
-class SettingsEntryValue(SimpleNamespace):
-    # pylint: disable=too-few-public-methods
-    """An object to store a value"""
+@dataclass
+class SettingsEntryValue:
+    """An object to store a value."""
 
+    #: The default value for the entry
     default: Any = Constants.NOT_SET
```
relrod, line 54: It'd be cool to nuke these `Any`s some day. Perhaps by making `SettingsEntryValue` [polymorphic](https://docs.python.org/3/library/typing.html#generics) over a type parameter that describes what the thing could be? i.e. instead of just a `SettingsEntryValue`, we could have `SettingsEntryValue[T]` and when we use it, it would be `SettingsEntryValue[str]` or whatever the actual type of the setting is.

Internally, we could `Union` it with `Constants` (or some new place we put `Constants.NOT_SET`) so that `NOT_SET` still works.

Just food for future thought.

relrod (approved):

PR #972, opened 2022-02-17
Avoid duplicate mounts when collection doc cache is in the CWD

### LINKED ISSUE #608: Collection documentation inside EE vs local
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
Cannot pull up collection documentation of collections contained within an EE.

```
ansible-navigator 1.1.0
```

##### CONFIGURATION
ansible-navigator.yml
```yaml
---
ansible-navigator:
  ansible:
    config: $PWD/ansible.cfg
    # cmdline: "--forks 15"
    # inventories:
    # - /tmp/test_inventory.yml
    # playbook: /tmp/test_playbook.yml

  ansible-runner:
    artifact-dir: $PWD/navigator_artifacts
    rotate-artifacts-count: 10
    timeout: 300

  app: welcome  # ‘collections’, ‘config’, ‘doc’, ‘images’, ‘inventory’, ‘replay’, ‘run’ or ‘welcome’

  collection-doc-cache-path: $PWD/cache.db

  color:
    enable: True
    osc4: True

  editor:
    # command: code
    command: vim +{line_number} {filename}
    # console: False
    console: True

  # documentation:
  #   plugin:
  #     name: shell
  #     type: become

  execution-environment:
    container-engine: podman
    enabled: true
    # environment-variables:
    #   pass:
    #     - ONE
    #     - TWO
    #     - THREE
    #   set:
    #     KEY1: VALUE1
    #     KEY2: VALUE2
    #     KEY3: VALUE3
    image: localhost/kiewit_ee_all:latest
    pull-policy: never
    # volume-mounts:
    # - src: "/test1"
    #   dest: "/test1"
    #   label: "Z"
    # container-options:
    # - "--net=host"

  # help-config: True

  # help-doc: True

  # help-inventory: True

  # help-playbook: False

  # inventory-columns:
  #   - ansible_network_os
  #   - ansible_network_cli_ssh_type
  #   - ansible_connection

  logging:
    level: warning
    append: False
    file: $PWD/navigator.log

  mode: interactive  # interactive, stdout

  playbook-artifact:
    enable: True
    replay: $PWD/navigator_replay.json
    save-as: $PWD/navigator_replay.json

```
ansible.cfg (in CWD)
```ini
[defaults]
forks = 20
precedence = all_plugins_inventory,groups_plugins_inventory,all_inventory,groups_inventory,all_plugins_play,groups_plugins_play
retry_files_enabled = true
host_key_checking = false
#collections_paths = collections/:~/.ansible/collections:/usr/share/ansible/collections
#roles_path = roles/:~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles
vault_identity_list = vcenter@/home/runner/workspace/kiewit/ansible_vaults/vcenter,azure@/home/runner/workspace/kiewit/ansible_vaults/azure,ansible_user@/home/runner/workspace/kiewit/ansible_vaults/ansible_user,kiewitadmin_user@/home/runner/workspace/kiewit/ansible_vaults/kiewitadmin_user,awx@/home/runner/workspace/kiewit/ansible_vaults/awx,snow@/home/runner/workspace/kiewit/ansible_vaults/snow
ssh_args = -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
callback_whitelist = timer, profile_roles, profile_tasks, junit
gathering = smart
fact_caching = jsonfile
fact_caching_connection = $HOME/ansible/facts
fact_caching_timeout = 600
nocows = true
# Use the YAML callback plugin.
stdout_callback = yaml
# Use the stdout_callback when running ad-hoc commands.
bin_ansible_callbacks = True
```

##### LOG FILE
N/A

##### STEPS TO REPRODUCE
As shown in my configs, if I launch `ansible-navigator` with no arguments, I expect to be at the welcome screen and when I list :images, I expect to be using the specified image.  This works as expected

```shell
NAME                                                                TAG               EXECUTION ENVIRONMENT                                   CREATED                                       SIZE
0│ansible-builder                                                     latest                            False                                   14 hours ago                                  539 MB
1│awx-ee                                                              latest                             True                                   2 hours ago                                   1.27 GB
2│kiewit_ee_all (primary)                                             latest                             True                                   About an hour ago                             5.99 GB
3│kiewit_ee_linux_instance                                            latest                             True                                   2 hours ago                                   3.43 GB
4│kiewit_ee_windows_instance                                          latest                             True                                   2 hours ago                                   3.42 GB
```

Moreover, if I press a number corresponding to the selected image (in this case, 2) and then view Ansible version and collections, I can confirm that the collections specified when building the image are there.

However, when I run the :collections command from the welcome screen, I'm told there are no collections.

```shell
WARNING
                                                          ──────────────────────────────────────────────────────────────────
                                                          humph. no collections were found in the following paths:
                                                          [HINT] Try installing some or try a different execution enviroment
                                                          ──────────────────────────────────────────────────────────────────
```

##### EXPECTED RESULTS

My expectation is that I can browse collection documentation for collections pre-installed "inside" the EE, but it appears that `ansible-navigator` is looking for collections installed locally.

I'm extremely excited about the prospect of self-contained EEs as it will make my job of setting up other developers, CI/CD, and AWX/Tower to have the same environment everywhere.  And `ansible-navigator` as a method to introspect and execute from an EE is amazing.

However, from an Ansible content developmental point of view, if I'm using a provided EE and writing a playbook, I want to pull up documentation specific to the collections contained within the EE to ensure I'm using modules, libraries, plugins, etc. correctly in my new content.

It seems weird, with this new paradigm, to expect that a developer using `ansible-navigator` as a runner/development aid to also have `ansible-galaxy` as part of a normal `ansible` install to obtain the collections locally.

Or am I doing/understanding things horribly wrong?

##### ACTUAL RESULTS

```shell
humph. no collections were found in the following paths:
```

##### ADDITIONAL INFORMATION

N/A

cidrblock:
Fixes #608 

In the case the collection doc cache path is in the CWD, don't add a volume mount because runner will add the CWD as a mount point.

- Added a `path_is_relative_to` function to utils since it is py3.9 +
- Added a check in the collections action for this condition
- Added tests, confirmed before and after manually
- Promoted the curses fixture to tests, so it could be used by unit tests

Minimal changes were made here to fix the issue, there is an opportunity for significant refactoring of the collection action to use pathlib.  Also- while there is probably also an opportunity to check for this in the runner subsystem as a safeguard, I still believe it should be checked for in the action that is adding the volume mount and the logic described within the action. Both of these could be addressed in future PRs or as needed if the issue arises again.

### DIFF
--- a/tests/unit/actions/collections/test_collection_doc_cache_path.py
+++ b/tests/unit/actions/collections/test_collection_doc_cache_path.py
@@ -34,7 +34,10 @@ class DuplicateMountException(RuntimeError):
 @pytest.mark.parametrize("doc_cache_path", DOC_CACHE_PATHS, ids=_id_description)
 @pytest.mark.usefixtures("patch_curses")
 def test_for_duplicates_sources(
-    doc_cache_path, monkeypatch: pytest.MonkeyPatch, tmp_path: Path, mocker
+    doc_cache_path,
+    monkeypatch: pytest.MonkeyPatch,
+    tmp_path: Path,
+    mocker,
 ):
     """Ensure duplicate volume mounts are not passed to runner.
 
@@ -49,7 +52,8 @@ def test_for_duplicates_sources(
     command = f"ansible-navigator collections '--cdcp={cdc_full_path!s}' --pp never"
     monkeypatch.setattr("sys.argv", shlex.split(command))
     run_cmd_mocked = mocker.patch(
-        "ansible_navigator.runner.command.run_command", side_effect=DuplicateMountException
+        "ansible_navigator.runner.command.run_command",
+        side_effect=DuplicateMountException,
     )
     monkeypatch.chdir(working_dir)
     with pytest.raises(DuplicateMountException):


src/ansible_navigator/utils.py
```diff
@@ -384,6 +385,22 @@ def pascal_to_snake(obj):
...
+    :param child: The path that may be a child
+    :param parent: The path that may be a parent
+    :returns: Indicates the child is a child of the parent
+    """
+    if sys.version_info >= (3, 9):
+        return child.is_relative_to(parent)
+    try:
+        child.relative_to(parent)
+        return True
+    except ValueError:
```
relrod, line 400: What a weird API (on pathlib's part).

relrod (approved):

src/ansible_navigator/utils.py
cidrblock: for sure.........  I had to name it child and parent just to make sense of it, and add the unit test so I can remember what the heck they were doing.........  I always considered my parents relatives, and I think they probably consider me one.

relrod: Well that, but the whole... using exceptions as a way to say "nope, not relative"

PR #977, opened 2022-02-19
Centralize mypy configuration, increase coverage and a few type fixes

cidrblock:
In an attempt to make the `mypy` configuration cleaner with improved documentation and explanations, cover use cases outside of `tox`, and avoid different results between environments:

- Move all `mypy` config out of pre-commit to `mypy.ini` except the python-version
- Remove all in code import import statement
  - The exception here is sphinx.spelling, b/c I could only exempt all of sphinx if in `mypy.ini`
- Add and note specific ignore imports in `mypy.ini`
- Add packages dependencies as needed in `pre-commit.yaml`
- Removed the `txt-report`, and the dependency on `lxml`
- Added `sphinx`, `pylint` and `mypy` to `test-requirements.txt` so it is a complete list of developer requirements for those using tool direct IDE extensions
- Although `mypy` will install it's own dependencies, they are still listed in the `.pre-commit-config.yaml` because `pre-commit.ci` won't do it
- Added `pre-commit` checks for `mypy` with 3.6 - 3.11
  - This was done not necessarily because navigator really needs it but more instead to eliminate the possibility of different developers getting different results when running `mypy` directly or running `mypy` with an interpreter different than CI.  By checking all version navigator support, this should catch any inconsistencies or differences.
  - This will also eliminate the question why 3.6 and 3.10 and not 3.8?
  - Once cached, `mypy` runs quickly so this should not take long locally.
  - Since 3.11 was running clean and is in late alpha, it was added. This should prevent any surprises later this year

This should allow for:
- Running `mypy` at the command line outside of `tox` with any 3.6-3.11 interpreter and check for 3.6-3.11 issues
- Using the `vscode` `mypy` extension
- Running `mypy` with a 3.6 interpreter and check 3.6 and 3.10
- Running `mypy` with a 3.10 interpreter and check for 3.6 and 3.10

No "code" changes were, rather a couple new ignore statements were added because the necessary changes were going to pretty wide and extend back into the actions.

The vscode settings for the mypy extension that seemed to work well were:

```
    "mypy.runUsingActiveInterpreter": true, #runs mypy from the venv
    "mypy.targets": []  #uses the files in the mypy.ini
```
Which provide feedback like this with the extension enabled:

image

https://marketplace.visualstudio.com/items?itemName=matangover.mypy

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -178,6 +178,7 @@ repos:
           - types-PyYAML
           - types-dataclasses
           - types-docutils
+          - types-setuptools # Avoid mypy trying to do it in precommit-ci
           - types-typed-ast
         args:
           - --python-version=3.6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -163,7 +163,6 @@ repos:
           - sphinx
           - types-PyYAML
           - types-docutils
-          - types-setuptools
         args:
           - --python-version=3.10
         pass_filenames: false
@@ -179,7 +178,6 @@ repos:
           - types-PyYAML
           - types-dataclasses
           - types-docutils
-          - types-setuptools
           - types-typed-ast
         args:
           - --python-version=3.6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -143,15 +143,15 @@ repos:
   - repo: https://github.com/pre-commit/mirrors-mypy.git
     rev: v0.931
     hooks:
-      # NOTE: The 3.10 dependencies should include all dependancies for all interpreter
+      # NOTE: The latest version dependencies should include all dependancies for all interpreter
       # NOTE: versions so the configuration doesn' grow out of control.
       # NOTE: As new interpreters are added or removed, the dependency list should be auditted
-      # NOTE: For version specific dependendcies, please not the interpreter version.
+      # NOTE: For version specific dependendcies, please note the interpreter version.
       # NOTE: These are ordered from latest version to earliest for no particular reason
       # NOTE: other than a way to keep them organized.
       - id: mypy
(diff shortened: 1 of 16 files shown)


ssbarnea (changes requested):
I do agree with most changes, especially the moving of args from mypy to mypy.ini but at the same time I do not think we need to run mypy with 5 and soon 6 python versions. It makes no sense, especially as type does not influence the runtime.

Not sure if you remember that I was against running mypy with more than one python version. Due to peer pressure I accepted running with min/max versions.

What is the next move, to run all linters with all python versions because it **might** be possible for one of them to fail at some point?

Also, I cannot refrain from mentioning that this change combines multiple unrelated changes, making harder to do a proper review and ultimately slowing down merging of changes that are perfectly normal.

I will mark as -1 but if feel free to ignore my vote if others approve the change, especially as on Monday i will be on PTO.

How about cherry picking some of these changes so we can merge them and lower the size of the PR?

cidrblock:
> How about cherry picking some of these changes so we can merge them and lower the size of the PR?

Fair.

WRT to the multiple python version, I totally agree, no influence on runtime.  That said, it was my way of proving the configuration of mypy would never surprise any developer locally if the happen to be using interpreter version 3.X.  It was also my way of making sure this PR was complete, because my assumption was, if mpy ran for all interpreter version it would likely indicate the configuration, imports and type changes were all correct.

That said, I appreciate the request to have smaller PRs.  I'll mark this as draft and incrementally work up to it. 

The reason that wasn't done originally was not an attempt to disregard the benefits of smaller PR, but instead because I didn't know were this was headed and I kept adding changes until the configuration was clean.

Now that we know what much or the end-state looks like (other than the outstanding multi-version concern) I can incrementally work up to it.

cidrblock:
Closing this but will ref in future PR.

PR #978, opened 2022-02-20
[meta][pylint] ignore too-few-public-methods

relrod:
Change:
- Add too-few-public-methods to the list of ignores
- Remove now-needless disables of this lint check

This check isn't very helpful; there are many cases a class might not have
many or any public methods, such as subclassing a class for the purpose of
type-safety and wanting a runtime representation of it (i.e., not using
NewType).

Test Plan:
- pre-commit run -a pylint

-------------

After this, these are the most ignored lint checks:

```
rick@mba ansible-navigator % grep 'pylint: *disable=' src tests docs share -R | cut -d# -f2 | cut -d= -f2 | sort | uniq -c | sort -nr
  33 unused-argument
  22 too-many-branches
  21 too-many-arguments
  13 too-many-instance-attributes
  11 protected-access
  10 too-many-locals
   8 preferred-module
   7 too-many-statements
   6 cyclic-import
   5 invalid-name
   4 broad-except
   2 too-many-nested-blocks
   2 no-self-use
   2 attribute-defined-outside-init
   1 unused-import
   1 too-many-return-statements
   1 too-many-public-methods
   1 too-many-ancestors
   1 super-init-not-called
   1 redefined-builtin
   1 no-else-return
   1 import-outside-toplevel
   1 import-error
   1 expression-not-assigned
   1 dangerous-default-value
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,6 +86,7 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
     disable = [
       "duplicate-code",
       "fixme",
+      "too-few-public-methods",
       "unsubscriptable-object",
     ]
     enable = [
--- a/share/ansible_navigator/utils/catalog_collections.py
+++ b/share/ansible_navigator/utils/catalog_collections.py
@@ -39,7 +39,6 @@ PROCESSES = (multiprocessing.cpu_count() - 1) or 1
 
 
 class CollectionCatalog:
-    # pylint: disable=too-few-public-methods
     """A collection cataloger."""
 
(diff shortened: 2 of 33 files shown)


cidrblock (approved):
Ha, the need for this is so obvious now with the PR............  Thanks for doing this.

webknjaz:
I wonder if changing the setting to `min-public-methods = 1` (instead of the default of `2`) could be done instead of disabling it all together: https://pylint.pycqa.org/en/latest/technical_reference/features.html#design-checker-options.

PR #979, opened 2022-02-20
Move mypy configuration to mypy.ini

cidrblock:
This PR moves the mypy configuration out of the `.pre-commit-config.yaml` into `mypy.ini`

- Added entry to not spell check the `mypy.ini` file with cspell.

More PRs to follow related to the configuration.

Related #977

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -154,19 +154,25 @@ repos:
       # NOTE: CI automations run both checks. Whenever the versions are
       # NOTE: updated in this config, a symmetric action must be performed
       # NOTE: in `tox.ini` as well.
-      - id: mypy
+      - &_base-mypy-setup-map
+        id: mypy
         alias: mypy-py310
         name: MyPy, for Python 3.10
-        additional_dependencies: &_mypy_deps
-          - types-dataclasses # Needed for checking py3.6 with a later interpreter
+        additional_dependencies:
+          - lxml == 4.6.4 # needed for `--txt-report`
           - types-docutils
           - types-PyYAML
         args:
           - --python-version=3.10
         pass_filenames: false
-      - id: mypy
+      - <<: *_base-mypy-setup-map
+        alias: mypy-py36
         name: MyPy, for Python 3.6
-        additional_dependencies: *_mypy_deps
+        additional_dependencies:
+          - lxml == 4.6.4 # needed for `--txt-report`
+          - types-dataclasses # Needed for checking py3.6 with a later interpreter
+          - types-docutils
+          - types-PyYAML
         args:
           - --python-version=3.6
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -143,72 +143,27 @@ repos:
   - repo: https://github.com/pre-commit/mirrors-mypy.git
     rev: v0.931
     hooks:
-      # NOTE: Here MyPy is executed twice -- once for the highest and once
-      # NOTE: for the lowest supported version. They are hardcoded in the
-      # NOTE: hook invocations below. Whenever one of these versions needs
-      # NOTE: to be updated, the values in the `alias` and `name` fields
-      # NOTE: should be adjusted along with the `--python-version` CLI
-      # NOTE: option.
-      # NOTE: The external developer environment is configured to only run
-      # NOTE: the check with the lowest version on the tox side while the
-      # NOTE: CI automations run both checks. Whenever the versions are
-      # NOTE: updated in this config, a symmetric action must be performed
-      # NOTE: in `tox.ini` as well.
-      - &_base-mypy-setup-map
-        id: mypy
+      # NOTE: The latest version dependencies should include all dependancies for all interpreter
+      # NOTE: versions so the configuration doesn' grow out of control.
+      # NOTE: As new interpreters are added or removed, the dependency list should be auditted
+      # NOTE: For version specific dependendcies, please note the interpreter version.
+      # NOTE: These are ordered from latest version to earliest for no particular reason
+      # NOTE: other than a way to keep them organized.
+      - id: mypy
         alias: mypy-py310
         name: MyPy, for Python 3.10
-        additional_dependencies:
-          - lxml == 4.6.4 # needed for `--txt-report`
-          - types-docutils
+        additional_dependencies: &_mypy_deps
           - types-PyYAML
+          - types-dataclasses # Needed for checking py3.6 with a later interpreter
+          - types-docutils
         args:
-          # FIXME: get rid of missing imports ignore
-          - --ignore-missing-imports
-          - --install-types
-          - --namespace-packages
-          - --non-interactive
-          - --pretty
           - --python-version=3.10
-          - --show-column-numbers
-          - --show-error-codes
-          - --show-error-context
-          #- --strict
-          #- --strict-optional
-          - --txt-report
-          - .tox/.tmp/.mypy/
-          - docs/
-          - share/
-          - src/
-          - tests/
         pass_filenames: false
-      - <<: *_base-mypy-setup-map
-        alias: mypy-py36
+      - id: mypy
         name: MyPy, for Python 3.6
-        additional_dependencies:
-          - lxml == 4.6.4 # needed for `--txt-report`
-          - types-dataclasses
-          - types-docutils
-          - types-PyYAML
+        additional_dependencies: *_mypy_deps
         args:
-          # FIXME: get rid of missing imports ignore
-          - --ignore-missing-imports
-          - --install-types
-          - --namespace-packages
-          - --non-interactive
-          - --pretty
           - --python-version=3.6
-          - --show-column-numbers
-          - --show-error-codes
-          - --show-error-context
-          #- --strict
-          #- --strict-optional
-          - --txt-report
-          - .tox/.tmp/.mypy/
-          - docs/
-          - share/
-          - src/
-          - tests/
 
   - repo: https://github.com/pycqa/pylint.git
     rev: v2.12.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -158,9 +158,9 @@ repos:
         alias: mypy-py310
         name: MyPy, for Python 3.10
         additional_dependencies: &_mypy_deps
-          - types-PyYAML
           - types-dataclasses # Needed for checking py3.6 with a later interpreter
           - types-docutils
+          - types-PyYAML
         args:
           - --python-version=3.10
         pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -143,12 +143,17 @@ repos:
   - repo: https://github.com/pre-commit/mirrors-mypy.git
     rev: v0.931
     hooks:
-      # NOTE: The latest version dependencies should include all dependancies for all interpreter
-      # NOTE: versions so the configuration doesn' grow out of control.
-      # NOTE: As new interpreters are added or removed, the dependency list should be auditted
-      # NOTE: For version specific dependendcies, please note the interpreter version.
-      # NOTE: These are ordered from latest version to earliest for no particular reason
-      # NOTE: other than a way to keep them organized.
+      # NOTE: Here MyPy is executed twice -- once for the highest and once
+      # NOTE: for the lowest supported version. They are hardcoded in the
+      # NOTE: hook invocations below. Whenever one of these versions needs
+      # NOTE: to be updated, the values in the `alias` and `name` fields
+      # NOTE: should be adjusted along with the `--python-version` CLI
+      # NOTE: option.
+      # NOTE: The external developer environment is configured to only run
+      # NOTE: the check with the lowest version on the tox side while the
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -154,61 +154,21 @@ repos:
...
-          - --txt-report
-          - .tox/.tmp/.mypy/
-          - docs/
-          - share/
-          - src/
-          - tests/
         pass_filenames: false
-      - <<: *_base-mypy-setup-map
-        alias: mypy-py36
+      - id: mypy
```
webknjaz, line 167: This still needs an alias.

.pre-commit-config.yaml
```diff
@@ -154,61 +154,21 @@ repos:
...
-      - &_base-mypy-setup-map
-        id: mypy
+      - id: mypy
         alias: mypy-py310
         name: MyPy, for Python 3.10
-        additional_dependencies:
-          - lxml == 4.6.4 # needed for `--txt-report`
-          - types-docutils
+        additional_dependencies: &_mypy_deps
           - types-PyYAML
```
webknjaz, line 161: Maybe keep this ordered?

.pre-commit-config.yaml
```diff
@@ -154,61 +154,21 @@ repos:
...
+          - types-dataclasses # Needed for checking py3.6 with a later interpreter
```
webknjaz, line 162: I'd prefer different deps for different envs, I don't think they should influence each other.

.pre-commit-config.yaml
cidrblock: How come?

.pre-commit-config.yaml
cidrblock: sure, yeah, good catch

mypy.ini
```diff
@@ -0,0 +1,12 @@
+[mypy]
+files = docs,share,src,tests
```
webknjaz, line 2: Would this work?
```suggestion
files =
  docs,
  share,
  src,
  tests,
```

cidrblock: let's try it, would make for a better diff

mypy.ini
```diff
@@ -0,0 +1,12 @@
...
+ignore_missing_imports = true
```
webknjaz, line 7: This comment was lost:
```suggestion
ignore_missing_imports = true  # FIXME: get rid of missing imports ignore
```

.pre-commit-config.yaml
```diff
@@ -154,61 +154,21 @@ repos:
...
-          - --namespace-packages
-          - --non-interactive
-          - --pretty
           - --python-version=3.6
-          - --show-column-numbers
-          - --show-error-codes
-          - --show-error-context
-          #- --strict
-          #- --strict-optional
-          - --txt-report
```
webknjaz, line 206: Any motivation for dropping this?

.pre-commit-config.yaml
webknjaz: This would allow being explicit about choosing what to test against.

cidrblock: Not sure I understand, there is no, and should be no configuration in the pre-commit related to the "what" now that the mypy.ini file exists.  Looking at it, I don't see that is does anything in this case.  Is it future proofing for something that might happen later?

.pre-commit-config.yaml
cidrblock: Will not in the description, I don't believe anyone was using it and it pulled in a versioned dep

webknjaz: Then maybe do this separately? The PR title says "move the config" after all.

cidrblock: ok, sure.

.pre-commit-config.yaml
cidrblock: Ok.

mypy.ini
cidrblock: I'll add it back.

.pre-commit-config.yaml
webknjaz: This allows doing `tox -e lint -- mypy-py36 --all-files` and not `tox -e lint -- mypy{an_unknown_and_surprising_unobvious_version_that} --all-files`

cidrblock: Still not sure I understand, with the following configuration

```
      - id: mypy
        alias: mypy-py310
        name: MyPy, for Python 3.10
        additional_dependencies:
          - lxml == 4.6.4 # needed for `--txt-report`
          - types-docutils
          - types-PyYAML
        args:
          - --python-version=3.10
        pass_filenames: false
      - id: mypy
        alias: mypy-py36
        name: MyPy, for Python 3.6
        additional_dependencies:
          - lxml == 4.6.4 # needed for `--txt-report`
          - types-dataclasses # Needed for checking py3.6 with a later interpreter
          - types-docutils
          - types-PyYAML
        args:
          - --python-version=3.6
```
Your first example works fine:

```
(venv) x1 ➜  ansible-navigator git:(mypy_1) ✗ tox -qe lint -- mypy-py36 --all-files
MyPy, for Python 3.6.....................................................Passed
```

and your second fails as expected

```
No hook with id `mypy-py35` in stage `commit`
```

webknjaz: It was to show that barebones `mypy` id would not make it obvious for the invoker what they run. Plus, using just `mypy` would match two hooks py3.10 and py3.6, not one. Aliases add additional id but they don't prevent the original one from being used.
Just try `tox -e lint -- mypy --all-files` and you'll see that it doesn't let you specify the second check only.

cidrblock: that said.................. I've added it back since it is unrelated to the move of the configuration.

I do that in a separate PR were is doesn't distract from moving the configuration.

mypy.ini
webknjaz: This is not only about the diff. A comma-separated list without any whitespaces is rather hard to consume and it looks ugly visually. Making a list with items being distinct on separate lines helps the list to be seen as a list and not as one huge string. Whitespaces after commas would help a bit too, of course, but I feel like this extra step makes it even better :)

.pre-commit-config.yaml
webknjaz: FWIW I think that my previous comment demonstrates that it shouldn't be changed, even in a separate PR.

mypy.ini
```diff
@@ -0,0 +1,18 @@
+[mypy]
+files =
+  docs,
+  share,
+  src,
+  tests
+# FIXME: get rid of missing imports ignore
+ignore_missing_imports = true
```
webknjaz, line 8: Let's keep the trailing comma for better diffs + it's best to put comments on the same line if the length allows (otherwise something unrelated may appear between the comment line and the setting line over the course of refactoring).
```suggestion
  tests,
ignore_missing_imports = true  # FIXME: get rid of missing imports ignore
```

cidrblock: I'll add it back, but will have to remove it again after.

It will fail:

```
MyPy, for Python 3.6....................................................................................Failed
- hook id: mypy
- exit code: 2

Generated TXT report (via XSLT): /home/bthornto/github/ansible-navigator/.tox/.tmp/.mypy/index.txt
docs/_ext/single_sourced_data.py: error: Duplicate module named
"single_sourced_data" (also at "docs/_ext/single_sourced_data.py")
docs/_ext/single_sourced_data.py: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
Found 1 error in 1 file (errors prevented further checking)
```

.pre-commit-config.yaml
cidrblock: The reason it shouldn't be changed is `pass_filenames` would need to be repeated.

```
[{'id': 'mypy', 'alias': 'mypy-py310', 'name': 'MyPy, for Python 3.10', 'additional_dependencies': ['lxml == 4.6.4', 'types-docutils', 'types-PyYAML'], 'args': ['--python-version=3.10'], 'pass_filenames': False}, {'id': 'mypy', 'alias': 'mypy-py36', 'name': 'MyPy, for Python 3.6', 'additional_dependencies': ['lxml == 4.6.4', 'types-dataclasses', 'types-docutils', 'types-PyYAML'], 'args': ['--python-version=3.6'], 'pass_filenames': False}]
[{'id': 'mypy', 'alias': 'mypy-py310', 'name': 'MyPy, for Python 3.10', 'additional_dependencies': ['lxml == 4.6.4', 'types-docutils', 'types-PyYAML'], 'args': ['--python-version=3.10'], 'pass_filenames': False}, {'id': 'mypy', 'alias': 'mypy-py36', 'name': 'MyPy, for Python 3.6', 'additional_dependencies': ['lxml == 4.6.4', 'types-dataclasses', 'types-docutils', 'types-PyYAML'], 'args': ['--python-version=3.6']}]
```

cidrblock: If it were, the deserialized yaml would be the same.  By having it, it is more likely the 2 will share a common configuration and something won't get missed.

mypy.ini
webknjaz: Is it about the comma? Or the inline comment? Leave out one of them that breaks things and apply the other, I just needed to understand why...

.pre-commit-config.yaml
webknjaz: Oh, I see now! You were talking about YAML anchors and I was talking about the pre-commit setting called `alias`. Is that right? I only wanted that alias to be preserved. But I also don't see a problem with YAML aliases — the mapping values can be overridden.

mypy.ini
cidrblock: With the comma mypy's effective configuration is:

```
 'files': ['docs', 'share', 'src', 'tests', ''],
```

without it is

```
 'files': ['docs', 'share', 'src', 'tests'],
```

I believe the empty entry causes my to run against the directories specified and everything, which results in 

```
docs/_ext/single_sourced_data.py: error: Duplicate module named "single_sourced_data" (also at "docs/_ext/single_sourced_data.py")
docs/_ext/single_sourced_data.py: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
```

webknjaz: Okay, good to know, so you're right to drop it. What about the inline comment, though?

cidrblock: Removing the comment, adding an empty line, on one line with a comma do not change the way it is parsed

webknjaz: > on one line with a comma do not change the way it is parsed

What about adding the comment to the line with the setting it refers to?

.pre-commit-config.yaml
cidrblock: lol.................................... :)  I should have asked which alias :)

webknjaz: Yeah, I should've commented on the removed line with the setting too. Now I see how this was confusing that this discussion is attached to the `id` option line...

ssbarnea (approved):

mypy.ini
cidrblock: The comment "would make for a better diff" did not mean to suggest there were not other reasons to do it, I was simply highlighting one of many, that I find important.

mypy.ini
```diff
@@ -0,0 +1,18 @@
+[mypy]
+files =
+  docs,
+  share,
+  src,
+  tests
```
webknjaz, line 6: I just realized that this move dropped the trailing backslashes. It's important to keep them as they were originally used to visually communicate that these are actually folders and not filenames.
```suggestion
  docs/,
  share/,
  src/,
  tests/
```

mypy.ini
webknjaz, line 8: Just to be clear about https://github.com/ansible/ansible-navigator/pull/979/files#r810629960, this is what I expect:
```suggestion
ignore_missing_imports = true  # FIXME: get rid of missing imports ignore
```
(unless it breaks the parser, of course)

mypy.ini
cidrblock: nice.

mypy.ini
cidrblock: It doesn't break the parse, but the value ends up not being a boolean, which is why it is on the preceding line.

```
mypy.ini: [mypy]: ignore_missing_imports: Not a boolean: true  # FIXME: get rid of missing imports ignore
```

cidrblock: BTW, I expect missing imports to be ignored.

webknjaz: > It doesn't break the parser, but the value ends up not being a boolean, which is why it is on the preceding line.

Oh, okay. That's what I was afraid of. I new that `configparser` would have some of such quirks and sometimes the applications (one example where I fixed it is `tox`) would cut the comments off but sometimes this is not implemented. It's disappointing that it's not possible but at least the "why" is now documented here. Resolving.

cidrblock:
```
mypy.ini: [mypy]: ignore_missing_imports: Not a boolean: true  # FIXME: get rid of missing imports ignore
```

cidrblock: See below

webknjaz: Resolved.

webknjaz (approved):
Brilliant! Thanks for taking time to implement what I've been meaning to do for a while :)

PR #982, opened 2022-02-20
Add sphinx to mypy deps in pre-commit and test-requirements

cidrblock:
Ongoing effort to remove ignore-imports from mypy.

- Added as a dependency in the precommit entries for python3.6 and 3.10
- Added to test-requirements so mypy can be used from the command line
- One type ignore b/c sphinx spelling is not typed.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,6 +4,7 @@ flake8-docstrings
 flake8-quotes
 libtmux
 lxml
+sphinx
 pre-commit
 pytest-cov
 pytest-mock
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -4,8 +4,8 @@ flake8-docstrings
 flake8-quotes
 libtmux
 lxml
-sphinx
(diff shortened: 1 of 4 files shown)


test-requirements.txt
```diff
@@ -4,6 +4,7 @@ flake8-docstrings
 flake8-quotes
 libtmux
 lxml
+sphinx
```
webknjaz, line 7: Why put it here? It's not related to tests...

cidrblock: should we have a dev-requirements file too?

cidrblock: Mypy does need it, and mypy could be considered a "test"....?

relrod (approved):

test-requirements.txt
cidrblock: @webknjaz let's iterate on where these should be. Currently test-requirements represents everything a dev would, or soon will.

PR #985, opened 2022-02-20
Add constant colors and decorations for the UI, implement in inventory action.

cidrblock:
This add constants for the colors in use for the UI and implments them for inventory.

This will position us to change the rest and future PRs can use these.

An enum was tested here, but the type changes would have cascaded deep into the UI types.

### DIFF
--- a/src/ansible_navigator/actions/inventory.py
+++ b/src/ansible_navigator/actions/inventory.py
@@ -1,5 +1,4 @@
 """Inventory subcommand implementation."""
-import curses
 import glob
 import json
 import os
@@ -19,8 +18,10 @@ from ..configuration_subsystem import ApplicationConfiguration
 from ..runner import AnsibleInventory
 from ..runner import Command
 from ..steps import Step
+from ..ui_framework import Color
 from ..ui_framework import CursesLinePart
 from ..ui_framework import CursesLines
+from ..ui_framework import Decoration
 from ..ui_framework import Interaction
 from ..ui_framework import dict_to_form
 from ..ui_framework import warning_notification
@@ -37,17 +38,25 @@ def color_menu(colno: int, colname: str, entry: Dict[str, Any]) -> Tuple[int, in
     :returns: The color and decoration
     """
     if colname in ["__name", "title", "inventory_hostname"]:
-        return 10, 0
+        return Color.BRIGHT_GREEN, Decoration.NORMAL
     if colname == "__taxonomy":
-        return 11, 0
+        return Color.BRIGHT_YELLOW, Decoration.NORMAL
     if colname == "description":
-        return 12, 0
+        return Color.BRIGHT_BLUE, Decoration.NORMAL
     if colname == "__type":
         if entry["__type"] == "group":
-            return 11, 0
-        return 12, 0
-    colors = [14, 13, 6, 5, 4, 3, 2]
-    return colors[colno % len(colors)], 0
+            return Color.BRIGHT_YELLOW, Decoration.NORMAL
+        return Color.BRIGHT_BLUE, Decoration.NORMAL
+    colors = [
+        Color.BRIGHT_CYAN,
+        Color.BRIGHT_MAGENTA,
+        Color.CYAN,
+        Color.MAGENTA,
+        Color.BLUE,
+        Color.YELLOW,
+        Color.GREEN,
+    ]
+    return colors[colno % len(colors)], Decoration.NORMAL
 
 
 def content_heading(obj: Any, screen_w: int) -> Union[CursesLines, None]:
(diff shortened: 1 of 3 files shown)


src/ansible_navigator/ui_framework/ui_constants.py
```diff
@@ -17,7 +17,7 @@ class Color:
     """4"""
     MAGENTA = curses.COLOR_MAGENTA
     """5"""
-    CYAN = curses.COLOR_MAGENTA
+    CYAN = curses.COLOR_CYAN
```
relrod, line 20: Well, both words have an 'a' and 'n' in them. Common misspelling. 🤣

src/ansible_navigator/ui_framework/ui_constants.py
```diff
@@ -0,0 +1,51 @@
+"""Constants for use with the user interface."""
+import curses
+
+
+class Color:
```
relrod, line 5: Maybe subclass `IntEnum`?

src/ansible_navigator/ui_framework/ui_constants.py
```diff
@@ -0,0 +1,51 @@
...
+    """12"""
+    BRIGHT_MAGENTA = MAGENTA + 8
+    """13"""
+    BRIGHT_CYAN = CYAN + 8
+    """14"""
+    BRIGHT_WHITE = WHITE + 8
+    """15"""
+
+
+class Decoration:
```
relrod, line 42: Here too.

src/ansible_navigator/ui_framework/ui_constants.py
cidrblock: yessss............

src/ansible_navigator/ui_framework/ui_constants.py
cidrblock: yessss2........

src/ansible_navigator/ui_framework/ui_constants.py
```diff
@@ -0,0 +1,53 @@
...
+class Decoration(IntEnum):
+    """Constants for text decoration.
+
+    These are rarely used within the interface.
+    """
+
+    NORMAL = curses.A_NORMAL
+    """0"""
+    UNDERLINE = curses.A_UNDERLINE
+    """131072"""
```
relrod, line 53: Maybe worth including the other attributes too, for future use?

Looks like there's a partial table [here](https://docs.python.org/3/howto/curses.html#attributes-and-color) but `ipython` shows more if I do `curses.A_` and hit tab.

cidrblock: Yeah, I was hesitant to do all of them because I'm not sure we want to use the all. We use inverse for status, but things like blink make me nervous. I was thinking we incrementally add them as needed. Might be easier to catch in future PRs.

webknjaz: Yeah, I'd do this incrementally or even integrate a third-party lib to avoid having to maintain this in-repo.

src/ansible_navigator/ui_framework/ui_constants.py
```diff
@@ -0,0 +1,53 @@
+"""Constants for use with the user interface."""
+import curses
+
+from enum import IntEnum
+
+
+class Color(IntEnum):
```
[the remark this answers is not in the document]
webknjaz: And maybe have an extra structure for `MenuColor`?

cidrblock:
This PR is being closed.

relrod (approved):

src/ansible_navigator/ui_framework/ui_constants.py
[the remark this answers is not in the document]
cidrblock: @webknjaz It shouldn't grow at all, so I don't think we need to worry about scale here.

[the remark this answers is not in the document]
webknjaz: For the context/history: I asked about this because I was going to send a refactoring suggestion for `color_menu()`. But I ended up deciding not to send it then so it's not relevant anymore.

src/ansible_navigator/actions/inventory.py
```diff
@@ -19,8 +18,10 @@
 from ..runner import AnsibleInventory
 from ..runner import Command
 from ..steps import Step
+from ..ui_framework import Color
```
webknjaz, line 21: (nit) it'd be clearer if this was
```suggestion
from ..ui_framework.ui_constants import Color
```
Well, even
```suggestion
from ..ui_framework.constants import Color
```

webknjaz (approved):

PR #986, opened 2022-02-20
Application Configuration to dataclass

cidrblock:
Switch ApplicationConfiguration from a SimpleNamepsace to a data class.

Only the minimal changes were made to pass tests/mypy/linters

And missing field that was in use added.


src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -134,16 +138,22 @@ class SubCommand(SimpleNamespace):
...
-    entries: List[SettingsEntry] = []
+    application_version: str
+    entries: List[SettingsEntry]
     internals: SimpleNamespace
+    post_processor: "NavigatorPostProcessor"
     subcommands: List[SubCommand]
-    post_processor = Callable
+
+    application_name: str = ""
+    #  = field(default_factory=list)
```
relrod, line 153: nuke

src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -134,16 +138,22 @@ class SubCommand(SimpleNamespace):
...
 
-class ApplicationConfiguration(SimpleNamespace):
+@dataclass
+class ApplicationConfiguration:
+    # pylint: disable=too-many-instance-attributes
     """The main object for storing an application config"""
 
-    application_name: str = ""
-    entries: List[SettingsEntry] = []
+    application_version: str
```
relrod, line 146: Prefer adding fields in a separate PR instead.

src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -134,16 +138,22 @@ class SubCommand(SimpleNamespace):
...
+    entries: List[SettingsEntry]
     internals: SimpleNamespace
+    post_processor: "NavigatorPostProcessor"
```
relrod, line 149: From a readability standpoint, I'd prefer just importing it unconditionally and not using a string for the type name.

cidrblock: It results in a circular import because post process import from here.....

src/ansible_navigator/configuration_subsystem/definitions.py
cidrblock: No choice really, it was already in use by the NavigatorConfiguration:

```
NavigatorConfiguration = ApplicationConfiguration(
    application_name=APP_NAME,
    application_version=VERSION,
```

src/ansible_navigator/configuration_subsystem/definitions.py
cidrblock: what probably really should be done is NavigatorPostProcessor should have a base class defined here, but, that's for another day :)

cidrblock:
In case anyone is curious about the comment style vs `#:` it allows for mouse over support in the IDE:
image

src/ansible_navigator/configuration_subsystem/definitions.py
cidrblock: Sorry, I missed explaining that...

relrod: Gotcha. Resolving then.

src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -134,16 +138,21 @@ class SubCommand(SimpleNamespace):
     epilog: Union[None, str] = None
 
 
-class ApplicationConfiguration(SimpleNamespace):
+@dataclass
+class ApplicationConfiguration:
+    # pylint: disable=too-many-instance-attributes
```
relrod, line 143: Can we globally kill this? It's pretty silly. Especially for dataclasses. They store things... data, even!

cidrblock: I'm game, although it might bring attention to the need for a class that should have some subclasses, I don't think I'd ever pick up on it.

src/ansible_navigator/configuration_subsystem/definitions.py
webknjaz: _(could've been caught by `eradicate`)_

src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -134,16 +138,21 @@ class SubCommand(SimpleNamespace):
     epilog: Union[None, str] = None
 
 
-class ApplicationConfiguration(SimpleNamespace):
+@dataclass
```
webknjaz, line 141: It's usually preferable to have immutable data structures
```suggestion
@dataclass(frozen=True, slots=True)
```

src/ansible_navigator/configuration_subsystem/definitions.py
cidrblock: I think that would be a great one to try out.  If you help out with the docstrings, we can do that next....

src/ansible_navigator/configuration_subsystem/definitions.py
cidrblock: The settings are reevaluated and updated with every `:command`. If you get time, pull down the PR and let me know how it works with it frozen...  I'm curious.

src/ansible_navigator/configuration_subsystem/definitions.py
webknjaz: I don't think it will need any effort beyond reverting the removal you did.

src/ansible_navigator/configuration_subsystem/definitions.py
webknjaz: Got it. Is it the same for every class in this module? If yes, maybe it's worth explaining in a code comment at the very top.

webknjaz: (feel free to "Resolve" this conversation when you read it)

cidrblock:
This PR is being closed.

relrod (approved):

PR #988, opened 2022-02-21
Rework KVS with types and extend MutableMapping instead of dict

relrod:
Change:
- Make KVS subclass MutableMapping[str, str] instead of dict
- Add types to KVS
- Move KVS into first-class package and mount it in EEs differently
- Add unit tests for KVS
- Touch up pre-commit config for mypy (drop --ignore-missing-imports,
  and drop tests and docs dirs, add a few missing deps.

Test Plan:
- Ran pre-commit for both mypy versions
- New unit tests for KVS


src/ansible_navigator/utils/key_value_store.py
```diff
@@ -61,28 +87,33 @@ def iteritems(self):
...
-        return list(self.iteritems())
-
-    def __contains__(self, key):
+        return KVSItemsView(self)
+
+    # mypy complains about this 'str' because it's more specific than the 'object'
+    # that dict's Mapping superclass wants. However, it's correct in our case.
+    # Our sqlite table expects string keys (column type 'text'). It's an error to
+    # pass something else. If we pass something too weird, the sqlite3 lib will
+    # throw an error at runrtime.
```
cidrblock, line 115: Should we check isinstance(key, str) here and raise our own exception?  Since we know we only want a str....

tests/unit/utils/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+import os
+import sys
+
+import pytest
+
+from ansible_navigator.utils.key_value_store import KeyValueStore
+
+
+@pytest.fixture
+def empty_kvs(tmp_path):
```
cidrblock, line 12: We haven't talked about it, how do you feel about typing in tests?  I've started to add a few and find it useful during debugging and for IDE mouseover docs

tests/unit/utils/test_key_value_store.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def test_kvs_keys(kvs):
+    """Test KVS keys()"""
+    # We get back a KVSKeysView which is just a collections.abc.KeysView in
+    # disguise.
+    keys = kvs.keys()
+    assert "banana" in keys
+    assert "foo" not in keys
+    assert len(keys) == len(kvs)
```
cidrblock, line 75: Could something like this work?  Since the pattern is repeated a few times..... might be handy to have

https://gist.github.com/cidrblock/19fa73e4c9f2c00e9f8ecdc9849edce8

cidrblock (changes requested):
Cool stuff

src/ansible_navigator/utils/key_value_store.py
cidrblock: Good point.... mypy does check the share directory so it should catch someone trying to stuff something odd in :)

src/ansible_navigator/actions/collections.py
```diff
@@ -319,8 +336,18 @@ def _run_runner(self) -> None:
         if self._args.execution_environment:
             self._logger.debug("running collections command with execution environment enabled")
             python_exec_path = "python3"
-
-            container_volume_mounts = [f"{share_directory}/utils:{share_directory}/utils"]
+            utils_lib = os.path.join(
```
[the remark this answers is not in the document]
relrod: We already use os.path a fair number of times in this action, so I kept it this way for consistency. Let's grab those in a separate PR?

tests/unit/utils/test_key_value_store.py
relrod: I'm not sure this would clean things up much, I'd still want to keep the `len()` assert. It's asserting that iterating over the keys (resp. values, items) produces the same number of results as the number of items the KVS thinks it has in it.

And the line below is testing that we don't alter the keys (resp. values, items) in some way unexpectedly.

So I'd still want to keep these asserts, I think.

cidrblock: ok......  I missed the len test wasn't testing the list.....It was still early :)

cidrblock (approved):

Issue #554, opened 2021-08-31
[maintenance][packaging] Set up PEP 517 settings in `pyproject.toml` explicitly


webknjaz:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

Currently, the project relies on implicit PEP 517 fallbacks. We need to declare the corresponding section in order to have better control and understanding of how the distribution packages are getting built.

PR #989, opened 2022-02-21
Remove docs dependency on setuptools_scm

cidrblock:
Part of ongoing work to remove `ignore_imports` for mypy.

In this case, rather than account for setuptools_scm in the mypy.ini as untyped (https://github.com/pypa/setuptools_scm/issues/501), simply remove the dependency on it entirely.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,7 +37,7 @@ copyright = author  # pylint:disable=redefined-builtin
 NAVIGATOR_VERSION = version("ansible_navigator")
 
 # The short X.Y version
-version = ".".join(NAVIGATOR_VERSION.split(".")[:2])
+major_minor_version = ".".join(NAVIGATOR_VERSION.split(".")[:2])
 
 # The full version, including alpha/beta/rc tags
 release = NAVIGATOR_VERSION
@@ -274,5 +274,5 @@ myst_substitutions = {
     "project": project,
     "release": release,
     "release_l": f"`v{release}`",
-    "version": version,
+    "version": major_minor_version,
 }
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -36,7 +36,8 @@ copyright = author  # pylint:disable=redefined-builtin
 
 NAVIGATOR_VERSION = get_version("ansible_navigator")
 
-# The short X.Y.Z version
+# The short X.Y version
+# including .Z, resulting in the X.Y.Z version
 version = ".".join(NAVIGATOR_VERSION.split(".")[:3])
 
 # The full version, including alpha/beta/rc tags
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,17 +3,16 @@
 # Ref: https://www.sphinx-doc.org/en/master/usage/configuration.html
 """Configuration file for the Sphinx docs."""
 
-from functools import partial
+import re
 from pathlib import Path
 from sys import path
 
-from setuptools_scm import get_version
+from ansible_navigator import _version
 
 
 # -- Path setup --------------------------------------------------------------
 
 PROJECT_ROOT_DIR = Path(__file__).parents[1].resolve()
-get_scm_version = partial(get_version, root=PROJECT_ROOT_DIR)
 
 # Make in-tree extension importable in non-tox setups/envs, like RTD.
 # Refs:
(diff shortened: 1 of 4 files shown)


webknjaz (changes requested):
setuptools-scm is necessary for building from Git checkout. It is by design that it produces unique versions for builds that make it easier to identify the source of a specific build. It should remain in use.

It is also visible @ https://ansible-navigator--989.org.readthedocs.build/en/989/changelog/#v1-1-0-alpha-1-as-of-today-unreleased-draft that this PR breaks both the displayed version and the hardcoded version identifier that became incompatible with PEP 440 (it must be compliant for PyPI uploads not to be rejected and for pip to understand the semantics of the given version).

docs/conf.py
```diff
@@ -36,23 +34,19 @@
 author = f"{project} project contributors"
 copyright = author  # pylint:disable=redefined-builtin
 
-# fmt: off
+NAVIGATOR_VERSION = version("ansible_navigator")
+
 # The short X.Y version
-version = ".".join(
```
webknjaz, line 41: This is a Sphinx setting. Don't lose it. Although, I'm still against breaking what setuptools-scm provides. If you want to reuse the installed dist metadata, you'd have to (1) integrate setuptools-scm into packaging first and then (2) somehow solve the self-reinstallation on every local build in an optimal way (without `tox -r`).

docs/conf.py
```diff
@@ -3,17 +3,15 @@
 # Ref: https://www.sphinx-doc.org/en/master/usage/configuration.html
 """Configuration file for the Sphinx docs."""
 
-from functools import partial
+
+from importlib.metadata import version
```
webknjaz, line 7: This shadows a built-in/internal Sphinx setting, rename it.

docs/conf.py
```diff
@@ -193,7 +187,7 @@
     "github_user": github_repo_org,
     "github_repo": github_repo_name,
     "github_version": "main/docs/",
-    "current_version": version,
```
webknjaz, line 196: You don't need to do this, you still have to resurrect the `version` variable as a string, since it's something that Sphinx uses internally and exposes as a `|version|` substitution.

docs/conf.py
```diff
@@ -36,23 +34,19 @@
...
-# fmt: off
-# The short X.Y version
-version = ".".join(
-    get_scm_version(
-        local_scheme="no-local-version",
-    ).split(".")[:3],
-)
+NAVIGATOR_VERSION = get_version("ansible_navigator")
+
+# The short X.Y.Z version
```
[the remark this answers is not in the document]
cidrblock: but it's totally wrong? :)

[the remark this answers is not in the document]
webknjaz: It's sort of a balance. I'd maybe put an additional inline comment to explain the rest?

pyproject.toml
```diff
@@ -45,6 +45,8 @@
     name = "Contributor-facing changes"
     showcontent = true
 
+[build-system]
+requires = ["setuptools>=45", "wheel", "setuptools_scm[toml]>=6.2"]
```
webknjaz, line 49: `wheel` shouldn't be in this list, `setuptools`' PEP 517 build backend auto-adds it already.

pyproject.toml
```diff
@@ -95,3 +97,5 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
 
     [tool.pylint.format]
     max-line-length = 100
+
+[tool.setuptools_scm]
```
webknjaz, line 101: This whole dist packaging setup refactoring should go into a separate PR as it's not related to docs. It could be accepted separately while the rest of this PR is still in a state that has to be rejected.

docs/conf.py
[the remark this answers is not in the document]
cidrblock: I'm assuming you mean this:  https://github.com/sphinx-doc/sphinx/blob/a8eb1aab72a514b7257c780c0c57432732a702fb/sphinx/templates/quickstart/conf.py_t

I'm going to leave this, for exactly the reason you mentioned, when diffed, it will highlight the the deviation from the tempalte.

pyproject.toml
webknjaz: (and of course move this to a separate PR as well)

pyproject.toml
cidrblock: It can't yet.  The _version file contains more than just the package version.

docs/conf.py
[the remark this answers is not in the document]
cidrblock: I'll compromise, brb

pyproject.toml
webknjaz: See also: https://github.com/ansible/ansible-navigator/issues/554

pyproject.toml
webknjaz: No, you can do that now. Just don't point it at an existing file. Instead, point it at a gitignored `_scm_version.py` that you'd import from the `_version.py`.
Example: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/pyproject.toml#L16 + https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/src/sphinxcontrib/towncrier/_version.py.

P.S. Also, the same approach is used in `ansible-pylibssh`.

webknjaz: Pro tip: create a [`_scm_version.pyi`](https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/src/sphinxcontrib/towncrier/_scm_version.pyi) to take care of typing.

cidrblock:
This PR is being clsoed.

Issue #553, opened 2021-08-31
[maintenance][packaging] Make `setuptools-scm` responsible for versioning


webknjaz:
##### ISSUE TYPE
 - Feature Idea

##### SUMMARY

The current versioning setup relies on remembering to re-hardcode the version constant on each release in addition to having to set the tag to the same value. This is usually fragile.
So I suggest integrating setuptools-scm as many projects in the Python ecosystem do. It allows relying on Git tags to get versions on build and also provides nice versions for intermediate commits.

Implementation detail: use a gitignored `write_to = "_scm_version.py"` as shown in https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/pyproject.toml#L16 + https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/src/sphinxcontrib/towncrier/_version.py#L5.
Pro tip: create a [`_scm_version.pyi`](https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/532e343/src/sphinxcontrib/towncrier/_scm_version.pyi) to take care of typing.

webknjaz:
Strong :-1: on pbr, it is a hackish use of deprecated distutils. Plus other DevTools projects use `setuptools-scm`. pbr is mostly used in the openstack bubble while the rest of the ecosystem relies on the project backed by PyPA.

PR #990, opened 2022-02-21
Move collection doc cache version to dedicated file

cidrblock:
This will allow for <s>the</s> a version file to be updated later based on repo tags rather than being hard coded in the project.

### DIFF
--- a/src/ansible_navigator/_version_doc_cache.py
+++ b/src/ansible_navigator/_version_doc_cache.py
@@ -2,7 +2,7 @@
 
 .. note::
 
-   __version_collection_doc_cache__:
+   ``__version_collection_doc_cache__``:
 
    Indicates the version of the schema of the collection doc cache
    this is checked during initialization, if the version of the cache


src/ansible_navigator/_version_doc_cache.py
```diff
@@ -0,0 +1,14 @@
+"""Version information for ansible-navigator collection doc cache.
+
+.. note::
+
+   __version_collection_doc_cache__:
```
webknjaz, line 5:
```suggestion
   ``__version_collection_doc_cache__``:
```

src/ansible_navigator/_version_doc_cache.py
```diff
@@ -0,0 +1,14 @@
...
+
+   Indicates the version of the schema of the collection doc cache
+   this is checked during initialization, if the version of the cache
+   differs from below, the cache will be rebuilt.  This should be
+   incremented when the schema changes and need not correspond to the
+   application version, although keeping the major in sync is probably
+   not a bad idea to minimize the amount of stale docs in the user's cache
+"""
+__version_collection_doc_cache__ = "1.0"
```
webknjaz, line 14: Maybe these docstrings should be separate?
```suggestion
"""Version information for ansible-navigator collection doc cache."""

__version_collection_doc_cache__ = "1.0"
"""
Indicates the version of the schema of the collection doc cache
this is checked during initialization, if the version of the cache
differs from below, the cache will be rebuilt.  This should be
incremented when the schema changes and need not correspond to the
application version, although keeping the major in sync is probably
not a bad idea to minimize the amount of stale docs in the user's cache
"""
```

src/ansible_navigator/_version_doc_cache.py
cidrblock: This was previously approved in #651

src/ansible_navigator/_version_doc_cache.py
cidrblock: I think it's fine as is.

cidrblock:
> I don't think `_version.py` should be reused for this. I'd rather have an additional dedicated `_scm_version.py` file that is gitignored as explained in #553 [#989 (comment)](https://github.com/ansible/ansible-navigator/pull/989#discussion_r811236727).

Description revised to read `a` version file rather than `the` version file.

webknjaz (approved):

PR #991, opened 2022-02-21
Transition from __version__ to version

cidrblock:
This will allow the `__version__` entry in the `._version` file eventually be `version`, and account for running from source when the `._version` file doesn't exist at all.

Another small step toward removing the hard coded version from the repo and letting setuptools_scm generate it during build.

(This will get cleaned up in 3-4 PRs)

### DIFF
--- a/src/ansible_navigator/__init__.py
+++ b/src/ansible_navigator/__init__.py
@@ -9,6 +9,6 @@ try:
     from ._version import __version__
 except ImportError:
     try:
-        from ._version import version
+        from ._version import version  # type: ignore[attr-defined]
     except ImportError:
         __version__ = "source"
--- a/src/ansible_navigator/__init__.py
+++ b/src/ansible_navigator/__init__.py
@@ -1,2 +1,14 @@
-"""The ansible-navigator application."""
-from ._version import __version__  # noqa: F401
+"""The ansible-navigator application.
+
+This is a transitional configuration during the migration
+from the hardcoded __version__ in _version to
+version being populated during the build.
+"""
+
+try:
+    from ._version import __version__
+except ImportError:
+    try:
+        from ._version import version
+    except ImportError:
+        __version__ = "source"
--- a/src/ansible_navigator/__init__.py
+++ b/src/ansible_navigator/__init__.py
@@ -8,7 +8,4 @@ version being populated during the build.
 try:
     from ._version import __version__
 except ImportError:
-    try:
-        from ._version import version  # type: ignore[attr-defined]
-    except ImportError:
-        __version__ = "source"
+    __version__ = "source"
--- a/src/ansible_navigator/_version.py
+++ b/src/ansible_navigator/_version.py
@@ -16,4 +16,5 @@
    not a bad idea to minimize the amount of stale docs in the user's cache
 """
 __version__ = "1.1.0a1"
+version = __version__
 __version_collection_doc_cache__ = "1.0"
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -9,7 +9,7 @@ from typing import List
 from typing import Tuple
 from typing import Union
 
-from .._version import __version__ as VERSION
+from .. import __version__ as VERSION
 from ..utils import ExitMessage
 from ..utils import LogMessage
 from ..utils import abs_user_path
--- a/src/ansible_navigator/configuration_subsystem/parser.py
+++ b/src/ansible_navigator/configuration_subsystem/parser.py
@@ -9,7 +9,7 @@ from typing import Dict
 from typing import Tuple
 from typing import Union
 
-from .._version import __version__
+from .. import __version__
 from ..utils import oxfordcomma
 from .definitions import ApplicationConfiguration
 from .definitions import Constants as C


src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,14 @@
...
+"""The ansible-navigator application.
+
+This is a transitional configuration during the migration
+from the hardcoded __version__ in _version to
+version being populated during the build.
+"""
+
+try:
+    from ._version import __version__
+except ImportError:
```
webknjaz, line 10: Could you move this complexity into `_version` per #553?

webknjaz: But I don't think we need to add the complexity now only to remove it later.

cidrblock: You are assuming there will be a scm_version file, which has not been added to any PR or discussed. 

Also- I find the justification pointing to an issue that is being updated in the background a little awkward.
image

webknjaz: I updated the issue because I realized it didn't have enough details and I didn't want to link the PR discussions because they tend to become hidden and hard to find on updates :man_shrugging:

src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,14 @@
-"""The ansible-navigator application."""
-from ._version import __version__  # noqa: F401
+"""The ansible-navigator application.
+
+This is a transitional configuration during the migration
+from the hardcoded __version__ in _version to
```
webknjaz, line 4: this line contains non-wrapped identifiers

src/ansible_navigator/__init__.py
webknjaz: FWIW I don't think this change is reasonable. And `version` var also doesn't exist so I don't see any point in using it as a fallback. The reason I wanted a separate module first is that this technique has been tried out in a few projects and corner cases have been addressed.

src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,14 @@
...
+version being populated during the build.
+"""
+
+try:
+    from ._version import __version__
+except ImportError:
+    try:
+        from ._version import version  # type: ignore[attr-defined]
+    except ImportError:
+        __version__ = "source"
```
webknjaz, line 14: I don't think this is the right value for the fallback. In source checkouts, even with an editable install mode, this won't happen.

src/ansible_navigator/__init__.py
webknjaz: Also, this whole fallback dance should be happening in a dedicated Python module, not in `__init__`.

src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,14 @@
...
+This is a transitional configuration during the migration
+from the hardcoded __version__ in _version to
+version being populated during the build.
+"""
+
+try:
+    from ._version import __version__
+except ImportError:
+    try:
+        from ._version import version  # type: ignore[attr-defined]
```
relrod, line 12: Could we avoid the type ignore by setting `version = __version__` in `_version` for now?

(yo dawg, I heard you like versions, so I put `version` in your `_version` so you can `version` while you `__version__`) ???

src/ansible_navigator/__init__.py
webknjaz: Plus, if the second import succeeds, it'll cause a `NameError` in other components.

src/ansible_navigator/__init__.py
cidrblock: But version _will_ exist, this separates the changes to the build from code changes.  when this goes in, the next PR can entirely be a change to the pyproject.toml file (I think)

src/ansible_navigator/__init__.py
cidrblock: It's only going to be there for a couple hours.

src/ansible_navigator/__init__.py
cidrblock: this will catch `python -m ansible_navigator` and return the version as `source`

src/ansible_navigator/__init__.py
cidrblock: oh, that's an interesting idea..........  I think that might work.

webknjaz: It's a bad idea. If this import succeeds, then there will be no `__version__` exposed in `__init__.py` which will cascade into other parts trying to access it.

src/ansible_navigator/__init__.py
webknjaz: > the next PR can entirely be a change to the pyproject.toml file (I think)

That change will have to be coupled with integrating the SCM-generated file into the project. Otherwise, it won't be really atomic.

webknjaz: > But version _will_ exist, this separates the changes to the build from code changes.

This still does not address the need to keep the detection code out of `__init__.py`.

src/ansible_navigator/__init__.py
webknjaz: That scenario should be covered by calling the version detection helper from setuptools-scm.

webknjaz: Although, it's probably not the best idea to support it because if it's not installed, the import machinery will fail to import the missing deps long before this variable will get used. And if it's installed, then it won't cause an ImportError. This line is probably unreachable.

src/ansible_navigator/_version.py
```diff
@@ -16,4 +16,5 @@
    not a bad idea to minimize the amount of stale docs in the user's cache
 """
 __version__ = "1.1.0a1"
+version = __version__
```
webknjaz, line 19: This variable is never accessed/used.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -9,7 +9,7 @@
 from typing import Tuple
 from typing import Union
 
-from .._version import __version__ as VERSION
+from .. import __version__ as VERSION
```
webknjaz, line 12: I don't think that the project internals should access version through the package init. They should just get it from the `_version` as before. And `_version` should guard against import errors internally.

webknjaz (changes requested):
At this point, I'm not convinced that this PR is even needed at all. It just adds complexity to the wrong place (`__init__.py`) that should avoid having any logic in it. That logic should go into `_version.py` and is tightly coupled with integrating `setuptools-scm` because it's a part of how it's used.

cidrblock:
Closing this PR in favor of a PR with a larger scope because it has become nearly impossible to go about it this way.

PR #993, opened 2022-02-21
Ignore PRs from any GitHub apps in Chronographer

webknjaz:
When the list in the config is explicit, then only those bots get ignored but now PRs from any GitHub Apps will not require change notes.

### DIFF
--- a/.github/chronographer.yml
+++ b/.github/chronographer.yml
@@ -2,10 +2,6 @@
 enforce_name:
   suffix: .md
 exclude:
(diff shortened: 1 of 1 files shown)


cidrblock (approved):
nice.

webknjaz:
FYI the CI failures are flaky and absolutely unrelated. Hopefully, restarting the jobs helps. The flakiness seems to be related to `ansible-galaxy` being unable to communicate with the API.

PR #994, opened 2022-02-21
Set version using setuptools_scm during build

cidrblock:
This will let setuptools_scm populate the version file during installation.  

- when running from source, the version will be returned as "source"
- code updates as needed, because the version written is just `version`
- removes hard coded version from source

Here's a sample:

```
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = '0.7.4.dev338+gb870a002.d20220221'
version_tuple = (0, 7, 4, 'dev338', 'gb870a002.d20220221')
```

I suspect over time additional work will need to be done here, but this will allow the mypy ignore_includes work to continue.

Related #553

The approach suggesting `_scm_version` can be implemented as needed when needed.  It did not seem necessary and there was little evidence of it being a common pattern.


pyproject.toml
```diff
@@ -45,6 +45,12 @@
     name = "Contributor-facing changes"
     showcontent = true
 
+[build-system]
+requires = [
+  "setuptools>=45",
+  "wheel",
```
webknjaz, line 51: This is not necessary, I even had to fix setuptools' upstream doc recently to clarify this.
```suggestion
```

pyproject.toml
```diff
@@ -45,6 +45,12 @@
...
+  "setuptools_scm[toml]>=6.2"
```
webknjaz, line 52: In fact, it's probably nicer to also make this sparse.
```suggestion
  "setuptools >= 45",
  "setuptools_scm[toml] >= 6.2"
```

src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,6 @@
 """The ansible-navigator application."""
-from ._version import __version__  # noqa: F401
+
+try:
+    from ._version import version
+except ImportError:
+    version = "source"
```
webknjaz, line 6: Is there any strong reason to rename the exposed variable? `__version__` is pretty much a de-facto standard.
```suggestion
    from ._version import version as __version__
except ImportError:
    __version__ = "source"
```

cidrblock: no, good idea....

src/ansible_navigator/__init__.py
```diff
@@ -1,2 +1,6 @@
 """The ansible-navigator application."""
-from ._version import __version__  # noqa: F401
+
+try:
```
webknjaz, line 3: Let's not have any logic in `__init__.py` and move it to a dedicated module. Having import-time calculations should be avoided when possible, or at least hidden from the public API layer.

webknjaz:
@cidrblock don't forget to add `src/ansible_navigator/.gitignore` with the following contents:
```gitignore
_version.py
```

src/ansible_navigator/__init__.py
webknjaz: FWIW I still insist on the import-fallback dance being in a dedicated module that only takes care of the version handling/normalization as per the single responsibility principle.

src/ansible_navigator/__init__.py
cidrblock: There is only one scenario that I am aware of where the ImportError will get hit.

cidrblock: I thought the cost of a try except was minimal, especially when it wasn't thrown.... and was being improved to be almost negligible.

webknjaz: Honestly, I don't think that scenario that you mentioned earlier is realistic. Have you been able to trigger it?

pyproject.toml
cidrblock: Without, these warnings are shown:

```
WARNING: Missing build requirements in pyproject.toml for file:///home/bthornto/github/ansible-navigator.
 WARNING: The project does not specify a build backend, and pip cannot fall back to setuptools without 'wheel'.
```

.gitignore
```diff
@@ -34,6 +34,9 @@ share/python-wheels/
 *.egg
 MANIFEST
 
+# Version created and populated by setuptools_scm
+_version.py
```
webknjaz, line 38: I'd prefer a `.gitignore` file in the folder where this is needed. The current entry as added right now excludes any file with this name. So we'd need to use a full path to be precise in what we ignore if it stays in this file.

Although, it probably should go to the end of the file and not be in the middle since it seems to be partially generated from https://gitignore.io, and putting things in the middle of the template makes it harder to update it or track the boilerplate separately from the projec-specific entries.

Still, I'd prefer a gitignore in a more accurate location.

src/ansible_navigator/__init__.py
webknjaz: But the main motivation here is to keep the package init clean.

pyproject.toml
```diff
@@ -45,6 +45,12 @@
     name = "Contributor-facing changes"
     showcontent = true
 
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+  "setuptools>=45",
+  "setuptools_scm[toml]>=6.2",
```
webknjaz, line 52: Mind keeping these sparse?
```suggestion
  "setuptools >= 45",
  "setuptools_scm[toml] >= 6.2",
```

src/ansible_navigator/__init__.py
cidrblock: If you have the deps installed and don't ever install ansible-navigator:

```
(venv) x1 ➜  src git:(build) python -m ansible_navigator --version
__main__.py source
(venv) x1 ➜  src git:(build) 
```

which has be necessary in the past when troubleshooting the share directory path resolution and manually putting it's contents in different locations vs. what setuptools would do.

pyproject.toml
cidrblock: Adding this `build-backend = "setuptools.build_meta"` eliminates the WARNINGS

src/ansible_navigator/__init__.py
cidrblock: I prefer this approach as it is obvious and easy to understand.

I looked for the scm_file pattern, it is in limited use:

https://sourcegraph.com/search?q=context:global+_scm_version.py&patternType=literal

Where writing to _version is significantly more common.

https://sourcegraph.com/search?q=context:global+write_to+%3D+%5C%22.*/_version.py%5C%22&patternType=regexp


There is one instance here: https://github.com/sphinx-contrib/sphinxcontrib-towncrier/pull/10
is was added without explanation or justification.

Another here, again without justification or explanation. https://github.com/ansible/pylibssh/pull/47

The related issue has no reasoning, explanation or justification: https://github.com/ansible/ansible-navigator/issues/553

It can be added when needed.

src/ansible_navigator/__init__.py
webknjaz: Does this mean that it is a supported mechanism — having people install the runtime deps manually but not installing/building the distribution itself?

Also, you seem to be confusing setuptools and pip. The former is a PEP 517 build back-end and the latter is a build front-end (which is what unpacks the content into the target locations).

.gitignore
cidrblock: It was added here, directly after build excludes because it is related to the build.

It is not uncommon for it to be in the .gitignore:

https://github.com/ansible/pylibssh/blob/devel/.gitignore#L398

https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/master/.gitignore#L132

https://github.com/ansible-community/ansible-lint/blob/main/.gitignore#L50

Nor it is always at the bottom.

pyproject.toml
webknjaz: That is correct. It is strongly recommended to have `build-backend` set explicitly, otherwise, the fallback is `setuptools.build_meta.__legacy__` and not `setuptools.build_meta` which may behave slightly differently.

.gitignore
webknjaz: > It was added here, directly after build excludes because it is related to the build.

I wish it was structured better. Ideally, the boilerplate should go first and have start/end markers.

> It is not uncommon for it to be in the .gitignore:

You're right. Yet this doesn't mean that it's not handy to have a dedicated file without a boilerplate: we have one right in this project, in the changelog fragments directory, for example.

Still, if you want to have it here, at least be precise about which one to exclude:
```suggestion
/src/ansible_navigator/_version.py
```

> [ansible/pylibssh@`devel`/.gitignore#L398](https://github.com/ansible/pylibssh/blob/devel/.gitignore?rgh-link-date=2022-02-21T18%3A40%3A42Z#L398)
> 
> [sphinx-contrib/sphinxcontrib-towncrier@`master`/.gitignore#L132](https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/master/.gitignore?rgh-link-date=2022-02-21T18%3A40%3A42Z#L132)
> 
> [ansible-community/ansible-lint@`main`/.gitignore#L50](https://github.com/ansible-community/ansible-lint/blob/main/.gitignore?rgh-link-date=2022-02-21T18%3A40%3A42Z#L50)
> 
> Nor it is always at the bottom.

It is, in all of the examples you've linked above. The `/.test-results/` entry is also project-specific which is why it shares the "bottom". In some places, there's a marker that shows where the boilerplate ends, like `# End of https://www.gitignore.io/api/git,linux,pydev,python,windows,pycharm+all,jupyternotebook,vim,webstorm,emacs,dotenv` but others don't have it for some reason.

src/ansible_navigator/__init__.py
cidrblock: It mean as a owner of the repo there may be times where troubleshooting of share directory related issues you will need to run directly from source without having built or installed it. I cannot think of a better explanation than "source" as the fallback, because this is the only scenario where it could/would realistically be hit, and you would be indeed running from source.

cidrblock: I am open to suggestions about what else to put here........... and understanding when else it could be hit......

pyproject.toml
cidrblock: sparse in this context took me a minute.... :)

.gitignore
cidrblock: I am going to move it, brb

cidrblock: ok, I prefer that, indicates it is there for a good reason and not a cut and paste from ?

src/ansible_navigator/__init__.py
webknjaz: > I prefer this approach as it is obvious and easy to understand.

"Obvious" is subjective. To me, it's more obvious to have things with dedicated responsibilities rather than multipurpose all-in-one modules.

> I looked for the scm_file pattern, it is in limited use:
> 
> [sourcegraph.com/search?q=context:global+_scm_version.py&patternType=literal](https://sourcegraph.com/search?q=context:global+_scm_version.py&patternType=literal)
> 
> Where writing to _version is significantly more common.
> 

Because it's the default example but it doesn't mean that it fits all sizes. It's coming from a small snippet in README that shows that this setting exists. It's like "hello world" examples — we don't copy-paste them into our code just because they exist, we adjust them to be reasonable. "Because somebody else did it" is not the best argument to do something really, unless there is a good explanation behind it, that is.

> [sourcegraph.com/search?q=context:global+write_to+%3D+%5C%22.*/_version.py%5C%22&patternType=regexp](https://sourcegraph.com/search?q=context:global+write_to+%3D+%5C%22.*/_version.py%5C%22&patternType=regexp)
> 
> There is one instance here: [sphinx-contrib/sphinxcontrib-towncrier#10](https://github.com/sphinx-contrib/sphinxcontrib-towncrier/pull/10) it was added without explanation or justification.

This was mostly a copy from pylibssh.

> Another here, again without justification or explanation. [ansible/pylibssh#47](https://github.com/ansible/pylibssh/pull/47)

I was working on that alone and didn't feel like it needed explaining but I did explain everything here.

> The related issue has no reasoning, explanation or justification: #553

I did explain it here and in other PRs, though.

> It can be added when needed.

It is needed. Exposing the details of a "sometimes-existing" module to a package `__init__` is not something I'd approve.

.gitignore
```diff
@@ -1,9 +1,16 @@
...
-/changelogs/.plugin-cache.yaml
+# The following entries have been rationalized and should all have detailed explanations
+
+# Version created and populated by setuptools_scm
+/src/ansible_navigator/_version.py
 
 # tmux integration test logs
 .test_logs
 
+# The following entries came from an untracted source
```
webknjaz, line 9: This looks like something that would belong in a separate PR.

pyproject.toml
```diff
@@ -45,6 +45,12 @@
     name = "Contributor-facing changes"
     showcontent = true
 
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+  "setuptools >= 45",
+  "setuptools_scm[toml] >= 6.2",
```
webknjaz, line 52: I just realized that this PR is missing the bit allowing Git archive installs to work:
```suggestion
  # Essentials
  "setuptools >= 45",

  # Plugins
  "setuptools_scm[toml] >= 6.2",
  "setuptools_scm_git_archive>=1.1",
```
(it's basically a part of the usual boilerplate)

Another part of this would be adding `.git_archival.txt` with the following content:
```txt
ref-names: $Format:%D$
```

And `.gitattributes` containing:
```txt
# Force LF line endings for text files
* text=auto eol=lf

# Needed for setuptools-scm-git-archive
.git_archival.txt  export-subst
```

pyproject.toml
```diff
@@ -76,8 +82,10 @@ skip_glob = ["tests/fixtures/common/collections*"] # Skip ansible content due to
     ]
 
     [tool.pylint.master]
-    # tm_tokenize is virtually vendored and shouldn't be linted as such
-    ignore = "tm_tokenize"
+    ignore = [
+      "/src/ansible_navigator/_version.py", # built by setuptools_scm
```
webknjaz, line 86: Does it work with a leading slash? I know that it is significant in gitignore but not sure about this config.

src/ansible_navigator/__init__.py
webknjaz: I think that if we're committed to supporting this case, there should be a test added verifying that it keeps working.

src/ansible_navigator/__init__.py
cidrblock: Feel free to submit a follow on PR.  There is no technical need for this.

pyproject.toml
webknjaz: Teach me a better word and I'll use it instead :)

.gitignore
cidrblock: Ok.  Let's see if this is better....

.gitignore
webknjaz: Maybe we should generate some boilerplate from https://www.toptal.com/developers/gitignore at some point and replace bits of the old one (as a separate effort) — it emits those markers automatically.

src/ansible_navigator/__init__.py
webknjaz: The violation of the best practices may not be a technical reason but is still important nevertheless: https://stackoverflow.com/a/68502792/595220.

src/ansible_navigator/__init__.py
cidrblock: I agree that might be a valuable test to have, although it not not within the scope of this PR.

.gitignore
cidrblock: @webknjaz LMK if the .gitignore changes are acceptable.

webknjaz: Yes, I can live with it, I guess. I'd still prefer a separate gitignore file for folders that actually get checked in.

src/ansible_navigator/__init__.py
cidrblock: Feel free to submit a follow on PR. There is no technical need for this.

pyproject.toml
cidrblock: Ha!  No it doesn't.....I'm running from source without having built/installed navigator.

pyproject.toml
cidrblock: What's the use case here?  I'm not sure this should/needs to be done in this PR.

src/ansible_navigator/__init__.py
cidrblock: @webknjaz I am resolving this conversation because for this PR the login will stay in the __init__.py

pyproject.toml
webknjaz: It must be included. Otherwise, things like `pip install https://github.com/ansible/ansible-navigator/archive/refs/tags/2.0.0.tar.gz` will produce broken versions.

webknjaz (changes requested):
I'm blocking this because I believe that it's vital for the logic to stay out of `__init__.py`.

src/ansible_navigator/__init__.py
cidrblock: @webknjaz I am resolving this conversation, the use case for running from "source" is real and I've had to do it in the past to resolve share directory location issues across different operating systems and in instances when navigator is installed in system locations as root on macOS.  This PR will not include tests for that scenario.

cidrblock:
> I'm blocking this because I believe that it's vital for the logic to stay out of `__init__.py`.

This PR is being closed for now.

PR #999, opened 2022-02-23
Implement setuptools_scm

### LINKED ISSUE #553: [maintenance][packaging] Make `setuptools-scm` responsible for versioning

cidrblock:
Setup `setuptools_scm` such that is will build the `_version.py` file during the build process.

Accommodations for linting errors were made because although the `_version.py` file may not exist in source, once built locally, it will be subject to all enforced linters and formatters.

New files were documented based on my limited understanding of _exactly_ what is going on in the background here, in an effort to explain their existence to futre us.

This should allow for the removal of the setuptools_scm import in docs (https://github.com/ansible/ansible-navigator/blob/main/docs/conf.py#L10) which should in turn get us closer to removing the `ignore_imports` for mypy (https://github.com/ansible/ansible-navigator/blob/main/mypy.ini#L8)

This is related to #994, but a cleaner approach.

Closes #553


.flake8
```diff
@@ -53,6 +53,8 @@ per-file-ignores =
   # If other ignores are added for a specific file in the section following this,
   # these will need to be added to that line as well.
 
+  # Q000: Allow single-quotes in _version.py because setuptools_scm will create it with them.
```
webknjaz, line 56: (educational) FTR that template *could* be changed with our own. Or maybe even contributed to the default template upstream. And even include comment-based typing information.
Of course, this is probably not something to include in this PR. JFYI. Feel free to "resolve" once you see this.

.git_archival.txt
```diff
@@ -0,0 +1,8 @@
+# The name of this file is hardcoded in the setuptools_scm_git_archive
```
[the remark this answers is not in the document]
webknjaz: _I think_ this could be tested with `git tag -a v2.0.0 v2.0.0 && git archive && python -m pip install ./that/created/archive.tar.gz && python -m pip show ansible-navigator` (on a fresh detached git clone).

pyproject.toml
```diff
@@ -45,6 +45,16 @@
...
 
+[build-system]
+build-backend = "setuptools.build_meta"
+requires = [
+  # Essentials
+  "setuptools >= 45",
+
+  # Plugins
+  "setuptools_scm[toml] >= 6.2",
+  "setuptools_scm_git_archive>=1.1",
```
webknjaz, line 56: For some reason, the surrounding whitespaces are missing here:
```suggestion
  "setuptools_scm_git_archive >= 1.1",
```

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = ansible-navigator
-version = attr: ansible_navigator.__version__
+version = attr: ansible_navigator._version.version
```
webknjaz, line 3: This doesn't get used when `setuptools-scm` is integrated. It's common to just drop it.
```suggestion
```
Keeping it would only create confusion because of the illusion that it's how the version is retrieved.

src/ansible_navigator/_version.pyi
```diff
@@ -0,0 +1,16 @@
...
+
+Provides type hint PEP 484 information for the ``_version.py``
+file because it is generated by setuptools_scm and may
+not exist when type checking is don and will not
+container type hint information.
+"""
+
+from typing import Union
+
+from .configuration_subsystem import Constants
```
webknjaz, line 13: This seems inaccurate. I only see `Constants` in `cli.py` but not `_version.py`

src/ansible_navigator/cli.py
```diff
@@ -30,9 +31,9 @@
 
 
 try:
```
webknjaz, line 33: Maybe we need something like
```suggestion
__version__: Union[Constants, str]
try:
```
?

src/ansible_navigator/_version.pyi
```diff
@@ -0,0 +1,16 @@
+# fmt: off
+# isort: skip_file
```
webknjaz, line 2: Are these important? It seems odd that these formatters would attempt reformatting this file.

cidrblock: They both do, so I excluded both.

src/ansible_navigator/_version.pyi
cidrblock: yeah, mypy picks up it being assigned a Constant in cli.py

```
(venv) x1 ➜  ansible-navigator git:(scm_version) mypy
src/ansible_navigator/cli.py:36:19: error: Incompatible types in assignment (expression has type "Constants", variable has type "str")  [assignment]
        __version__ = Constants.NOT_SET
                      ^
```

src/ansible_navigator/cli.py
cidrblock: yeah, this is a better way to handle it on the ultimate assignment

src/ansible_navigator/_version.pyi
cidrblock: Not any more

src/ansible_navigator/_version.pyi
cidrblock: no longer relevant :)

webknjaz:
@cidrblock have you considered adding a change note for contributors? (`internal`)

cidrblock:
> @cidrblock have you considered adding a change note for contributors? (`internal`)

I did, but then thought this is really prep work until we get GH building and releasing it didn't really make an effective difference. to anyone...

webknjaz:
> I did, but then thought this is really prep work until we get GH building and releasing it didn't really make an effective difference to anyone...

Yeah, maybe. I wonder if it'd be important to mention in the docs, then.

PR #1001, opened 2022-02-23
Fix incorrect version reporting in changelogs

cidrblock:
By default, RTD does a fetch --limit 50, if a tag is not present in the last 50 commits, the version reported by setuptools_scm will be incorrect and appears as ``v0.1.dev...`` in the towncrier changelog.

Another approach is to enable ``DONT_SHALLOW_CLONE`` for the repo

https://docs.readthedocs.io/en/stable/feature-flags.html#feature-flags
This was done for ansible-navigator on the day of this commit.

The changes force a `git fetch --unshallow` https://github.com/pypa/setuptools_scm/blob/7950bdaa3e5d9da42434f3e7a85d05971ebeca1f/src/setuptools_scm/git.py#L110


webknjaz:
@cidrblock look at https://ansible-navigator--1001.org.readthedocs.build/en/1001/changelog/ — it does not have the tags fetched. IIRC this is fixable via the RTD config and needs to be included.

Also, the tox env needs to be changed to always recreate the tox virtualenv so that locally the version is updated.

.readthedocs.yml
```diff
@@ -1,15 +1,12 @@
 # https://docs.readthedocs.io/en/stable/config-file/v2.html
-
```
webknjaz, line 2: Maybe avoid including formatting changes?

.readthedocs.yml
```diff
@@ -20,6 +17,8 @@ formats: all
 
 python:
   install:
-  - method: setuptools
-    path: .
-  - requirements: docs/requirements.in
+    - method: pip
+      path: .
+    - method: setuptools
+      path: .
```
webknjaz: This shouldn't be necessary. `pip` will build and install it in the previous step. And at that point, all of the version metadata must already be present in the Git checkout.

webknjaz: Ah, wrong diff.

.readthedocs.yml
```diff
@@ -1,9 +1,12 @@
 # https://docs.readthedocs.io/en/stable/config-file/v2.html
-
 ---
-
 version: 2
 
+build:
```
webknjaz, line 5: Could you submit this in a separate PR?

docs/conf.py
```diff
@@ -36,16 +35,27 @@
...
-    get_scm_version(
-        local_scheme="no-local-version",
-    ).split(".")[:3],
-)
 
-# The full version, including alpha/beta/rc tags
-release = get_scm_version()
+# on_rtd is whether we are on readthedocs.org, this line of code grabbed from docs.readthedocs.org
+# https://docs.readthedocs.io/en/latest/faq.html?highlight=environ#how-do-i-change-behavior-for-read-the-docs
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
```
webknjaz, line 41: This shouldn't be necessary. Aim for having the correct metadata in the package instead. I think I solved that with https://github.com/ansible/pylibssh/blob/0610b59/.readthedocs.yml#L28 in the past.

setup.py
```diff
@@ -5,6 +5,7 @@
 
 import glob
 import os
+import shutil
```
webknjaz, line 8: This seems forgotten.
```suggestion
```

tox.ini
```diff
@@ -134,6 +134,7 @@ commands =
 isolated_build = true
 passenv =
   SSH_AUTH_SOCK
+recreate = true
```
webknjaz, line 137: (educational) Note that this will guarantee an up-to-date version in the metadata at the expense of populating a fresh virtualenv on each execution.

docs/conf.py
```diff
@@ -3,17 +3,16 @@
...
 """Configuration file for the Sphinx docs."""
 
-from functools import partial
+import os
+
+from importlib.metadata import version as get_version
 from pathlib import Path
 from sys import path
 
-from setuptools_scm import get_version
```
webknjaz, line 10: What is the reason you don't want to keep this? I think it's rather reasonable to have this as a docs dependency.

docs/conf.py
```diff
@@ -36,16 +35,27 @@
...
-)
 
-# The full version, including alpha/beta/rc tags
-release = get_scm_version()
+# on_rtd is whether we are on readthedocs.org, this line of code grabbed from docs.readthedocs.org
+# https://docs.readthedocs.io/en/latest/faq.html?highlight=environ#how-do-i-change-behavior-for-read-the-docs
+on_rtd = os.environ.get("READTHEDOCS", None) == "True"
+if on_rtd:
+    rtd_version = os.environ["READTHEDOCS_VERSION"]
+    release = rtd_version
```
webknjaz, line 44: Looks like one of the vars needs to be converted to a short version while the other should preserve the full one. The problem with this is that RTD does not know about those local and dev version details that setuptools-scm extracts from Git. So it will probably make it impossible to replicate the behavior exactly.

webknjaz: I think the commit hash is somehow exposed on RTD but it's still not the same — the version base that is supposed to be calculated from Git for intermediate (in-between-the-tags) commits is nowhere to be found.

webknjaz (changes requested):
> This is the suggested approach in the setuptools_scm project: [pypa/setuptools_scm#usage-from-sphinx](https://github.com/pypa/setuptools_scm#usage-from-sphinx)

The more I look into this the more I realize (and start remembering) that the suggested approach does not cover the use-cases we've got. The main one being the Towncrier drafts preview integration.

So if the only motivation behind is change is "because the README only has this example", I think that breaking the integration I've polished while implementing it across many projects is not worth it.

> This should also get us one step closer to removing the ignore_import flag for mypy

What is this about?

That said, I don't see how the suggested approach can realistically work for us. If there's some annoyance with typing, maybe just add an inline type-ignore for now and drop the rest?

cidrblock:
> What is this about?

Originally, yes, simply avoid a type ignore[import], but it doesn't appear this has ever worked correctly. So now it's a testing ground to see if it can be fixed.

tox.ini
cidrblock: "Also, the tox env needs to be changed to always recreate the tox virtualenv so that locally the version is updated."

I think I misunderstood the ask here...

docs/conf.py
```diff
@@ -8,12 +8,36 @@
...
+    present in the last 50 commits, the version reported by setuptools_scm
+    will be incorrect and appears as ``v0.1.dev...`` in the towncrier changelog.
+    Another approach is to enable ``DONT_SHALLOW_CLONE`` for the repo
+    https://docs.readthedocs.io/en/stable/feature-flags.html#feature-flags
+    This was done for ansible-navigator on the day of this commit.
+
+    :param args: The arguments
+    :param kwargs: The keyword arguments
+    :returns: The parsed version
+    """
```
webknjaz, line 31: Since it's not a lambda, we can properly declare the expectations here now:
```suggestion
    """
    assert 'pre_parse' not in kwargs
```

docs/conf.py
```diff
@@ -8,12 +8,36 @@
...
+
+def parse_with_fetch(*args, **kwargs) -> str:
+    """If the repo is found to be shallow, fetch a full.
+
+    By default, RTD does a fetch --limit 50, if a tag is not
+    present in the last 50 commits, the version reported by setuptools_scm
+    will be incorrect and appears as ``v0.1.dev...`` in the towncrier changelog.
+    Another approach is to enable ``DONT_SHALLOW_CLONE`` for the repo
+    https://docs.readthedocs.io/en/stable/feature-flags.html#feature-flags
+    This was done for ansible-navigator on the day of this commit.
```
webknjaz, line 26: I think this part should probably be in a code comment. WDYT?

docs/conf.py
```diff
@@ -8,12 +8,36 @@
...
+
+    :param args: The arguments
+    :param kwargs: The keyword arguments
+    :returns: The parsed version
+    """
+    return parse(*args, pre_parse=fetch_on_shallow, **kwargs)
+
+
+get_scm_version = partial(get_version, parse=parse_with_fetch)
```
webknjaz, line 35: Do we want to have this conditional for just RTD?

tox.ini
```diff
@@ -103,6 +103,9 @@ deps =
   # -c{toxinidir}/docs/requirements.txt
 description = Build The Docs
 changedir = {toxinidir}/docs
+# Reinstall so setuptools-scm runs and updates the _version.py
+# intentionally not in alphabetical order due to execution order
+commands_pre = {envpython} -m pip install --no-deps -e {toxinidir}
```
webknjaz, line 108: Also, given that the current implementation does not use the dist metadata, this isn't necessary, strictly speaking.

tox.ini
```diff
@@ -103,6 +103,9 @@ deps =
...
 commands =
   # Retrieve possibly missing commits:
   -git fetch --unshallow
```
webknjaz, line 111: With the same code in `conf.py`, this could be dropped given that this operation remains unconditional there.

webknjaz (commented):
So here's a few ideas to check out before merging.

docs/conf.py
cidrblock: interesting..........I guess good insurance

docs/conf.py
cidrblock: I was thinking what if someone somehow managed to do this locally....what would the harm be, also, if the experimental func ever changes or acts up, it should be able to repro locally w/o having to set the magic RTD envvar

tox.ini
cidrblock: True, but the tox env not being up to date seems odd, what I really should add is `git clean -iX -- docs/pkg/` because I've been bitten by that with failed tox docs builds a bunch

tox.ini
cidrblock: right!

cidrblock: know if this will gets the tags too?
```
--unshallow
If the source repository is complete, convert a shallow repository to a complete one, removing all the limitations imposed by shallow repositories.

If the source repository is shallow, fetch as much as possible so that the current repository has the same history as the source repository.
```

docs/conf.py
cidrblock: I think the explanation is helpful in the docstring since it's such an odd thing
image

webknjaz: Ah, okay. Usually, the motivation (the "why") goes into comments but the "how"/"what" goes into docstrings.

docs/conf.py
webknjaz: Agreed. Just making sure all the questions are asked and answered :)

docs/conf.py
webknjaz: Not just insurance but also expectation declaration per contract programming principles :)

webknjaz (approved):
Brilliant! Thanks for taking care of this!

tox.ini
webknjaz: True, that is a great candidate for the next PR :)

tox.ini
webknjaz: Not sure. I've been using the two commands for years so maybe it wasn't true at the time I first introduced it.

cidrblock:
> Brilliant! Thanks for taking care of this!

Couldn't have done it w/o your help.

PR #1002, opened 2022-02-23
Reorder import of key_value_store

cidrblock:
The file only exists when running in an EE, not during type checking.

Not previously caught as we still ignore import errors, this is another small fix on the path to remove that.

### DIFF
--- a/share/ansible_navigator/utils/catalog_collections.py
+++ b/share/ansible_navigator/utils/catalog_collections.py
@@ -14,7 +14,6 @@ from collections import OrderedDict
 from datetime import datetime
 from json.decoder import JSONDecodeError
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


relrod (approved):

share/ansible_navigator/utils/catalog_collections.py
```diff
@@ -30,14 +31,18 @@
...
 try:
-    from key_value_store import KeyValueStore
-except ImportError:
     from ansible_navigator.utils.key_value_store import KeyValueStore
+except ImportError:
+    if not TYPE_CHECKING:
+        from key_value_store import KeyValueStore
 
 
-# pylint: enable=import-error
```
relrod, line 39: 🔭 🤣

PR #1004, opened 2022-02-24
Centralize json dump and dumps

### LINKED ISSUE #1003: Unicode characters are not written to a file when using `:json`
Related #944 

When the Unicode character support went in, it only accounted for presenting unicode characters in the TUI but not when writing to files.

cidrblock:
Centralizes the json dumping into one location, to prevent issues like the below and will allow for a different encoder in one place.

Fixes #1003 

(yaml can be moved into this file as well to centralize all serializers)

### DIFF
--- a/src/ansible_navigator/utils/serialize.py
+++ b/src/ansible_navigator/utils/serialize.py
@@ -1,0 +1,36 @@
+"""Abstractions for common serialization formats."""
+
+from functools import lru_cache
+import json
+from typing import Any
+from typing import IO
+from typing import NamedTuple
+
+
+class JsonParams(NamedTuple):
+    """The parameters for json dump and dumps."""
+
+    indent: int = 4
+    sort_keys: bool = True
+    ensure_ascii: bool = False
+
+
+def json_dump(dumpable: Any, file_handle: IO, params: NamedTuple = JsonParams()) -> None:
+    """Serialize and write the dumpable to a file.
+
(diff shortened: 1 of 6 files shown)


src/ansible_navigator/utils/serialize.py
```diff
@@ -0,0 +1,36 @@
...
+
+class JsonParams(NamedTuple):
+    """The parameters for json dump and dumps."""
+
+    indent: int = 4
+    sort_keys: bool = True
+    ensure_ascii: bool = False
+
+
+def json_dump(dumpable: Any, file_handle: IO, params: NamedTuple = JsonParams()) -> None:
```
relrod, line 18:
```python
Any
```

🤮

src/ansible_navigator/utils/serialize.py
```diff
@@ -0,0 +1,36 @@
...
+    """Serialize and write the dumpable to a file.
+
+    :param dumpable: The object to serialize
+    :param file_handle: The file handle to write to
+    :param params: Parameters to override the defaults
+    """
+    json.dump(dumpable, file_handle, **params._asdict())
+
+
+def json_dumps(dumpable: Any, params: NamedTuple = JsonParams()) -> str:
```
relrod, line 28:
```python
Any
```

🤮

relrod (approved):

PR #1009, opened 2022-02-25
Make Sphinx builds reproducible with a lockfile

webknjaz:
This patch just adds and integrates a pip-tools-generated constraints file w/o adding any of the pip-tools machinery at the moment.


cidrblock:
I don't know........................

It doesn't seem like there is a good answer here, either we accept the occasion broken something due to a version or go down this road.

My concern here is we don't have a process in place or tooling to update/audit this list...

and probably don't have the time to implement something right now

thinking

webknjaz:
Tooling is `pip install pip-tools`. Nothing else is needed strictly speaking.

ssbarnea (commented):
While I am in favour of having test dependencies lock files I slowly went against having more than one lock file per project.

I state this based on maintenance experience with projects already having this, where I ended up having to replace multi locks with a single one.

I see no reason why docs requirements will ever be allowed to conflict with project test ones and I prefer being able to install all testing deps (including docs) in a single development environment.

Over the last year I faced multiple problems with dependency updating when having multiple lock files, so I am against. 

But I am in favour of using one constraints.txt file per project, navigator has none yet.

Out of curiosity, what was the reason for creating this change? Did you see a recent docs build failure on navigator?

This change prevent this.

webknjaz:
This is a long-planned change. I postponed making a lockfile for docs when adding the initial docs configuration. Also, per experience, it's best to have non-conflicting lockfiles for each env. But you could have an additional file combining them. I only care that the per-env ones exists, I have a PoC of an automation to batch-update them so it's not a concern. Although, I don't mean to include it in this PR.

PR #1013, opened 2022-02-25
Add yaml to prettier

cidrblock:
Also- limit the file types to only those we care about rather than everything.


cidrblock:
pre-commit.ci run

cidrblock:
pre-commit.ci run

cidrblock:
pre-commit.ci autofix

PR #1014, opened 2022-02-26
`dataclass` serialization capabilities (`:settings` related)

cidrblock:
- Add new dataclass base class for content moving forward
- ContentBase has default dictionary factories which can be overridden per serf/view
- Added pytest-subtests

This is not in use yet, be will be for `:settings`

The last thing to do is to add tests demonstrating full vs. normal views, but I thought I'd put it out for review now.

The one thing I don't like and would appreciate suggestions with is the need for `TYPE_CHECKING`


src/ansible_navigator/utils/serialize_defs.py
```diff
@@ -0,0 +1,10 @@
+"""Definitions related to content serialization."""
```
alisonlhart, line 1: Out of curiosity, is this needed in its own file vs. having it in ./serialize.py?

alisonlhart: Ah, I found the answer to my question immediately. I see the import in other files. Carry on.

cidrblock: _This_ is a great question, which is where I need a little help.  The definition was moved out of serialize.py, as part of an effort to avoid the circular import when type_checking......, although it didn't eliminate it I left it there.

I don't think the TYPE_CHECKING condition is necessarily a bad thing, but it does make me wonder if there isn't a better way to structure everything....

@relrod @alisonlhart If you have any ideas about a better layout, I'm listening...

src/ansible_navigator/ui_framework/content_defs.py
```diff
@@ -0,0 +1,90 @@
...
+        serf: SerializationFormat,
+        view: ContentView,
+    ) -> DictT:
+        """Convert thy self into a dictionary.
+
+        :param serf: The serialization format
+        :param view: The content view
+        :returns: A dictionary created from self
+        """
+        if (view, serf) == (ContentView.FULL, SerializationFormat.JSON):
```
webknjaz, line 54: Please invent a better name for that serf thing.

src/ansible_navigator/ui_framework/content_defs.py
```diff
@@ -0,0 +1,90 @@
...
+            return self.serialize_json_full()
+        if (view, serf) == (ContentView.FULL, SerializationFormat.YAML):
+            return self.serialize_yaml_full()
+        if (view, serf) == (ContentView.NORMAL, SerializationFormat.JSON):
+            return self.serialize_json_normal()
+        if (view, serf) == (ContentView.NORMAL, SerializationFormat.YAML):
+            return self.serialize_yaml_normal()
+        return asdict(self)
```
webknjaz, line 62: I suggest having a map instead of multiple if-blocks:
```suggestion
        converter_map = {
            (ContentView.FULL, SerializationFormat.JSON): self.serialize_json_full,
            (ContentView.FULL, SerializationFormat.YAML): self.serialize_yaml_full,
            (ContentView.NORMAL, SerializationFormat.JSON): self.serialize_json_normal,
            (ContentView.NORMAL, SerializationFormat.YAML): self.serialize_yaml_normal,
        }

        try:
            dump_self_as_dict = converter_map[view, serf]
        except KeyError:
            return asdict(self)
        else:
            return dump_self_as_dict()
```

src/ansible_navigator/ui_framework/content_defs.py
cidrblock: Thanks. Will do. naming is hard.

tests/unit/utils/test_serialize_dataclass.py
```diff
@@ -0,0 +1,212 @@
...
+            serialization_format=SerializationFormat.JSON,
+        )
+        result_template = (
+            '{{\n    "attr_01": "False__j_{view}",\n'
+            '    "attr_02": "2__j_{view}",\n'
+            '    "attr_03": "three__j_{view}"\n}}'
+        )
+        assert result == result_template.format(view=content_view)
+
+    @staticmethod
```
webknjaz, line 197: This is extremely weird and unnatural in the pytest ecosystem. It makes no sense to nest the tests under a class. Just make pure functions instead.

tests/unit/utils/test_serialize_dataclass.py
```diff
@@ -0,0 +1,212 @@
...
+        :param content_view: The content view
+        :param serialization_tuple: The suffix, serialization format
+        """
+        content = ContentTestOverride()
+        result = content.asdict(
+            content_view=content_view,
+            serialization_format=serialization_tuple[1],
+        )
+        for _key, value in result.items():
+            assert value.endswith(f"_{serialization_tuple[0]}_{str(content_view)}")
```
webknjaz, line 176: for such cases, it's customary to use `pytest-subtest` to see all of the iterations show up separately in the report.

tests/unit/utils/test_serialize_dataclass.py
cidrblock: [@pytest.mark.parametrize](https://docs.pytest.org/en/latest/how-to/parametrize.html#pytest-mark-parametrize) allows one to define multiple sets of arguments and fixtures at the test function or class.

Seems documented: https://docs.pytest.org/en/latest/how-to/parametrize.html

tests/unit/utils/test_serialize_dataclass.py
cidrblock: nice TIL, added...

tests/unit/utils/test_serialize_dataclass.py
webknjaz: It does. But the whole "put pure functions under a class" part is pointless.

tests/unit/utils/test_serialize_dataclass.py
```diff
@@ -0,0 +1,214 @@
...
+        assert result == result_template.format(view=content_view)
+
+    @staticmethod
+    def test_content_to_yaml_override(content_view: ContentView):
+        """Test the conversion of the dataclass with overrides to ``YAML``.
+
+        :param content_view: The content view
+        """
+        content = ContentTestOverride()
+        result = serialize(
```
webknjaz, line 206: What is result?

src/ansible_navigator/utils/serialize.py
```diff
@@ -26,6 +41,37 @@
 # pylint: enable=unused-import
 
 
+def serialize(
+    content_view: "ContentView",
+    content: "ContentBase",
+    serialization_format: SerializationFormat,
+    file_mode: str = "w",
+    filename: str = None,
```
webknjaz, line 49: Why build APIs around strings when you can use Pathlib-like interfaces? (with things supporting `__fspath__()`)

tests/unit/utils/test_serialize_dataclass.py
cidrblock: ok, it was just convenient :)

cidrblock: done

src/ansible_navigator/ui_framework/content_defs.py
cidrblock: done

tests/unit/utils/test_serialize_dataclass.py
cidrblock: I can't tell if you are joking here.................So I'll answer

image

src/ansible_navigator/utils/serialize.py
cidrblock: It was built like this so we could transition the consumers of it across the project without addressing the fact the file is a string elsewhere..........but, I can add a Union here so it doesn't propagate the use of a str moving forward.

cidrblock: Add path.

Related: https://github.com/ansible/ansible-navigator/issues/697

tests/unit/utils/test_serialize_dataclass.py
webknjaz: yep. stdlib unittest also has this feature, if you ever need it.

tests/unit/utils/test_serialize_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+        content=content,
+        content_view=content_view,
+        serialization_format=SerializationFormat.YAML,
+    )
+    expected = "---\nattr_01: false\nattr_02: 2\nattr_03: three\n"
+    assert result == expected
+
+
+@pytest.mark.parametrize(**ParametrizeFormat()._asdict())
+@pytest.mark.parametrize(**ParametrizeView()._asdict())
```
webknjaz, line 160: I see this is repeated a few times so for such cases, consider assigning the mark to a variable and decorate with that:
```python
with_view_params = pytest.mark.parametrize(**ParametrizeView()._asdict())  # <-- put this somewhere above the test declarations
```
```suggestion
@with_view_params
```

src/ansible_navigator/utils/serialize.py
```diff
@@ -63,7 +122,11 @@ class YamlStyle(NamedTuple):
     allow_unicode: bool = True
 
 
-def human_dump(obj: Any, filename: str = None, file_mode: str = "w") -> Optional[str]:
+def human_dump(
+    obj: Any,
+    filename: Optional[Union[Path, str]] = None,
```
webknjaz, line 127: I'm almost sure that `Optional` is implicit when the default is `None`. So this should be enough:
```suggestion
    filename: Union[Path, str] = None,
```

src/ansible_navigator/utils/serialize.py
```diff
@@ -26,6 +43,37 @@
...
+    """Serialize a dataclass based on format and view.
+
+    :param content_view: The contest view
+    :param content: The content dataclass to serialize
+    :param serialization_format: The serialization format
+    :param file_mode: The mode for the file operation
+    :param filename: A filename to write to
+    :returns: The serialized content
+    """
+    content_as_dict = content.asdict(
```
webknjaz, line 62: While the public API is permissive and accepts both `str` and `Path`, the internals don't need to work with `str`. So make sure it's path-like right away:
```suggestion
    if filename is not None:
        filename = Path(filename)
    content_as_dict = content.asdict(
```
Then, in the converters, only work with `Path`.

src/ansible_navigator/utils/serialize.py
```diff
@@ -34,6 +82,17 @@ class JsonParams(NamedTuple):
     ensure_ascii: bool = False
 
 
+def _json_dump(dumpable: Dict, filename: Union[Path, str], file_mode: str):
```
webknjaz, line 85: Per comment above, always use `Path` internally:
```suggestion
def _json_dump(dumpable: Dict, filename: Path, file_mode: str):
```

src/ansible_navigator/utils/serialize.py
```diff
@@ -34,6 +82,17 @@ class JsonParams(NamedTuple):
...
+    """Create a file handle and dump json.
+
+    :param dumpable: The object to dump
+    :param filename: The file name for the file
+    :param file_mode: The file mode for the operation
+    """
+    with open(file=filename, mode=file_mode, encoding="utf-8") as file_handle:
```
webknjaz, line 92: And having it always as a `Path`, you can call it's methods instead of having to work with low-level str-based APIs:
```suggestion
    with filename.open(mode=file_mode, encoding="utf-8") as file_handle:
```

tests/unit/utils/test_serialize_dataclass.py
cidrblock: Wow.... Hadn't considered that... I'll do this.

cidrblock: So much cleaner :)

src/ansible_navigator/utils/serialize.py
cidrblock: You are 100% right, but it is discouraged:

"This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit."

https://www.python.org/dev/peps/pep-0484/#union-types

cidrblock: We'll enforce it at some point: https://github.com/ansible/ansible-navigator/issues/1048

relrod: :nauseated_face: implicit Optional sounds horrible. Please keep the `Optional` here.

(Also I tested locally and mypy fails without explicit Optional, as it should.)

I'm fine with type inference. I'm less fine with behind the scenes magic.

src/ansible_navigator/utils/serialize.py
cidrblock: If we change the signature of human dump to Optional[Path] it will cause typing issues elsewhere in the code:

```
src/ansible_navigator/actions/open_file.py: note: In member "run" of class "Action":
src/ansible_navigator/actions/open_file.py:109:50: error: Argument "filename" to "human_dump" has incompatible type "str"; expected "Optional[Path]"  [arg-type]
```

That said, If we leave human_dump as Path or str, we can make serialize Optional[Path].  This will require the caller be updated to a Path before they migrate to serialize, if they write to a file, which is not all that common.

It just raises the priority of this a little: https://github.com/ansible/ansible-navigator/issues/697

src/ansible_navigator/utils/serialize.py
cidrblock: yep, nice.  Won't be able to do it for yaml yet.

src/ansible_navigator/ui_framework/content_defs.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class ContentView(Enum):
+    """The content view."""
+
+    FULL = "full"
+    NORMAL = "normal"
+
+
+DictValueT = TypeVar("DictValueT")
```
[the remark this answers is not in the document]
cidrblock: Changed and noted. Pylint will complain about the useless suppression for us later

tests/unit/utils/test_serialize_dataclass.py
cidrblock: I think I misunderstood here, I believe you were suggesting a better naming convention for the variables....

Updated.

src/ansible_navigator/utils/serialize.py
cidrblock: It is enforced now: https://github.com/ansible/ansible-navigator/pull/1050

thedoubl3j (approved):
looks great, excited to see more dataclass stuff and it becoming more prevalent for actions

src/ansible_navigator/utils/serialize.py
```diff
@@ -26,6 +43,37 @@
...
+def serialize(
+    content_view: "ContentView",
+    content: "ContentBase",
+    serialization_format: SerializationFormat,
+    file_mode: str = "w",
+    filename: Optional[Path] = None,
+) -> Optional[str]:
+    """Serialize a dataclass based on format and view.
+
+    :param content_view: The contest view
```
ganeshrn, line 55: Type `s/contest/content`

ganeshrn (approved):

goneri:
This break the network-ee jobs:

```
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/bin/ansible-navigator", line 5, in <module>
    from ansible_navigator.cli import main
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/cli.py", line 20, in <module>
    from .action_runner import ActionRunner
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/action_runner.py", line 19, in <module>
    from ansible_navigator.actions import kegexes
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/actions/__init__.py", line 22, in <module>
    from ..ui_framework import Interaction
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/ui_framework/__init__.py", line 4, in <module>
    from .content_defs import ContentBase
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/ui_framework/content_defs.py", line 10, in <module>
    from ..utils.compatibility import TypeAlias
  File "/home/zuul/src/github.com/ansible/network-ee/.tox/integration/lib/python3.8/site-packages/ansible_navigator/utils/compatibility.py", line 12, in <module>
    from typing_extensions import TypeAlias
ModuleNotFoundError: No module named 'typing_extensions'
```

PR #1024, opened 2022-02-28
Define `CursesLine` and `CursesLines` as NewType (`:settings` related)

cidrblock:
Allows them to be used as if they were an instantiated class rather than referred to as an alias.
It cleans up the definitions and return values.

Minimal code change were made, limited to a var name change for consistency or the way the return value was generated.

TIL, Changed in version 3.10: NewType is now a class rather than a function. There is some additional runtime cost when calling NewType over a regular function. However, this cost will be reduced in 3.11.0.

https://docs.python.org/3/library/typing.html#newtype


thedoubl3j (approved):
+1

ganeshrn (approved):

PR #1029, opened 2022-03-01
Move pull_policy to pull.policy

### LINKED ISSUE #951: `pull-policy` should be a dictionary in the settings (`pull: policy: ""`)

cidrblock:
This moves pull policy under the pull key in the settings file.

This will result in the next release being a major as it is a breaking change.

Fixes #951

### DIFF
--- a/docs/changelog-fragments.d/1029.breaking.md
+++ b/docs/changelog-fragments.d/1029.breaking.md
@@ -1,0 +1,12 @@
+Moved the configuration of ``pull-policy`` under the ``pull`` key
+in the settings file.
+
+Here is a sample:
+```yaml
+ansible-navigator
+  execution-environment:
+    pull:
+      policy: never
+```
+
+-- by {user}`cidrblock`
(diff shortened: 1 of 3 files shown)


ganeshrn:
ansible-language-server followed the similar for pull-policy as that of ansible-navigator. 
https://github.com/ansible/ansible-language-server#language-server-settings
Should we do a similar change in ALS as well? It will require a major version change for both ALS and extension.

ganeshrn (approved):

cidrblock:
> ansible-language-server followed the similar for pull-policy as that of ansible-navigator. https://github.com/ansible/ansible-language-server#language-server-settings Should we do a similar change in ALS as well? It will require a major version change for both ALS and extension.

I'd say if we were going to do it, we should do it fast before there is broad adoption of using EEs with both.  I'm guessing it is still low.

Maybe we could time it with a navigator release and the lint 6 release?

PR #1031, opened 2022-03-02
Add ability to transform settings to a presentable dataclass (`:settings` related)

cidrblock:
Picked from #957 

Add the ability to transform the current settings into human readable, mostly flattened objects that can be presented in the UI.

- Constants are resolved to their values
- A psuedo entry is created for the settings file
- Unit tests added for the xform

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/defs_presentable.py
+++ b/src/ansible_navigator/configuration_subsystem/defs_presentable.py
@@ -15,6 +15,7 @@ from typing import Union
 from .definitions import CliParameters
 from .definitions import Constants as C
 from .definitions import SettingsEntry
+from .navigator_configuration import Internals
 from .utils import create_settings_file_sample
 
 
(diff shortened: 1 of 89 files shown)


src/ansible_navigator/configuration_subsystem/defs_presentable.py
```diff
@@ -0,0 +1,208 @@
...
+    @property
+    def current(self) -> str:
+        """Convert the current value into a string.
+
+        :returns: The current value as a string
+        """
+        return str(self.current_value)
+
+    def get(self, attribute: str):
+        """Allow this dataclass to be treated like a dictionary.
```
relrod, line 97: Except this will throw if the key doesn't exist, and `dict#get` will return `None`.

cidrblock: That was intentional, because we shouldn't be getting attributes from a data that don't exist.

Can confirm when the settings PR goes in, but I can't think of anyplace where we require or have a hard coded key/attribute name required in something headed to the UI.

cidrblock: (it's also noted in the description below)

thedoubl3j (approved):
`defs_presentable` still feels off in naming to me but not a hill I am willing to die on. lgtm

ganeshrn (approved):

PR #1033, opened 2022-03-02
Convert percentages, dataclass capable (`:settings` related)

cidrblock:
Depends on #1014 

The `Any` here should be replaced with `ContentBase` when it merges.

This is a rewrite of the convert percentages function to account for the data being either a dictionary or dataclass.

New tests to confirm.

### DIFF
--- a/src/ansible_navigator/ui_framework/utils.py
+++ b/src/ansible_navigator/ui_framework/utils.py
@@ -10,9 +10,11 @@ from typing import Dict
 from typing import List
 from typing import Union
 
+from .content_defs import ContentBase
+
 
 def convert_percentage(
-    content: Union[Dict[str, Any], Any],
+    content: Union[Dict[str, Any], ContentBase],
     columns: List[str],
     progress_bar_width: int,
 ) -> None:
--- a/src/ansible_navigator/ui_framework/utils.py
+++ b/src/ansible_navigator/ui_framework/utils.py
@@ -3,28 +3,48 @@
 import functools
 import re
 
+from dataclasses import is_dataclass
 from math import floor
+from typing import Any
+from typing import Dict
 from typing import List
+from typing import Union
 
 
-def convert_percentage(data_dict: dict, keys: List, progress_bar_width: int) -> None:
-    """convert a string % to a little progress bar
-    not recursive
-    80% = 80%|XXXXXXXX  |
+def convert_percentage(
+    content: Union[Dict[str, Any], Any],
+    columns: List[str],
+    progress_bar_width: int,
+) -> None:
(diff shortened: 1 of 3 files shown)


src/ansible_navigator/ui_framework/utils.py
```diff
@@ -3,28 +3,48 @@
...
 from typing import List
+from typing import Union
 
 
-def convert_percentage(data_dict: dict, keys: List, progress_bar_width: int) -> None:
-    """convert a string % to a little progress bar
-    not recursive
-    80% = 80%|XXXXXXXX  |
+def convert_percentage(
+    content: Union[Dict[str, Any], Any],
```
relrod, line 15: This can simplify to `Any` for now, since `A ∪ ⊤` ~ `T`

(where T is the [top type](https://en.wikipedia.org/wiki/Top_type))

src/ansible_navigator/ui_framework/utils.py
```diff
@@ -3,28 +3,48 @@
...
-                data_dict[key] = "COMPLETE".center(progress_bar_width, " ")
-            else:
-                chars_in_progress_bar = floor(progress_bar_width / 100 * int(value[0:-1]))
-                data_dict["_" + key] = value
-                data_dict[key] = ("\u2587" * chars_in_progress_bar).ljust(progress_bar_width)
+            new_value = _string_to_progress(value, progress_bar_width)
+            if isinstance(content, dict):
+                content["_" + column] = value
+                content[column] = new_value
+            elif is_dataclass(content):
```
relrod, line 32: What happens if neither of these cases is true, since the type currently says that can happen?

src/ansible_navigator/ui_framework/utils.py
cidrblock: The Any there is just a place holder for ContentBase in #1014

src/ansible_navigator/ui_framework/utils.py
cidrblock: The Any there is just a place holder for ContentBase in https://github.com/ansible/ansible-navigator/pull/1014

At that point, mypy should protect it.  If something did slip through one of a few things would happen:

1) There would be a traceback because the object didn't have a `get` function
2) There would be traceback because the object didn't have a `_` attribute corresponing to the column name that had a %
3) The object would not be modified if it did not have any `%` strings, did have a get.

thedoubl3j (approved):
lgtm, first time I have ever seen an assert for a progress bar. thought my browser page had done something funky there

ganeshrn (approved):

PR #1036, opened 2022-03-02
Switch from `SimpleNamespace` to `dataclass` in command runner

cidrblock:
Just to ensure the data structure is not abused in the future.

### DIFF
--- a/src/ansible_navigator/command_runner/command_runner.py
+++ b/src/ansible_navigator/command_runner/command_runner.py
@@ -13,7 +13,7 @@ from typing import Union
 PROCESSES = (multiprocessing.cpu_count() - 1) or 1
 
 
-@dataclass
+@dataclass(frozen=False)  # updated with results after run
 class Command:
     """command obj"""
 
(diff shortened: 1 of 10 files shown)


src/ansible_navigator/command_runner/command_runner.py
```diff
@@ -12,15 +13,16 @@
 PROCESSES = (multiprocessing.cpu_count() - 1) or 1
 
 
-class Command(SimpleNamespace):
+@dataclass
```
relrod, line 16: Can we freeze it?

cidrblock: It gets updated after it is instantiated with the outcome of the command:
```
def run_command(command: Command) -> None:
    """run a command"""
    try:
        proc_out = subprocess.run(
            command.command,
            stdout=subprocess.PIPE,
            stderr=subprocess.PIPE,
            check=True,
            universal_newlines=True,
            shell=True,
        )
        command.stdout = proc_out.stdout
    except subprocess.CalledProcessError as exc:
        command.stderr = str(exc.stderr)
```

cidrblock: Now noted in the desc with frozen=False.

relrod (approved):

PR #1037, opened 2022-03-02
Switch from `SimpleNamespace` to `dataclass` in puller

cidrblock:
Just to ensure the data structure is not abused in the future.

### DIFF
--- a/src/ansible_navigator/image_manager/puller.py
+++ b/src/ansible_navigator/image_manager/puller.py
@@ -3,7 +3,7 @@ import logging
 import shlex
 import subprocess
 
-from types import SimpleNamespace
+from dataclasses import dataclass
 from typing import List
 from typing import Union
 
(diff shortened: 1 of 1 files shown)


src/ansible_navigator/image_manager/puller.py
```diff
@@ -14,7 +14,8 @@
 from ..utils.functions import shlex_join
 
 
-class ImageAssessment(SimpleNamespace):
+@dataclass
```
relrod, line 17: Freeze? Or does it get mutated?

cidrblock: It gets incrementally updated as the process goes:

e.g.:
`self._assessment.pull_required = pull`

cidrblock: Now noted in the description with frozen = False

relrod (approved):

PR #1038, opened 2022-03-02
Switch from `SimpleNamespace` to frozen `dataclass` in defintions

cidrblock:
This not not need to be mutated.


src/ansible_navigator/configuration_subsystem/definitions.py
```diff
@@ -131,7 +130,8 @@ def settings_file_path(self, prefix: str) -> str:
         return sfp.replace("_", "-")
 
 
-class SubCommand(SimpleNamespace):
+@dataclass(frozen=True)
+class SubCommand:
     """An object to hold a subcommand"""
 
     name: str
```
relrod, line 137: Can we fix the `Union[None, str]` -> `Optional[str]` down on 139 while we're at it?

relrod (approved):

src/ansible_navigator/configuration_subsystem/definitions.py
relrod: Approved either way

PR #1041, opened 2022-03-02
Test auto-update outstanding PRs

cidrblock:

PR #1044, opened 2022-03-02
Rename action, add label

cidrblock:
Related #1041


cidrblock:
Related to #1041, will be removing

PR #1055, opened 2022-03-04
typing-extensions is a strong dependency now

goneri:
`typing-extensions` is a strong dependency as soon as you don't run Python 3.10.

See: https://github.com/ansible/ansible-navigator/pull/1014#issuecomment-1059339161


setup.cfg
```diff
@@ -41,6 +41,8 @@ install_requires =
     pyyaml
     importlib-resources; python_version < "3.9.0"
     dataclasses; python_version < "3.7.0"
+    # used for TypeAlias, can be removed once Python 3.10 is required
+    typing-extensions;python_version < '3.10'
```
cidrblock, line 45:
```suggestion
    typing-extensions; python_version < '3.10'
```

requirements.txt
```diff
@@ -1 +0,0 @@
-.
```
cidrblock, line 1: This one needs to stay, as it serves as the proxy to setup.cfg...

cidrblock:
Hey @goneri Thanks for this.........  I can't really think of a way around it unless we `if TYPE_CHECKING` everywhere....... which.....  I think we porbably do, so we don't have a runtime dep.  @relrod Thoughts?

requirements.txt
goneri: My bad. Let's get this fixed.

setup.cfg
goneri: I removed the space on purpose to match the format in the example: https://www.python.org/dev/peps/pep-0508/#environment-markers

But I can reverse that.

cidrblock:
Hey @goneri , I'll do a PR to limit the need for that to just TYPE_CHECKING, it's my mess I can clean it up.  Give me 10 minutes

goneri:
Great, thanks.

PR #1057, opened 2022-03-04
Test basic stdout functionality without test-requirements

cidrblock:
Run some basic navigator commands without having installed the test requirements.

We'll run across all the python version just in case, the tests only take < a minute to run.  Locally it is 4 seconds because they  are multi threaded.

This should flesh out issues like the one found in #1056.  Once we have better support for collections/images/settings with stdout, those can be added here.

One small issue in Command runner fixed, capturing the stdout in the exception.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -93,7 +93,7 @@ setenv =
   {[testenv]setenv}
 
 [testenv:no-test-deps]
-description = Run unittes under {basepython} ({envpython})
+description = Run smoke(unittest) under {basepython} ({envpython})
 deps =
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -92,6 +92,15 @@ deps =
 setenv =
   {[testenv]setenv}
 
+[testenv:no-test-deps]
+description = Run unittes under {basepython} ({envpython})
```
relrod, line 96:
```suggestion
description = Run unittest under {basepython} ({envpython})
```

relrod (approved):

PR #1061, opened 2022-03-07
Fix EE volume mount issue

cidrblock:


src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -380,60 +473,59 @@ def execution_environment_volume_mounts(
...
-                        f" entry: '{mount_obj}'. Value of '{str(exc)}' key not provided."
+                        f"The following {settings_entry} entry could not be parsed:"
+                        f" {list_entry} ({entry.value.source.value}), {str(exc)}"
                     )
                     exit_messages.append(ExitMessage(message=exit_msg))
-                    exit_hint_msg = (
-                        " Valid keys are 'src', 'dest' and 'label'. Note: label key is optional."
+                    exit_msg = (
+                        "The value of execution-environment.volume-mounts"
+                        + "should be a list of dictionaries"
```
relrod, line 495:
```suggestion
                        + " should be a list of dictionaries"
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,34 +74,111 @@ class VolumeMountOption(Enum):
...
-    src: str
-
-    #: The destination path in the container for the volume mount.
-    dest: str
-
-    #: Options for the bind mount.
+    fs_destination: str
+    """The destination file system path in the container for the volume mount"""
+    fs_source: str
+    """The source file system path of the volume mount"""
```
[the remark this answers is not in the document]
cidrblock: they are in alphabetical order... it's easier for my brain....

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,34 +74,111 @@ class VolumeMountOption(Enum):
...
+        :returns: A populated volume mount
+        """
+        options = dictionary.get("label", "")
+        if options:
+            option_list = [
+                const
+                for const in VolumeMountOption
+                for option in options.split(",")
+                if const.value == option
+            ]
```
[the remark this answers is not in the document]
cidrblock: I think gettattr will be better, pulling it out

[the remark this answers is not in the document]
cidrblock: yanked it

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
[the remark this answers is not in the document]
cidrblock: I'll need to remove the default value for options.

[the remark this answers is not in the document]
cidrblock: Done

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
+        # Validate and populate _options
+        if isinstance(self.options_string, str):
+            if not self.options_string == "":
+                options = []
+                option_values = [o.value for o in VolumeMountOption]
+                for option in self.options_string.split(","):
+                    if option not in option_values:
+                        errors.append(
+                            f"Unrecognized label: '{option}',"
+                            " available labels include {oxfordcomma(option_values)}.",
```
relrod, line 131: I still don't get why we're calling these labels. They're options:

- [`podman run` manpage](https://docs.podman.io/en/latest/markdown/podman-run.1.html#volume-v-source-volume-host-dir-container-dir-options)
- [`docker run` docs](https://docs.docker.com/engine/reference/run/#volume-shared-filesystems)

Some of the options happen to control relabeling behavior. But they aren't labels.

```suggestion
                            f"Unrecognized option: '{option}',"
                            " available options include {oxfordcomma(option_values)}.",
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -354,86 +404,101 @@ def execution_environment_volume_mounts(
...
-            C.ENVIRONMENT_VARIABLE,
-            C.USER_CLI,
-        ]:
-            volume_mounts = flatten_list(entry.value.current)
-            for mount_path in volume_mounts:
-                parts = mount_path.split(":")
-                if len(parts) > 3:
+        entry_name = entry.settings_file_path(prefix="")
+        entry_source = entry.value.source
+        volume_mounts: List[VolumeMount] = []
```
relrod, line 421: Should this be a set?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -354,86 +404,101 @@ def execution_environment_volume_mounts(
...
+
+        if entry_source in (C.ENVIRONMENT_VARIABLE, C.USER_CLI):
+            hint = (
+                "Try again with format <source-path>:<destination-path>:<labels>'."
+                " Note: label is optional."
```
relrod, line 426:
```suggestion
                "Try again with format <source-path>:<destination-path>:<options>'."
                " Note: options may be excluded."
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -354,86 +404,101 @@ def execution_environment_volume_mounts(
...
+            )
+            mount_strings = flatten_list(entry.value.current)
+            for mount_str in mount_strings:
+                src, dest, labels, *left_overs = chain(mount_str.split(":"), repeat("", 3))
```
relrod, line 430:
```suggestion
                src, dest, options, *left_overs = chain(mount_str.split(":"), repeat("", 3))
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -354,86 +404,101 @@ def execution_environment_volume_mounts(
...
-            parsed_volume_mounts = []
-            volume_mounts = to_list(entry.value.current)
-            for mount_obj in volume_mounts:
-                if not isinstance(mount_obj, dict):
+                try:
+                    volume_mounts.append(
+                        VolumeMount(
+                            fs_source=src,
+                            fs_destination=dest,
+                            options_string=labels,
```
relrod, line 444:
```suggestion
                            options_string=options,
```

But I still think this field is weird -- we should probably have a function or static method that converts a string to a `Set[VolumeMountOption]` and just pass that, it makes more sense semantically.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -354,86 +404,101 @@ def execution_environment_volume_mounts(
...
+                exit_messages.append(ExitMessage(message=exit_msg))
+                exit_msg = hint
+                exit_messages.append(ExitMessage(message=exit_msg, prefix=ExitPrefix.HINT))
+            else:
+                for volume_mount in entry.value.current:
+                    try:
+                        volume_mounts.append(
+                            VolumeMount(
+                                fs_source=volume_mount.get("src"),
+                                fs_destination=volume_mount.get("dest"),
```
relrod, line 473: `.get()` will return `None` if the key doesn't exist, but the type of `VolumeMount`'s `fs_source` and `fs_destination` say these can't be optional. This shouldn't typecheck..?

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
+
+    def __post_init__(self):
+        """Post process the ``VolumeMount`` and perform sanity checks.
+
+        :raises VolumeMountError: When a viable VolumeMount cannot be created
+        """
+        # pylint: disable=too-many-branches
+        errors = []
+        # Validate the source
+        if isinstance(self.fs_source, str):
```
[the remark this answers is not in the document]
cidrblock: It's coming from a yaml file so we have no idea what it really is.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
+                errors.append("Source not provided.")
+            elif not Path(self.fs_source).exists():
+                errors.append(f"Source: '{self.fs_source}' does not exist.")
+        else:
+            errors.append(f"Source: '{self.fs_source}' is not a string.")
 
-    #: Options for the bind mount.
-    options: List[VolumeMountOption] = field(default_factory=list)
+        # Validate the destination
+        if isinstance(self.fs_destination, str):
```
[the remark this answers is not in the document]
cidrblock: It coming from a yaml/json file so we have no idea what it is.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
+            if self.fs_destination == "":
+                errors.append("Destination not provided.")
+        else:
+            errors.append(f"Destination: '{self.fs_destination} is not a string.")
+
+        # Validate and populate _options
+        if isinstance(self.options_string, str):
```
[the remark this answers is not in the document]
cidrblock: It's coming from a yaml/json file so we don't know what it is.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
-    dest: str
+    fs_source: str
+    """The source file system path of the volume mount"""
+    fs_destination: str
+    """The destination file system path in the container for the volume mount"""
+    settings_entry: str
+    """The name of the settings entry requiring this volume mount"""
+    source: C
+    """The settings source for this volume mount"""
+    options_string: str = ""
```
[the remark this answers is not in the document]
cidrblock: The only way I can think of is to not use a dataclass but a generic class instead, and pass the string into the __init__.....

I think it would be messy compared to this.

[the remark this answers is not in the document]
relrod: Can't we convert it before we call the constructor at all?

```python
options: Set[VolumeMountOption] = VolumeMount.options_from_str(options_str)
VolumeMount(src, dest, options)
```

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -66,31 +72,78 @@ class VolumeMountOption(Enum):
...
+    """Comma delimited options"""
+    _options: List[VolumeMountOption] = field(default_factory=list)
```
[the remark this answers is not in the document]
cidrblock: yep :)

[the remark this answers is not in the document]
cidrblock: It needs to be ordered, so we can make it a tuple of unique values that correspond to the order requested...

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: I think it should contain only uniq items..... let me see if a set breask anything else, I don't think it should

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: Nah, I think we need to make the switch all in one PR

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: Nah, I think we need to make the switch all in one PR

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: Nah, I think we need to make the switch all in one PR

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: entry.value.current is a List of Any
volume_mount is then an Any
so
volume_mount.get("src") is an Any

which satisfies fs_source being a string

cidrblock: If entry value current was something else, I think it would cascade all the way down...

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: I can do another PR to switch it everywhere all at once, but right now I think anything facing the user should be label because that is how it is documented...

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: I can do another PR to switch it everywhere all at once, but right now I think anything facing the user should be label because that is how it is documented...

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
cidrblock: I can do another PR to switch it everywhere all at once, but right now I think anything facing the user should be label because that is how it is documented...

relrod (approved):

Issue #625, opened 2021-11-19
Custom EEs and -m stdout provide no output.


dallasspohn:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY
If I run the playbook with the custom EE, the playbook succeeds without any output to stdout.
ansible-navigator run --eei localhost/ee-motd-minimal --pp never -i inventory -m stdout motd-test.yml
Exit code of 0, everything completes successfully, but no playbook output.

##### ANSIBLE-NAVIGATOR VERSION

```
[student@workstation builder-validate]$ ansible-navigator --version
ansible-navigator 1.0.0

```

##### CONFIGURATION
cat ansible-navigator.yml 
---
ansible-navigator:
  ansible:
    config: ./ansible.cfg

  execution-environment:
    image: ee-supported-rhel8:2.0
    pull-policy: missing

##### STEPS TO REPRODUCE

ansible-navigator run --eei localhost/ee-motd-minimal --pp never -i inventory -m stdout motd-test.yml

##### EXPECTED RESULTS

Should see stdout

##### ACTUAL RESULTS

nothing

cidrblock:
@dallasspohn Ty for the issue

would you mind running it again with `--ll debug --la false` and provide the log output here?

the log details should give us some indication of what's going on behind the scenes

dallasspohn:
@cidrblock Here is the output.
```

[student@workstation builder-validate]$ ansible-navigator run --eei localhost/ee-motd-minimal --pp never -i inventory -m stdout motd-test.yml
[student@workstation builder-validate]$ ansible-navigator run --eei localhost/ee-motd-minimal --pp never -i inventory -m stdout motd-test.yml --ll debug --la false 
[student@workstation builder-validate]$ 
```

cidrblock:
Sorry, I meant the contents of the log file :)

jmighion:
@IPvSean JFYI

dallasspohn:
[student@workstation builder-validate]$ cat ansible-navigator.log 
211119205701.446 INFO 'ansible_navigator.setup_logger' New ansible-navigator instance, logging initialized
211119205701.446 INFO 'ansible_navigator.setup_logger' New ansible-runner instance, logging initialized
211119205701.446 DEBUG 'ansible_navigator.main' Share directory '/usr/lib/python3.8/site-packages/share/ansible_navigator' not found (development path)
[...]

PR #1072, opened 2022-03-09
Include dependency versions and locations in debug log

cidrblock:
Should help resolve issues related to versions.

e.g. #625


src/ansible_navigator/cli.py
```diff
@@ -44,6 +46,17 @@
 logger = logging.getLogger(PKG_NAME)
 
 
+def log_dependencies():
```
relrod, line 49: return type?

Issue #1069, opened 2022-03-09
playbook-artifact file name with OS timezone-based time


akira6592:
##### ISSUE TYPE

- Feature Idea

##### SUMMARY

I also want the time based on the OS time zone as a key to be included in the file name of playbook-artifact. Not just ts_utc.

akira6592:
@cidrblock Thank you for enhancement!

PR #1086, opened 2022-03-14
Enable the use of `time_zone` for logging

cidrblock:
Related #1069 

- Move logging related tests into a unit/logging subfolder
- Add a Formatter to cli.py for the logging timestamp, eventually I think the logging setup should have its own file in the root but was not in scope for this PR
- Eliminated the use of unittest in the current logging test and refactored to use a dataclass


src/ansible_navigator/cli.py
```diff
@@ -75,6 +77,37 @@ def pull_image(args):
         error_and_exit_early(image_puller.assessment.exit_messages)
 
 
+class Formatter(logging.Formatter):
```
ganeshrn, line 80: Since the class is not specific to navigator cli functionality this can be moved to `utils/formatter.py` along with some of the other formatting function like `now_iso`?

cidrblock: I was thinking about a follow-up PR that moved both the Formatter and setup_logger function into its own file called logging. I didn't want to do it in this PR because the diff for setup_logger would be lost with the mew file.  I do think they should stay together tho

ganeshrn (approved):

Issue #1089, opened 2022-03-16
ansible.cfg inventory is not respected with ansible-navigator


cidrblock:
AAP-558

PR #1111, opened 2022-03-19
Use inventory if configured in `ansible.cfg`

cidrblock:
Fixes: #1089 

- Load and store the ansible.cfg file from the $CWD, or other if not using an EE
- Use the ansible.cfg file as inventory source if present
- Switched the `-i` settings entry to not require a value
- Added integration tests for both `run` and `inventory` using an ansible.cfg
- Unit tests to ensure corrupted ansible.cfg files create an exit condition
- Updated the tox.ini to specify the config file, this was necessary because with the use of the venv for the tests, the coverage files were now being written in the root (the coverage was failing without it)

This places the `ansible.cfg` contents in the Internals so it can be used by other post-processors or settings entries in the future if needed.

The ansible.cfg file path can be added to each setting entry in `:settings` in a future PR and possibly a psuedo ansible.cfg entry to show the file contents


src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -482,6 +484,17 @@ def inventory(entry: SettingsEntry, config: ApplicationConfiguration) -> PostPro
                 else:
                     # file path
                     entry.value.current.append(abs_user_path(inv_entry))
+
+        # Something may be required
+        if config.app == "inventory" and entry.value.current is C.NOT_SET:
+            if config.entry("help_inventory").value.current is False:
```
ganeshrn, line 490: Can the two `if` conditions be combined `if config.app == "inventory" and entry.value.current is C.NOT_SET and config.entry("help_inventory").value.current is False`

cidrblock: yes, it looked a little wierd, so I'll set vars for the conditions

ganeshrn (approved):

PR #1113, opened 2022-03-20
Switch to `ANSIBLE_INVENTORY` for inventory environment variable

cidrblock:
No reason to not use the ansible default.


src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -365,7 +365,7 @@ class Internals:
         SettingsEntry(
             name="inventory",
             cli_parameters=CliParameters(action="append", nargs="+", short="-i"),
-            environment_variable_override="ansible_navigator_inventories",
+            environment_variable_override="ansible_inventory",
```
ganeshrn, line 368: Can we have a list of valid environment variables that can be set?

cidrblock: No, we only support one, so no way to have both as a transition....

We could add it, but it would be a bit of work because of how the unit tests were written to test precedence, we'd have to include each of the envars

ganeshrn (approved):

PR #1129, opened 2022-03-22
Update changelog for #1103 to reflect inventory.entries

cidrblock:
Related #1103 
Related #1128


cidrblock:
Merging this as the other went in already

PR #1148, opened 2022-03-25
Add `to_effective` and `to_sources` with tests

cidrblock:
Transform the current settings into a settings file and settings source list

These will be exposed as new CLI parameters and appended to the artifact file in future PRs.

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/transform.py
+++ b/src/ansible_navigator/configuration_subsystem/transform.py
@@ -10,7 +10,7 @@ from typing import Tuple
 from typing import Union
 
 from ..utils.compatibility import importlib_resources
-from ..utils.functions import dict_merge_in_place_list_replace as dict_merge
+from ..utils.dict_merge import in_place_list_replace
 from ..utils.functions import shlex_join
 from .definitions import ApplicationConfiguration
 from .definitions import Constants
@@ -51,7 +51,7 @@ def to_effective(
                 settings_path=path,
                 placeholder=current,
             )
-            dict_merge(rebuilt, partial)
+            in_place_list_replace(rebuilt, partial)
     return SettingsFileType(rebuilt)
 
 
--- a/src/ansible_navigator/configuration_subsystem/transform.py
+++ b/src/ansible_navigator/configuration_subsystem/transform.py
@@ -7,13 +7,67 @@ from typing import Any
 from typing import Dict
 from typing import List
 from typing import Tuple
+from typing import Union
 
 from ..utils.compatibility import importlib_resources
+from ..utils.functions import dict_merge_in_place_list_replace as dict_merge
+from ..utils.functions import shlex_join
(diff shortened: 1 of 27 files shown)


src/ansible_navigator/configuration_subsystem/transform.py
```diff
@@ -7,13 +7,67 @@
...
+            current: Union[bool, int, str, Dict, List] = entry.value.current
+            # It is necessary to un post-process here
+            if path == "ansible-navigator.ansible.cmdline":
+                # post-processed into a list
+                current = shlex_join(entry.value.current)
+            elif path == "ansible-navigator.execution-environment.volume-mounts":
+                current = []
+                for mount in entry.value.current:
+                    parts = mount.split(":")
+                    v_mount = {"src": parts[0], "dest": parts[1]}
```
relrod, line 45: What if these indices don't exist? (Then the post processor would have failed at startup, but it feels awkward to rely on that here)

src/ansible_navigator/configuration_subsystem/transform.py
```diff
@@ -7,13 +7,67 @@
 from typing import Dict
 from typing import List
 from typing import Tuple
+from typing import Union
 
 from ..utils.compatibility import importlib_resources
+from ..utils.functions import dict_merge_in_place_list_replace as dict_merge
```
relrod, line 13: Rather than giving it a crazy name and aliasing, I'd just rename the function.

src/ansible_navigator/utils/functions.py
```diff
@@ -159,6 +160,57 @@ def clear_screen() -> None:
...
+    If left is missing a key from right, the key from right will be added
+
+    :param left: Dict to be merged into
+    :param right: Dict to be merged from
+    :return: Merged dict
+    """
+    key = None
+    try:
+        if left is None or isinstance(left, (str, int, float, bool)):
+            # Border case for first run or if a is a primitive
```
relrod, line 192: should this be

```suggestion
            # Border case for first run or if left is a primitive
```
?

relrod (approved):

src/ansible_navigator/configuration_subsystem/transform.py
cidrblock: After post-processing, they are all a string "src:dest:option":

https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py#L387

src/ansible_navigator/configuration_subsystem/transform.py
cidrblock: There are so many ways to merge dicts, net-new, append lists, this is why the name is what it is.....

Let me think...........

cidrblock: Dedicated files! much cleaner

Issue #603, opened 2021-10-23
ansible-navigator artifacts don't contain EE image, args or runtime information


tonykay:
For debugging and other purposes the log artifact created per run contains no information about the EE, name, tags, runtime, etc. So logs passed, for example, back to an SME, 3rd level support etc., lack encapsulated context. 

##### ISSUE TYPE

 - Feature Idea

##### SUMMARY

Provide an extensive `execution-environment:` dictionary within the log so the exact context is captured. Useful attributes include:

- host OS
- container runtime (including version)
- EE name and tag
- CLI arguments.
- volume mounts
- extra vars

Goal: Anyone with the log should be able to exactly recreate the environment and test without additional files or context being supplied.

cidrblock:
This should follow #577, as it should be able to reuse the generation of structured data from the configuration subsystem

cidrblock:
@tonykay Because the collection of system and EE information would slow down the artifact creation, this was broken into 2 PRs.

The first adds the settings information to the artifact, that is all internal to ansible-navigator so it should have little if any impact on speed.
The second adds a `--diagnostics` flag to ansible-navigator that build a json file full of information about the local system and the configured execution environment, this takes about 5s to run.

Both are linked above and should be in 2.x when released.

PR #1152, opened 2022-03-25
Add settings information to artifact file

cidrblock:
Partial #603 

This adds settings information to the artifact, chipping away at 603.


src/ansible_navigator/actions/run.py
```diff
@@ -418,7 +420,7 @@ def _init_replay(self) -> bool:
             return False
 
         version = data.get("version", "")
-        if version.startswith("1."):
+        if version.startswith("1.") or version.startswith("2."):
```
ganeshrn, line 423: Looks like this condition might need to be updated on every major release. Can't we use `>=` operator.

cidrblock: This was an odd one, because the addition of settings wasn't a breaking change for the TUI, but I updated the version of the artifact to match the major release coming out.

What we really need is a json schema for the artifact file that checks for at least the keys we care about, but there isn't really time for that right now.

I'll add an issue about creating a schema file for it, then we can just check it's sanity in the future.  There's no guarantee it hasn't been modified since navigator wrote it so it's probably a good check to add.

cidrblock: https://github.com/ansible/ansible-navigator/issues/1159

ganeshrn (approved):

Issue #405, opened 2021-05-31
[ui] Notification and warning message boxes should allow for text with color per line


cidrblock:

cidrblock:
![image](https://user-images.githubusercontent.com/18386516/139677787-d48cd0a1-6db8-4c2d-a71b-bed5110d28eb.png)

cidrblock:
the hint line could be yellow and the "Could" line red

cidrblock:
![image](https://user-images.githubusercontent.com/18386516/139677965-c99c2dfc-f2a1-45db-a5bd-4bad8cf554c7.png)

cidrblock:
The goal, to make it look more like the stdout errors

PR #1173, opened 2022-04-01
Clean-up settings messages via stderr and TUI

cidrblock:
Related #405 

- Cleanup the output of messages for settings issues
- Use the same format in the TUI
- Corrected the terminal colors to reflect the theme defaults
- Moved the Exit.... classes out of functions to a definitions file
- Dropped the boolean message because it was a repeat
- Send messages to stderr

This is how things look now, curses on the left, stderr on the right
image


src/ansible_navigator/ui_framework/field_curses_information.py
```diff
@@ -0,0 +1,54 @@
...
+from .curses_window import Window
+from .form_handler_information import FormHandlerInformation
+from .sentinels import Unknown
+from .sentinels import unknown
+from .validators import FieldValidators
+
+
+@dataclass
+class FieldCursesInformation:
+    """An information field mad of curses lines."""
```
ganeshrn, line 17: Should it be `made` instead of `mad` :-)

ganeshrn (approved):

src/ansible_navigator/ui_framework/field_curses_information.py
cidrblock: :)

PR #1181, opened 2022-04-05
Add generated configuration schema file

ssbarnea:
Fixes: #1143


cidrblock:
Ok, saw a couple things... I think for now we might need to do it manually since it might take a little editing since the defaults show the users real directory name...

cidrblock:
BTW, it does look like the version snip is working though

cidrblock (approved):

PR #1182, opened 2022-04-05
Settings file migration

cidrblock:
Migrate a settings file from v1 to v2 format


src/ansible_navigator/utils/version_migration/v1_v2_settings_file.py
```diff
@@ -0,0 +1,384 @@
...
+        """Perform the v1 to v2 migration.
+
+        :param args: Positional arguments
+        :param kwargs: Keyword arguments
+        """
+        if not self.content:
+            with self.settings_file_path.open("r", encoding="utf-8") as f:
+                try:
+                    self.content = yaml.load(f, Loader=Loader)
+                except (yaml.scanner.ScannerError, yaml.parser.ParserError):
```
ganeshrn, line 49: This should handle JSON file format as well right?

ganeshrn: Also, can this method be moved to the base class as it looks to handle common logic across migrations?

cidrblock: yeah, it'll load the JSON just fine, I did add a test specifically using a JSON file because I wanted to see it work.

We centralized all the writing of files into the serialization code, I'd like to do the same with the loading of yaml too

We do it in 5 other places and this will be 6.  

If it's ok, can we leave this here and I'll log an issue to centralize yaml load into a utils/deserialize file and we can shift them all over

ganeshrn (approved):

ssbarnea (approved):

PR #1184, opened 2022-04-06
Add schema defaults

cidrblock:
Add default values to be used when generating a schema, rather than having locally specific default values.


ssbarnea (approved):

ssbarnea:
Also tested manually to see if home directory is no longer leaked. Thanks!

PR #1201, opened 2022-04-12
Remove flake8-quotes

ssbarnea:
This fixes a problem where presence of this plugin caused errors on other projects, as we did
not have custom configuration for it.

Related: https://github.com/zheller/flake8-quotes/issues/105

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -37,6 +37,7 @@ extend-exclude =
 # IMPORTANT: avoid using ignore option, always use extend-ignore instead
 # Completely and unconditionally ignore the following errors:
 extend-ignore =
+  Q  # Safeguard neutering of flake8-quotes : https://github.com/zheller/flake8-quotes/issues/105
   E203,  # annoy black by allowing white space before : https://github.com/psf/black/issues/315
   F401,  # duplicate of pylint W0611 (unused-import)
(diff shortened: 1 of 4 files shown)


cidrblock:
I think this can go too:

```

# flake8-quotes
# https://github.com/zheller/flake8-quotes
# keep at bottom of configuration because some IDEs
# may see the single double quote in the inline-quotes
# entry as an unterminated string and not properly
# highlight the remainder of the file
inline-quotes = "
```

at the bottom of the flake8 file

.flake8
```diff
@@ -37,6 +37,7 @@ extend-exclude =
 # IMPORTANT: avoid using ignore option, always use extend-ignore instead
 # Completely and unconditionally ignore the following errors:
 extend-ignore =
+  Q  # flake8-quotes makes not sense for black users : https://github.com/zheller/flake8-quotes/issues/105
```
cidrblock, line 40: Is this needed if flake8-quotes isn't installed?

cidrblock (changes requested):

.flake8
ssbarnea: If user install flake8-quotes on system python, vscode will complain about such rules. Adding the "Q" here avoids this issue.

cidrblock (approved):

Issue #1224, opened 2022-05-10
Failure to write sample settings to a settings file


cidrblock:
```
ansible_navigator settings --sample > ansible-navigator.yml
```
result in "file cannot be empty", because the shell creates it.

PR #1225, opened 2022-05-10
Fix settings issue when directing to a settings file

cidrblock:
Fixes: #1224

In the case of `settings --sample` being directed to a settings file, allow it to be empty, but return immediately w/o processing it.


thedoubl3j (approved):
lgtm shipit

PR #1275, opened 2022-08-23
Enable flake8-future-annotations

shatakshiiii:


ssbarnea (dismissed):
That is super, you will only need to check the docs that is failing and add few classes to `nitpick_ignore` list to make it report green.

ssbarnea (approved):

cidrblock (approved):
+1

PR #1283, opened 2022-08-26
Enable pyupgrade

shatakshiiii:


src/ansible_navigator/utils/functions.py
```diff
@@ -478,8 +473,7 @@ def shlex_join(tokens: Iterable[str]) -> str:
     :param tokens: The iterable of strings to join
     :returns: The iterable joined with spaces
     """
-    if sys.version_info >= (3, 8):
-        return shlex.join(split_command=tokens)
+    return shlex.join(split_command=tokens)
     return " ".join(shlex.quote(token) for token in tokens)
```
ssbarnea, line 477: You must remove this line (second return) as that was for old python version and pyupgrade did not know to remove it. That should sort the issue.

PR #1289, opened 2022-09-14
Updates in docstring

shatakshiiii:


src/ansible_navigator/ui_framework/ui.py
```diff
@@ -103,11 +107,21 @@ def __call__(
         color_menu_item: Callable = lambda *args, **kwargs: (0, 0),
         content_heading: Callable = lambda *args, **kwargs: None,
     ) -> "Interaction":
-        """Refer to and keep in sync with UserInterface.show"""
+        """Refer to and keep in sync with UserInterface.show.
+
+        :param obj: description
```
ssbarnea, line 112: replace "description" with something that makes sense, otherwise it does not make sense to add docstring parameter information for a method that already has the same information as type information.

I guess that you added this to sort one warning from flake8, but w/o description it does not make sense.

@cidrblock WDYT about this? Which path we should take? I am bit inclined towards avoiding docstring for params that do not add any real value (aka boilerplating to make flake8 happy).

shatakshiiii: Tried adding some context information in place of "description".

ssbarnea (approved):

PR #1305, opened 2022-10-05
Enable black formatting

cidrblock:
This aligns the configuration of black with other devtools maintained projects.

Black will now reformat via precommit, instead of simply warning that it would have.


ssbarnea (approved):

ssbarnea:
I cannot believe we forgot that.

PR #1316, opened 2022-10-13
Diagnostics file permissions

shatakshiiii:
Summary :
- To ensure the diagnostics file is not written to a world readable location
- Setting the file permissions to user only before writing

### DIFF
--- a/src/ansible_navigator/data/ansible-navigator.json
+++ b/src/ansible_navigator/data/ansible-navigator.json
@@ -493,7 +493,7 @@
     "required": [
         "ansible-navigator"
     ],
-    "title": "ansible-navigator settings v2.1",
+    "title": "ansible-navigator settings v2.2",
     "type": "object",
-    "version": "2.1"
+    "version": "2.2"
 }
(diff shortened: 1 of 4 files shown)


cidrblock (changes requested):

cidrblock (approved):
awesome :)

btw, if you haven't seen it, all of the file writing for navigator is handled here: https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/utils/serialize.py  We should talk about if we should move this and use it in the serialize funcs as the default way of writing all files, and what the default permissions would be (or just write the file if no special perms are needed)  That is way outside the scope of this PR, so it can wait for sure.

shatakshiiii:
thank you :)

> btw, if you haven't seen it, all of the file writing for navigator is handled here: https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/utils/serialize.py We should talk about if we should move this and use it in the serialize funcs as the default way of writing all files, and what the default permissions would be (or just write the file if no special perms are needed) That is way outside the scope of this PR, so it can wait for sure.

We should definitely discuss about that. Meanwhile I'm going through it: https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/utils/serialize.py and trying to understand how we can approach this.

Issue #1215, opened 2022-04-28
exec action doesn't return the original exit code


goneri:
##### ISSUE TYPE

- Bug Report

##### SUMMARY

The `exec` action return 0 even in case of failure.

```
$ ansible-navigator exec --eei my-ee --mode=stdout --pull-policy=never -- bash -c 'exit 1'
$ echo $?
```

##### ANSIBLE-NAVIGATOR VERSION

```
$ ansible-navigator --version
ansible-navigator 2.0.1.dev3+g2d24704
```

shatakshiiii:
Could you please check if the issue persists with the latest release of ansible-navigator or not. If the issue is being resolved you can go ahead and close it.

Akasurde:
@goneri ^^

goneri:
Hi @shatakshiiii and @Akasurde, I'm afraid it's still broken.

```
$ ansible-navigator --version
ansible-navigator 2.2.0
$ ansible-navigator exec --eei quay.io/ansible/awx-ee:latest --mode=stdout --pull-policy=never -- bash -c 'exit 1' && echo problem still exists
problem still exists
```

shatakshiiii:
Thanks @goneri for investigating it again. We'll look forward to fix that.

PR #1330, opened 2022-11-02
Update exec action to return with correct exit code

shatakshiiii:
Resolves: https://github.com/ansible/ansible-navigator/issues/1215


ssbarnea (approved):
LGTM but wait for @cidrblock to approve this one as I am not sure if there are other undesired side effects that I may miss.

Issue #1221, opened 2022-05-10
Disable and fix errors, cspell allowCompoundWords


cidrblock:
Remove:

```
languageSettings:
  - languageId: python
    allowCompoundWords: true
```
from cspell.config.yaml, and fix errors or PR to python dictionary

Related: https://github.com/streetsidesoftware/cspell/releases/tag/v5.20.0

cidrblock:
and discouraged: https://github.com/streetsidesoftware/cspell/issues/2092

PR #1333, opened 2022-11-08
Disable cspell allowCompoundWords

shatakshiiii:
Fixes: #1221


ssbarnea (commented):
This is changing the API - I am not so sure it was a good idea to change arg names. Also the same about `argspec` which is well known compound-name, i never seen someone using `arg_spec`. I wonder what @cidrblock thinks about this.

Practical example: having `colno` and `col_name` in same method feels worse than what we have now.

Similar `nonblocking` being split...

PR #1358, opened 2022-12-14
Source runtime and test image from dockerfile

cidrblock:
Ongoing work to centralize the default ee.

1) move the docker file to the source so it is packaged
2) update dependabot to look in that location for docker ecosystem
3) pull from that file for the runtime default
4) use runtime logic to provide the default for tests

additional PRs:
1) will remove the tox entry
2) put the pull in pytest for the default ee
3) move other image names into dockerfiles
4) get rid of the constant being set to a function :)
5) Move from quay to ghcr when centralized

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ version: 2
 updates:
   - package-ecosystem: docker
     # We use this file to load the last known good version of the image
-    directory: .config/
+    directory: .src/ansible_navigator/data
     schedule:
       interval: daily
     labels:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,7 @@ version: 2
 updates:
   - package-ecosystem: docker
     # We use this file to load the last known good version of the image
-    directory: .src/ansible_navigator/data
+    directory: src/ansible_navigator/data/
     schedule:
       interval: daily
     labels:
--- a/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_configuration.py
@@ -13,6 +13,7 @@ from ..utils.functions import abs_user_path
 from ..utils.functions import get_share_directory
 from ..utils.functions import oxfordcomma
 from ..utils.key_value_store import KeyValueStore
+from ..utils.packaged_data import retrieve_content
 from .definitions import ApplicationConfiguration
 from .definitions import CliParameters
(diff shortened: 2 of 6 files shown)


.github/dependabot.yml
```diff
@@ -4,7 +4,7 @@ version: 2
 updates:
   - package-ecosystem: docker
     # We use this file to load the last known good version of the image
-    directory: .config/
+    directory: .src/ansible_navigator/data
```
ssbarnea, line 7: Remove the dot here or it will not work. Also add the trailing slash like in original.

src/ansible_navigator/data/default_ee_dockerfile
```diff
@@ -0,0 +1 @@
+FROM quay.io/ansible/creator-ee:v0.9.2
```
ssbarnea, line 1: You MUST keep the name `Dockerfile` or dependabot will not work. There is no way to configure custom filenames.

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
```diff
@@ -88,6 +89,17 @@ def generate_share_directory():
     return share_directory
 
 
+def retrieve_default_ee_image() -> str:
```
ssbarnea, line 92: Add lru caching here, the value never changes.

ssbarnea (changes requested):

.github/dependabot.yml
cidrblock: done

src/ansible_navigator/data/default_ee_dockerfile
cidrblock: https://github.com/dependabot/dependabot-core/compare/v0.67.1...v0.67.2

although I'm not sure how to test it

src/ansible_navigator/configuration_subsystem/navigator_configuration.py
cidrblock: runtime it only gets called once at initialization

it will get called multiple time in the tests, but I intend to replace the CONSTANT with a session fixture eventually

src/ansible_navigator/data/default_ee_dockerfile
ssbarnea: You cannot, if the path does not work, we make another PR to fix it.

cidrblock:
Will address requested changes in another PR if there is an issue

PR #1360, opened 2022-12-14
Retrieve the default EE using pytest

cidrblock:
Ongoing work to centralize the default ee.

1. Use a session-scoped autorun fixture to pull the default EE
2. Remove that work from tox 

additional PRs:

1. move other image names into dockerfiles
2. get rid of the constant being set to a function :)
3. Move from quay to ghcr when centralized

related: #1358 

Note: This PR also limits the number of builder builds we perform, and increases the timeout, as this is a common place for CI failures

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -128,7 +128,7 @@ def test_dir_fixture_dir(request):
 
 @pytest.fixture(scope="session", name="default_ee_image_name")
 def _default_ee_image_name() -> str:
-    """Retreive the default EE image name."""
+    """Retrieve the default EE image name."""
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -25,7 +29,7 @@
 
 
 @pytest.fixture(scope="session", name="valid_container_engine")
-def fixture_valid_container_image():
+def fixture_valid_container_engine():
```
cidrblock, line 32: this was just a little unrelated cleanup while in the neighborhood

ssbarnea (approved):

ssbarnea (approved):

PR #1363, opened 2022-12-15
Simplify test ids

ssbarnea:


cidrblock (approved):
I'm not a fan of having to manually index a long list of tests, but they also don't change that much.

I would prefer a-z or 01-99 so my OCD doesn't kick in when looking at the test tree in VSCode.

Either way, if it helps run from the command line let's do it.

Someday I might get so annoyed I switch them to something that is ordered correctly in vscode

PR #1365, opened 2022-12-15
Always print image prologue

cidrblock:
REgardless of the need to pull, always print the image pull summary.  This will help with image caching debugging.


shatakshiiii (approved):

cidrblock:
codecov hadn't finished

PR #1383, opened 2023-01-10
Allow fixture update env var

cidrblock:
Related #1368

Need to let this environment variable through for fixture updates


cidrblock:
merging as it is blocking the creator-ee version upgrade

PR #1384, opened 2023-01-10
Set the prompt for each tmux session to `bash$`

cidrblock:
This will remove the bash version from the test fixtures.

Other changes typical for fixture generation on a different system, different python version and test name updates.

Related #1376


shatakshiiii (approved):

ssbarnea:
@cidrblock looking at the diff I realise that huge amount of stuff that would cause tests to fail in the future, even time, as we check the age of the containers, publishing a new creator-ee will break tests too.

I am going to merge it because it looks to work in good direction but there is still a lot more to change.

ssbarnea (approved):

PR #1392, opened 2023-01-11
Mask image ages

cidrblock:
Although this won't catch 100% of these fixture changes, it should eliminate some.


cidrblock:
merging w/o approval as this is only minor test/fixture changes

PR #1393, opened 2023-01-11
Mask hostname from exec

cidrblock:
Mask out the hostname from the exec test fixture


cidrblock:
Merging, simple test/fixture change

PR #1377, opened 2023-01-09
Add doc about debugging ansible-navigator

shatakshiiii:
- Adds information about debugging ansible-navigator using Vscode.
- Examples around debugging *subcommands* in Navigator.

PR #1398, opened 2023-01-12
Add more configuration entry in vscode settings

shatakshiiii:
- This PR adds example configuration to debug ansible-navigator subcommands.
- Example of debugging `exec` subcommand is mentioned.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,11 +6,18 @@
       "name": "Debug tests",
       "type": "python",
       "request": "launch",
-      "module": "ansible_navigator",
-      // Uncomment args attribute to debug exec subcommand with some parameter.
-      // "args": ["exec", "--", "pwd"],
       "purpose": ["debug-test"],
       "console": "integratedTerminal"
+    },
+    // Configuration for pure debugging (use args attribute accordingly)
+    {
+      "name": "Python: Module",
+      "type": "python",
+      "request": "launch",
+      "module": "ansible_navigator",
+      "args": ["exec", "--", "pwd"],
+      "cwd": "${workspaceFolder}/src",
+      "justMyCode": false
     }
   ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,6 +6,9 @@
       "name": "Debug tests",
       "type": "python",
       "request": "launch",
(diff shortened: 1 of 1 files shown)


cidrblock:
The one I added is for debugging tests.. Did you mena to add an additional entry for just pure debugging?

```
 {
      "name": "ansible-navigator",
      "type": "python",
      "request": "launch",
      "module": "ansible_navigator",
      "args": ["exec", "--", "pwd"],
      "cwd": "${workspaceFolder}",
      "justMyCode": false
    }
```
somehting like that? ^

shatakshiiii:
> The one I added is for debugging tests.. Did you mean to add an additional entry for just pure debugging?
> 
> ```
>  {
>       "name": "ansible-navigator",
>       "type": "python",
>       "request": "launch",
>       "module": "ansible_navigator",
>       "args": ["exec", "--", "pwd"],
>       "cwd": "${workspaceFolder}",
>       "justMyCode": false
>     }
> ```
> 
> somehting like that? ^

Yes, I tried adding exact piece as an additional configuration, but somehow the "Debug tests" wasn't working just as it is for me. Although not sure if this is the correct way of doing this. If not I can remove this, as we will have these  information as a document with [PR](https://github.com/ansible/ansible-navigator/pull/1377).

ssbarnea:
Please don't add comments in JSON as we are never sure if these will work or not and editing the file using vscode UI will also break them.

For optional stuff, you can always add extra commands in addition to default "debug" one.

cidrblock:
Aren't the comments fine?  I thought vscode fully supported JSONC for these files:

https://code.visualstudio.com/docs/languages/json#_json-with-comments

ssbarnea:
They work here but what happens when you try to add an new entry? Also the comments here are not visible to the user in vscode which makes them kinda useless.

I was proposing adding the current commented example as a permanent run/debug configuration instead of a comment, at least users will be able to run the "pwd" and the "customization" is a no brainer when you see the extra launcher.

Basically we would "document" it w/o writing comments, and making that visible in vscode UI.

cidrblock:
ah! yes let's add individual entries, I think that is a great idea

cidrblock (approved):

PR #1409, opened 2023-01-19
Use a valid container engine in exit code tests

cidrblock:
Without being specified the default of `podman` was used.  This will allow the test to use the "valid" container engine for the system where the tests are being run.

This was identified while running tests on macos with docker.


cidrblock:
Since the scope here is really small and will allow macos testing to continue, I'll merge it.

PR #1413, opened 2023-01-20
Run stdout exit code test with a TTY

cidrblock:
See #1141


cidrblock:
Since this a dupe, merging, as it will unblock the macOS testing

Issue #1115, opened 2022-03-21
ansible-navigator run can't use --ask-pass option


DNTYO:
##### ISSUE TYPE

- Bug Report

##### SUMMARY

```
~ $ ansible-navigator run ping.yml -i hosts --ask-pass --mode stdout
^CTerminated
```

##### ANSIBLE-NAVIGATOR VERSION

```
~ $ ansible-navigator --version
ansible-navigator 1.1.0
```

##### CONFIGURATION

```
~ $ pwd
/home/podman
~ $ ls ansible-navigator.yaml
ls: ansible-navigator.yaml: No such file or directory
```

##### LOG FILE

```
~ $ ansible-navigator run --help-playbook --mode stdout
  -k, --ask-pass        ask for connection password
```

```
~ $ ansible-navigator run ping.yml -i hosts --ask-pass -vvv --mode stdout
ansible-playbook [core 2.11.3.post0]
  config file = None
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/local/bin/ansible-playbook
  python version = 3.8.6 (default, Jan 29 2021, 17:38:16) [GCC 8.4.1 20200928 (Red Hat 8.4.1-1)]
  jinja version = 2.10.3
  libyaml = True
No config file found; using defaults
!! STOP !!
```

```
~ $ cat ansible-navigator.log
220321060709.190 WARNING 'ansible_navigator.run' Dirty exit, killing the pid
```

##### STEPS TO REPRODUCE

I use --ask-pass option for "ansible-navigator run" command

##### EXPECTED RESULTS

I look "SSH password: " prompt.
I can input ssh password.

##### ACTUAL RESULTS

I didn't look "SSH password: " prompt.
I can't input ssh password.

##### ADDITIONAL INFORMATION

Dockerfile about create container image for use ansible-nabigator
```
FROM alpine:edge

RUN echo -e "https://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
    apk --no-cache add podman tzdata fuse-overlayfs && \
    apk --no-cache add gcc musl-dev libffi-dev oniguruma-dev && \
[...]

PR #1419, opened 2023-01-23
Add `enable_prompts` parameter for run subcommand

shatakshiiii:
This adds a new CLI parameter `--enable-prompts`  (abbreviated as `--ep`) for run subcommand, that disables playbook artifact creation and executes playbook in mode stdout bydefault.

Closes: https://github.com/ansible/ansible-navigator/issues/1115


tests/integration/actions/run/test_stdout_tmux.py
```diff
@@ -79,6 +79,15 @@ class ShellCommand(UiTestStep):
         ).join(),
         present=["usage: ansible-playbook [-h]"],
     ),
+    ShellCommand(
```
cidrblock, line 82: The test should include a playbook that prompts, this might be a little tricky since I don't think we have an example of one. I'm happy to help with this

cidrblock (approved):
Looking awesome! TY

tests/integration/_tmux_session.py
```diff
@@ -294,7 +294,7 @@ def interaction(
             if showing:
                 if isinstance(search_within_response, str):
                     for line in showing:
-                        if search_within_response in line:
+                        if (showing[0] == "SSH password:") or (search_within_response in line):
```
cidrblock, line 297: let's do this differently.... we shouldn't need to touch the tmux session file....

tests/integration/actions/run/test_stdout_tmux.py
```diff
@@ -116,6 +99,23 @@ class ShellCommand(UiTestStep):
         ).join(),
         present=["from.ansible.cfg", "ok=1"],
     ),
+    ShellCommand(
+        comment="run playbook with enable-prompts and show prompting",
+        user_input=StdoutCommand(
+            cmdline=f"{playbook_path} -k --enable-prompts",
+            execution_environment=True,
+        ).join(),
+        present=["password:"],
```
cidrblock, line 108: we should be able to set

`search_within_response = "SSH password"`

here, since it is a property of the ShellCommand class

Issue #520, opened 2021-07-29
AR rotate artifats count does not work


mnecas:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

When running 
`ansible-navigator run playbook.yml --eei ovirt:local -vvv -m stdout --rad /tmp/test`
it creates the artifacts dir with all files and it stays there.
But when I run
`ansible-navigator run playbook.yml --eei ovirt:local -vvv -m stdout --rad /tmp/test --rac 0`
it creates the dir only when running and after the playbook finishes it removes the artifacts dir.
The same goes when I put there any number.

##### ANSIBLE-NAVIGATOR VERSION

```
ansible-navigator 1.0.0
```

##### CONFIGURATION

Ansible runner version (master release)
```
❯ ansible-runner --version
2.0.0.0a4.dev69
```

cidrblock:
Related: https://github.com/ansible/ansible-navigator/issues/497

PR #1438, opened 2023-02-12
Runner artifact cleanup

eamigo:
The artifact_dir was only deleted if --rac was set by the user. 

This will now delete the entire temporary private_data_dir instead of just the artifacts_dir from the last run if the private_data_dir is a generated temp directory. 

Closes: #497 
Closes: #520


cidrblock:
Approving this, although I'm going to leave #497 open because we really need to not be using __del__ as the cleanup mechanism bb/c it's not guaranteed to always run

Issue #1385, opened 2023-01-10
ansible-navigator run can't directly access a playbook provided by a collection


zjpeterson:
##### ISSUE TYPE

- Bug Report

##### SUMMARY

`ansible-playbook` is able to [directly access a playbook](https://docs.ansible.com/ansible/latest/collections_guide/collections_using_playbooks.html#using-a-playbook-from-a-collection) provided by a collection. Trying to do the same thing with `ansible-navigator run` results in the tool trying to resolve it as a filename.

##### ANSIBLE-NAVIGATOR VERSION

```
ansible-navigator 2.2.0
```

##### STEPS TO REPRODUCE

- Create a collection with a playbook located in the `playbooks/` directory
- Try to run `ansible-navigator run namespace.collection.playbook_name --eei container-with-collection-installed`

##### EXPECTED RESULTS

Playbook runs

##### ACTUAL RESULTS

Fully-qualified playbook name gets parsed as though it were a file and an error like the following is returned:
```
ERROR! the playbook: /working/dir/namespace.collection.playbook_name could not be found
```

##### ADDITIONAL INFORMATION

Running `ansible-playbook namespace.collection.playbook_name` does succeed with the collection installed locally, so this is something Ansible is supposed to be able to do.

Can be worked around by creating and running a one-line playbook like the following
```
- import_playbook: namespace.collection.playbook_name
```

cidrblock:
Thanks for this! We'll look into it and post back with findings

PR #1440, opened 2023-02-13
Add support to direct access a playbook provided by a collection for :run subcommand

shatakshiiii:
**Summary:**

   This PR adds support for FQCN playbook in navigator i.e. `namespace.collection.playbook_name` format.
   Fixes: #1385


src/ansible_navigator/actions/run.py
```diff
@@ -166,6 +168,32 @@ def filter_content_keys(obj: dict[Any, Any]) -> dict[Any, Any]:
...
+        playbook_path = playbook
+        start_path = playbook.rsplit("/", 1)[0]
+        playbook_name = playbook.split(".")[-1]
+        end_paths = [f"{playbook_name}.yml", f"{playbook_name}.yaml"]
+        for end_path in end_paths:
+            file = glob.glob(os.path.join(start_path, "**", end_path), recursive=True)
+            if len(file) != 0:
+                playbook_path = file[0]
+                break
+    return playbook_path
```
sivel, line 194: I'm not sure this code will work the way users will ultimately want it to.  When calling a playbook from a collection, it is just represented by an FQCN and should not be referenced by path.

Additionally, an FQCN for a playbook does support a special case of including the file extension, but that is more for compatibility than anything.

An FQCN for a playbook is just like anything else:

```
{namespace}.{name}.{resource}
```

So we shouldn't attempt to dereference a FQCN to a file path I don't think.  If we did that, then the playbook would not be treated as part of a collection. ansible-core only knows a playbook is part of a collection if the playbook resource that is passed to `ansible-playbook` is an FQCN.  Otherwise, it's just a path to an arbitrary playbook file, without "self" collection context.

sivel: To clarify one point, both of these are equivalent for `ansible-playbook`:

```
ansible-playbook namespace.name.some_playbook
ansible-playbook namespace.name.some_playbook.yml
```

cidrblock (approved):

shatakshiiii:
Thanks @sivel  for those detailed comments, it helped a lot!

PR #1453, opened 2023-03-20
Drop python 3.8 support

shatakshiiii:
With this, Ansible-Navigator will be dropping support for **Python 3.8**


cidrblock:
There's 2 in here that can be removed: https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/utils/compatibility.py#L18  the cooresponding import statements cna be updated to directly import from the libs

ssbarnea (changes requested):
Search for py38 and 3.8 in source code and you will spot few other places that needs to be changed.

I use ripgrep aka `rg` to do that but is up to you.
```
tests/fixtures/integration/actions/templar/test_direct_interactive_ee_readme.py/test/0.json:19:109:[Omitted long line with 1 matches]
docs/faq.md:229:58:$ ansible-navigator exec -- ansible-test sanity --python 3.8
.github/workflows/release.yml:21:38:      - name: Switch to using Python 3.8 by default
.github/workflows/release.yml:24:27:          python-version: 3.8
requirements.txt:2:57:# This file is autogenerated by pip-compile with python 3.8
src/ansible_navigator/tm_tokenize/_types.py:5:42:    from typing import Protocol  # python3.8+
src/ansible_navigator/utils/functions.py:471:39:    ``shlex.join`` was new in version 3.8
.pre-commit-config.yaml:142:19:        args: ["--py38-plus"]
```

I guess you can skip updating the fixtures if they are passing, even if they include references to py38.

PR #1456, opened 2023-03-20
Down version for select deps

cidrblock:
`libtmux` - something changed related to screen width
`ansible-core` - ansible inventory output order appears to have changed

These should be upgraded after tests are green. 

Related #1450

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@
 #    pip-compile --extra=docs --extra=test --no-annotate --output-file=requirements.txt --strip-extras setup.cfg
 #
 ansible-builder==1.2.0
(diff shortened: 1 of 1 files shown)


ssbarnea:
@cidrblock While these might make the test green they will prevent dependabot and us from upgrading dependencies. Not sure what we can do to prevent similar problems. 

The reality is that AFAIK none of these changes affects the runtime but both breaks the tests, and we should really be able to run the tests with all supported versions.

ssbarnea (approved):

cidrblock:
These deps should and will be upgraded individually after this PR. Just taking it apart.........

PR #1459, opened 2023-03-21
ansible-core 2.14.3

cidrblock:
Upversion ansible-core in the requirements to 2.14.3

- Needed to sort the hosts and groups in the inventory TUI since ansible changed it's sort order for the ansible-inventory command
- Increased the pane height for the settings stdout tests so the entire schema could be seen in the resulting fixtures
- Ensure `received_output_list` is always populated to avoid `UnboundLocalError: cannot access local variable 'received_output_list' where it is not associated with a value`. There are only small mentions of this in later 3.11 version so I'm not entirely sure why we are hitting it just now.

all other changes are simply expected fixture updates


shatakshiiii (approved):

ssbarnea (approved):

cidrblock:
USing admin permissions, this seem unrelated to the changes within

docs/readthedocs.org:ansible-navigator — Read the Docs build failed!

PR #1463, opened 2023-03-24
Docs and type hints in test integration root (D/DAR)

cidrblock:
Ongoing work to improve type hins and doc string.

Changes are made to files in the root of the test intregration directory.

This also restores format on save accidentally changed in #1450


cidrblock:
self approving since it's all test related, docstring and type hints

PR #1472, opened 2023-03-31
Update schema

cidrblock:
Update schema to reflect version 3


cidrblock:
Need to release a patch, pushing this in

Issue #1437, opened 2023-02-09
Can't run playbook from configuration with extra-vars in parameters


eleanorcuallado:
##### ISSUE TYPE

- Bug Report

##### SUMMARY
When running `ansible-navigator run` with extravar parameters but no playbook name, the command will take the extravar parameter as a playbook, even if it is defined in the configuration.

##### ANSIBLE-NAVIGATOR VERSION

```
ansible-navigator 2.2.0
```
Using with python 3.8.13

##### CONFIGURATION
Relevent configuration:
```
ansible-navigator:
  playbook:
    path: "playbook.yml"
```
##### LOG FILE
Nothing relevent found in the logs.

##### STEPS TO REPRODUCE
Here are two examples that reproduce this issue:
- `ansible-navigator run -i <inv> -e "my_var=value"`
- `ansible-navigator run -i <inv> -- -e "my_var=value"`

##### EXPECTED RESULTS
ansible-navigator should realize no playbook is given in the command line, and use the one given in the configuration.

##### ACTUAL RESULTS
ansible-navigator uses the first string it finds as a playbook:
- in the first example, it uses "my_var=value"
- in the second example, it uses "-e"

##### ADDITIONAL INFORMATION
Interface when running the first example:
![image](https://user-images.githubusercontent.com/34506134/217780917-4dda2673-71b1-4a4d-8126-e36916741439.png)
Interface when running the second example:
![image](https://user-images.githubusercontent.com/34506134/217781028-6301d2d7-2124-431a-b587-093c31aecf1c.png)

ansible-navigator is able to use the playbook when using `ansible-navigator run -i <inv>`, meaning it is able to find and use the playbook; it seems the problem comes from the automatic detection of playbook parameter, that is too aggressive when looking for a string parameter (most interestingly though, it recognizes the `-e` string itself as a parameter, but not its value that comes after it). Since the `-i` parameter is not impacted by the issue (it can be used without it being recognized as a playbook name), it would seem the issue does not appear for every parameter.

cidrblock:
ansible-navigator run -i <inv> -e "my_var=value"

This one I can understand b/c argparse isn't aware of the config file and its entries, and navigator is also unaware of the `-e` since we don't track every ansible-playbook parameter. I don't think this can be fixed.

ansible-navigator run -i <inv> -- -e "my_var=value"

[...]

PR #1474, opened 2023-04-03
Allow extra parameters when playbook is defined in configuration/settings file

shatakshiiii:
**Summary:** 

With this, all the parameters that are not directly supported by ansible-navigator and usually used via a  `--` delimiter, can also be used when playbook is defined in the navigator's settings file.

This will allow `ansible-navigator run` to support extra parameters when playbook is specified in the configuration (instead of command itself)

Example: `ansible-navigator run -- -e my_var=value`

When configuration can be something like :
```
---
ansible-navigator:
  ansible:
    playbook:
      path: "playbooks/ping.yml"
 ```
      
Fixes: https://github.com/ansible/ansible-navigator/issues/1437


cidrblock (approved):
looks great

PR #1489, opened 2023-04-19
Enable testing with devel branches

ssbarnea:


ssbarnea:
@shatakshiiii This PR proves the fact that navigator cannot be tested with unreleased versions of ansible-lint, even if these versions are not really breaking anything.

This uncovered the fact that navigator testing implementation is too fragile and needs to be refactored to allow this kind of testing.

cidrblock:
@shatakshiiii is correct, we can do a follow up PR for lint from the main branch later.

PR #1494, opened 2023-04-27
Adjust Docs For Ansible Vault

timway:
* The instructions created a shell variable that would not have been passed into execution environment causing the described process to fail
* A friend let me know the instructions were not working and I confirmed it myself locally (@penguinpages)


ssbarnea (approved):

cidrblock:
TY @timway ! Good find

PR #1503, opened 2023-04-28
Compatibility with ansible-builder v3

cidrblock:
No code changes necessary, only test updates.

Pardon the removal of isort from the vscode settings, it was driving me nuts.


cidrblock:
test changes only, self-approving

Issue #1448, opened 2023-03-15
ERROR! Invalid callback for stdout specified: awx_display


davidgleiss:
##### ISSUE TYPE

- Bug Report

##### SUMMARY

I cannot start any playbook with ansible-navigator.

OS:
```
Operating System: Ubuntu 22.04.2 LTS              
Kernel: Linux 5.15.0-67-generic
```

##### ANSIBLE-NAVIGATOR VERSION

```
ansible-navigator 2.2.0
```

##### CONFIGURATION

```yaml
---
ansible-navigator:
  ansible:
    config:
      path: .ansible.cfg
    inventory:
      help: false
      entries:
        - inventory.yaml
  execution-environment:
    pull:
      policy: missing
    container-engine: docker
    enabled: true
    image: quay.io/ansible/awx-ee:latest

  playbook-artifact:
    enable: false

  logging:
    append: true
    file: /tmp/ansible-navigator.log
    level: warning
```

##### LOG FILE

(https://gist.github.com/davidgleiss/de768448496cd63f6ef218dcd8c54406)

##### STEPS TO REPRODUCE

```bash
ansible-navigator run test.yaml -m stdout
```

or

```bash
ansible-navigator run test.yaml
```

##### EXPECTED RESULTS

Playbook should be executed

##### ACTUAL RESULTS

```
[WARNING]: error in 'jsonfile' cache plugin while trying to create cache dir /runner/artifacts/1a0ae560-374c-4084-a8bf-4e3a65e37f31/fact_cache : b"[Errno 13] Permission denied:
'/runner/artifacts/1a0ae560-374c-4084-a8bf-4e3a65e37f31'"
ERROR! Invalid callback for stdout specified: awx_display
```

##### ADDITIONAL INFORMATION

```
Image: awx-ee:latest (primary) (Information about ansible and ansible collections)                                                                                     
 0│---
 1│ansible:
 2│  collections:
 3│    details:
 4│      '@NAMESPACE@.@NAME@': 3.0.1
 5│      amazon.aws: 5.3.0
 6│      ansible.posix: 1.5.1
 7│      ansible.windows: 1.13.0
 8│      awx.awx: 21.13.0
 9│      azure.azcollection: 1.14.0
10│      google.cloud: 1.1.3
11│      kubernetes.core: 2.4.0
12│      openstack.cloud: 2.0.0
13│      redhatinsights.insights: 1.0.7
14│      theforeman.foreman: 3.9.0
15│  version:
[...]

PR #1539, opened 2023-05-20
Set user to root with podman

cidrblock:
Related: https://github.com/ansible/ansible-builder/issues/541
Related: https://github.com/ansible/ansible-navigator/issues/1448

Due to changes with the container build introduced with builder version 3, we will now run as user=root in a podman container.

See the builder issue above for details

### DIFF
--- a/src/ansible_navigator/runner/base.py
+++ b/src/ansible_navigator/runner/base.py
@@ -86,6 +86,14 @@ class Base:
         self.finished: bool = False
         self.status: str | None = None
         self._runner_args: dict = {}
+
+        # when the ce is podman, set the container user to root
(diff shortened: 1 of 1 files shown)


ssbarnea (approved):

ssbarnea:
@cidrblock I approved it but I need to mention that I do not see a test for it, and that is dangerous as it prevents us from introducing a regression in the future.

cidrblock:
@ssbarnea , soory I should have linked 2 issues related to that:

https://github.com/ansible/ansible-navigator/issues/1540
https://github.com/ansible/creator-ee/issues/330

PR #1542, opened 2023-05-22
Update creator-ee version to 0.18.0

shatakshiiii:
Closes: #1532


ajinkyau (approved):

ssbarnea (approved):
Looks ok, the big question is if we can remove some of these fixtures, so we would not need to update 80 files on every creator-ee bump.

priyamsahoo (approved):

shatakshiiii:
> Looks ok, the big question is if we can remove some of these fixtures, so we would not need to update 80 files on every creator-ee bump.

These fixtures largely affect the integration tests we have, for each action (images in this case). That's why we end up updating all those files.

PR #1549, opened 2023-06-12
Fix the broken VSCode test tree link in Contributing Guidelines

shatakshiiii:
Summary:

- This will fix the broken VSCode test tree link in [CONTRIBUTING.md](https://github.com/ansible/ansible-navigator/blob/ac408aa77a5d8d00e51c17daff4a255324df6da4/.github/CONTRIBUTING.md)
- To make the [Contributions](https://ansible.readthedocs.io/projects/navigator/) heading more precise in navigator's RTD.


ssbarnea (changes requested):
What is the reasoning behind this PR?

We have the document inside `.github` because that is the only location other than repository root that is recognized by github for marking a repository as having a contributing documentation.

When we build our mkdocs, we can use a symlink to that file, making everyone happy.

shatakshiiii:
Based on what I read over [here](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) looks like it is ok to have a contribution guidelines file in either project repository's root, docs, or .github folder.

The idea is, if it is possible to have `CONTRIBUTING.md` file inside docs folder, then how about just keeping one inside docs/ and remove it from .github/ because eventually we would need to render it as docs ?

Linking one example of this over here: https://github.com/ansible/ansible-lint/blob/main/docs/contributing.md

Still I need suggestions on this, as to how we want these files or folder to look moving ahead!

cidrblock (approved):
Will github find it here?

docs/contributing/CONTRIBUTING.md

approving, if GH can't find it we'll need to move it again

shatakshiiii:
> Will github find it here?
> 
> docs/contributing/CONTRIBUTING.md
> 
Updates on this,

Did some search and looks like github can't find it in `docs/contributing/CONTRIBUTING.md`. 

Hence, Fixed the broken link in this PR by introducing similar images folder in .github/ so that the VScode test tree snippet is visible for anyone browsing https://github.com/ansible/ansible-navigator/contribute.

And a minor change in homepage of RTD, so that people are aware of the fact that `Contributions` is not a clickable link,
rather just a heading :)

Issue #1559, opened 2023-06-24
FAQ: please consider HISTCONTROL when storing the vault password in an environment variable


pcfe:
##### ISSUE TYPE

- Doc issue

##### SUMMARY

 With 7fc69ec42ad713cc85b1ea4037b164788ed62548 in #1171 comes relying on `HISTCONTROL` being set to either _ignorespace_ or _ignoreboth_.

That is suboptimal on systems where `HISTCONTROL` is set to _ignoredups_

Please include a note to this effect to the FAQ entry.

PR #1560, opened 2023-06-24
fix vault file with env var

pcfe:
attempts to address #1559 

closes: #1559


shatakshiiii (approved):

shatakshiiii:
Thank you for this fix @pcfe!

PR #1574, opened 2023-07-14
Add tests for `lint`, without an EE

shatakshiiii:
This will add the interactive tests for `lint` subcommand, without an EE:

- test_direct_interactive_noee
- test_welcome_interactive_noee


tests/fixtures/integration/actions/lint/test_direct_interactive_noee.py/test/0.json
```diff
@@ -0,0 +1,22 @@
...
+    "comment": "lint interactive without ee",
+    "additional_information": {
+        "present": [
+            "Issues were found while applying the settings"
+        ],
+        "absent": [],
+        "compared_fixture": false
+    },
+    "output": [
+        "Warning: Issues were found while applying the settings.",
```
cidrblock, line 13: These tests should pass and have output similar to what the tests w/o the EE get.... This looks like ansible-lint with an EE doesn't work at all...

cidrblock (changes requested):

tests/fixtures/integration/actions/lint/test_direct_interactive_noee.py/test/0.json
shatakshiiii: yes, right now ansible-lint does not work without execution environment enabled. Okay, so I'll add lint as a requirement first and then update these tests to reflect the same output.

priyamsahoo (approved):

PR #1578, opened 2023-07-20
Check for ansible-lint version if not using an EE

shatakshiiii:


src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -655,6 +656,24 @@ def lintables(
...
+                except subprocess.CalledProcessError:
+                    exit_messages.append(
+                        ExitMessage(
+                            message=("ansible-lint does not seem to be installed correctly.")
+                        ),
+                    )
+                    exit_messages.append(
+                        ExitMessage(
+                            message=(
+                                "Ensure the command `ansible-lint --version` can be run prior to"
```
priyamsahoo, line 671: Can we make this message more generic? Something like, `Ensure that ansible-lint is installed in the environment`?

cidrblock: I prefer the original, this check is follow on to finding the executable, to ensure not only that is it installed, but it is a working installation.

priyamsahoo: Okay. @shatakshiii, I think you have to revert the changes :)

priyamsahoo (approved):

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
shatakshiiii: reverted that change :)

PR #1663, opened 2024-02-05
Update libtmux

ssbarnea:
Related: https://github.com/tmux-python/libtmux/issues/468


shatakshiiii:
One related issue: https://github.com/tmux-python/libtmux/issues/468

PR #1678, opened 2024-02-12
Auto-retry tests that fail randomly

ssbarnea:


ssbarnea:
I will ignore the coverage failure as it is a rounding issue, not really reduction.

PR #1686, opened 2024-02-13
Resolve mypy disallow_any_generics (partial)

ssbarnea:


ssbarnea:
Ignored codecov report as being caused by rounding errors.

PR #1714, opened 2024-03-12
Improve the FAQ on how to use ansible vaults with ansible-navigator

rseabra:
This is a better way to use ansible vaults with ansible-navigator, I really didn't like the existing solutions, but actually one of them together with our usage of gpg inspired me, so here's the PR. :)


rseabra:
I don't know how to add a label, but it's an enhancement.

shatakshiiii (approved):

PR #1756, opened 2024-05-13
Address some of the ruff exceptions (miscellaneous)

shatakshiiii:
This PR removes `PLC1901`, `RUF009`, `RUF100`, `S110` from the ruff ignore list.


src/ansible_navigator/actions/run.py
```diff
@@ -425,7 +425,7 @@ def _init_replay(self) -> bool:
             return False
 
         version = data.get("version", "")
-        if version.startswith("1.") or version.startswith("2."):
+        if version.startswith("1.", "2."):
```
ssbarnea, line 428: Wow, i did not know about this one, i bet is also faster. Cool.

PR #1757, opened 2024-05-13
Address ruff `RET`

shatakshiiii:


ssbarnea (approved):

ssbarnea:
I seen a lot of code without coverage, but fixing it should be done in different PR.

PR #1760, opened 2024-05-14
Address ruff `TRY`

shatakshiiii:


src/ansible_navigator/cli.py
```diff
@@ -125,7 +125,7 @@ def run(args: ApplicationConfiguration) -> ActionReturn:
     if args.mode == "stdout":
         try:
             result = run_action_stdout(args.app.replace("-", "_"), args)
-            return result
+            return result  # noqa: TRY300
```
ssbarnea, line 128: Not ok to use noqa for this, read https://docs.astral.sh/ruff/rules/try-consider-else/ gives an example that should work just fine.

src/ansible_navigator/configuration_subsystem/configurator.py
```diff
@@ -182,7 +182,7 @@ def _apply_settings_file(self) -> None:
                     if self._params in (["settings", "--sample"], ["settings", "--gs"]):
                         return
                     msg = "Settings file cannot be empty."
-                    raise ValueError(msg)
+                    raise ValueError(msg)  # noqa: TRY301
```
ssbarnea, line 185: Are you sure that just using the simple raise from https://docs.astral.sh/ruff/rules/raise-within-try/ would break the code here?

PR #1781, opened 2024-05-30
Address ruff `PTH110`

shatakshiiii:


ssbarnea (commented):
PTH ones is not only about using the Path, but also about changing our internals to make use of path objects instead of strings. Current approach does have performance implications.

Still, I will not downvote it, we could change internals in follow-up if that requires significant extra changes. I just wanted to be sure that you know that aspect.

PR #1945, opened 2025-04-09
Avoid generation of output fixtures when not needed

ssbarnea:
This change should greatly increase the portability of test suite across platforms and different ansible versions.

AAP-43297
AAP-43336


cidrblock (commented):
Even if the fixture is not being used for a comparison, I think it helpful to "see" what's on the screen at that step of a given test.

For navigator the layout and rendering is as important as any string on the screen.

Was there a specifc issuee or error you encountered having the fixtures even when they were not being compared?

ssbarnea:
> Even if the fixture is not being used for a comparison, I think it helpful to "see" what's on the screen at that step of a given test.
> 
> For navigator the layout and rendering is as important as any string on the screen.
> 
> Was there a specifc issue or error you encountered having the fixtures even when they were not being compared?

Most of the tests are failing when using the fixtures due to different contexts, like different paths.

I was *never* able to run the full test suite on my MacOS and Ubuntu VM. I even seen occasional failures on CI caused by different rendering, see https://github.com/ansible/ansible-navigator/actions/runs/14357057299/job/40248901217#step:7:268 --- that failure in particular had nothing to do with the change, rerun did produce result that passed.

My NO.1 priority right now is to ensure that we CAN run the full test suite on any development machine that we use (assuming we have podman/docker installed).

The reality is that the final output on terminal might change for reasons outside our control. We should only perform behavior checking and not "screenshot" comparison. Imagine that we would do the same while testing vscode extension, it will be impossible to pass the testing because minor rendering changes are natural/expected to happen over time. Instead we should check that "output looks as expected", that we can spot the right information is present on screen, not that all the data on the screen is identical.

The second issue with fixtures is that I found impossible to regenerate them in a reliable way, as when I try to regenerate them I endup with changes in the "dumped" fixtures that are specific to the machine generating them.

I do understand your desired to be able to compare, but fixtures in their current form are not the right answer. Instead we should just dump the screenshots during test execution and include them into build archive, so we can download and compare them when needed.

PR #1950, opened 2025-04-14
Avoid generation of output fixtures when not needed

ssbarnea:
This change should greatly increase the portability of test suite
across platforms and different ansible versions.

Also increased container engine sanity timeout to 15s as previous
run of this job failed with timeout on this step.

AAP-43297
AAP-43336


shatakshiiii:
The output in fixtures has always been helpful while debugging any test failure, or just to see how the actual output should look like. Apart from reducing the fixture size, I am not sure how this change would be helpful.

Would like to get Brad's opinion on this.
cc: @cidrblock

PR #1955, opened 2025-04-14
Make navigator image pulling silent

ssbarnea:
Because podman in particular is overly verbose even when image is
already present, we switch to a default quiet mode.


ssbarnea:
We need to discuss about this but as someone that wants to use navigator from the cli to run multiple commands, I find the current default console output extremely verbose.

To describe it as user story: as someone that uses ansible-navigator in stdout mode, I expect to see only the output of executed command on screen on subsequent runs. There is no need to be told about container pulling each time I run, same output again and again.

I would not mind seeing something when a container IS PULLED (change occurred), but not when it was already updated.

```
❯ ansible-navigator --mode stdout run tests/fixtures/integration/actions/stdout/site.yml
> ansible-navigator --mode stdout run
-------------------------------------------------------------------------------------
Execution environment image and pull policy overview
-------------------------------------------------------------------------------------
Execution environment image name:     ghcr.io/ansible/community-ansible-dev-tools:latest
Execution environment image tag:      latest
Execution environment pull arguments: None
Execution environment pull policy:    tag
Execution environment pull needed:    True
-------------------------------------------------------------------------------------
Updating the execution environment
-------------------------------------------------------------------------------------
Running the command: podman pull ghcr.io/ansible/community-ansible-dev-tools:latest
Trying to pull ghcr.io/ansible/community-ansible-dev-tools:latest...
Getting image source signatures
Copying blob 8db0d92572bb skipped: already exists
Copying blob 1bbfeea407d0 skipped: already exists
Copying blob eb29a22534f3 skipped: already exists
Copying blob 98d7987d6554 skipped: already exists
Copying blob 8838ad2cbc16 skipped: already exists
Copying blob 4f4fb700ef54 skipped: already exists
Copying blob 2e3dd634db74 skipped: already exists
Copying blob dbe008bc4058 skipped: already exists
Copying blob c1256d4fd711 skipped: already exists
Copying blob 787afa220d3e skipped: already exists
Copying blob d7e8204cbcce skipped: already exists
Copying blob dbd88e24823d skipped: already exists
Copying blob 4f4fb700ef54 skipped: already exists
Copying blob b9954ca01d64 skipped: already exists
Copying blob 495cd171fb5d skipped: already exists
Copying blob adbad3bdbc8d skipped: already exists
Copying blob 4f4fb700ef54 skipped: already exists
Copying config d48945577c done   |
Writing manifest to image destination
d48945577ca57cd9579c43ed9f31e0278375693f22aaac46106129469169fcd1
<< ANSIBLE OUTPUT (the useful output >>
```

@tima Please take a look at this, maybe even try to run the commands yourself to get an impression so we can decide if we are to make a change and if we how we should do it.

In fact is more about it, the entire data seen on screen is on stdout but the reality is that the text that I find annoying is effectively "logging" information, not the output of the executed command.

I can quote from one of our favorite resources https://clig.dev/#output

> Send output to stdout. The primary output for your command should go to stdout. Anything that is machine readable should also go to stdout—this is where piping sends things by default.
> 
> Send messaging to stderr. Log messages, errors, and so on should all be sent to stderr. This means that when commands are piped together, these messages are displayed to the user and not fed into the next command.

To translate these for current case, I see the lines I want to remove as "messaging", and the ansible-playbook output as the "output" of the command.

cidrblock:
> We need to discuss about this but as someone that wants to use navigator from the cli to run multiple commands, I find the current default console output extremely verbose.

Because the default is to pull if tag ==latest (which is right), there may be a delay in bringing up the TUI. As long as the user is provided an explanation and the additional information is available via a cli param, I'm fine with some changes.

PR #1962, opened 2025-04-15
Enable uv support during testing

ssbarnea:


.config/constraints.txt
```diff
@@ -1,14 +1,9 @@
-#
-# This file is autogenerated by pip-compile with Python 3.10
-# by the following command:
-#
-#    pip-compile --all-extras --no-annotate --output-file=.config/constraints.txt --strip-extras pyproject.toml
-#
-# For testing with ansible-core 2.18, please restore the ansible-core lines manually
```
cidrblock, line 7: can thhis line be left in since it's a manual change?

cidrblock (approved):

.config/constraints.txt
ssbarnea: @cidrblock That guidance is no longer relevant/correct. The file is generated but we have the power to change the comment as that is defined inside the `pyproject.toml`. Still, I do not think that this is needed. Basically it will use 2.18 on all supported python version.

Even looking at the content from the repo, the comment does not make sense, primarily because ansible-core version is determined by the python version already.

PR #1979, opened 2025-05-05
Improve path normalization with a fallback for non-existent paths

shatakshiiii:
### SUMMARY

- Fixes the `FileNotFoundError` for paths that do not exist.

- Moves the cache path generation outside of the loop (unrelated to the above error, just found and corrected it)

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -647,7 +647,15 @@ class NavigatorPostProcessor:
         exit_messages: list[ExitMessage] = []
 
         if entry.value.current is not C.NOT_SET:
-            normalized = Path(entry.value.current).resolve().as_posix()
+            try:
+                normalized = Path(entry.value.current).resolve(strict=strict_resolve).as_posix()
+            except FileNotFoundError:
+                message = (
+                    f"Path `{entry.value.current}` does not exist."
+                    " Falling back to non-strict resolution."
+                )
+                messages.append(LogMessage(level=logging.WARNING, message=message))
+                normalized = Path(entry.value.current).resolve().as_posix()
             if normalized != entry.value.current:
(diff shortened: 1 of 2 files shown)


cidrblock (dismissed):

shatakshiiii:
Even though this PR solves one problem i.e. running ansible-navigator commands. It stills crashes while I debug the code locally. So moving it to draft as I need to check if there is a need of strict_resolve at line https://github.com/ansible/ansible-navigator/blob/main/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py#L650

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -647,7 +647,7 @@ def _normalize_path(
         exit_messages: list[ExitMessage] = []
 
         if entry.value.current is not C.NOT_SET:
-            normalized = Path(entry.value.current).resolve(strict=strict_resolve).as_posix()
```
ssbarnea, line 650: I see it problematic to resolve to path that does not exist. In most cases this is a configuration mistake and once that is likely to cause other problems.

Still, I can understand that in some cases, this might happen without being a show-stopper (maybe linux paths that do not exist on macos?). For this reason I suggest this, try using strict, catch the exception as a warning message, log it and fallback to non strict one. This should make it work without having to add other configuration options.

src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -647,7 +647,15 @@ def _normalize_path(
...
+            try:
+                normalized = Path(entry.value.current).resolve(strict=strict_resolve).as_posix()
+            except FileNotFoundError:
```
ssbarnea, line 652: To avoid the coverage error, for this particular case you can use `# pragma: no cover`

ssbarnea (approved):

PR #2057, opened 2025-10-07
chore: use reusable tox workflow

ssbarnea:


alisonlhart:
Moving to draft due to CI failures.

Issue #2055, opened 2025-10-07
Bug report: ansible-navigator on macOS misconfigures SSH agent socket and defaults to an unwritable SSH control path for non-root UIDs


netopsengineer:
##### ISSUE TYPE

- Bug Report

##### SUMMARY

On macOS with Docker Desktop, `ansible-navigator` auto-passes a launchd SSH agent socket path from the host that is unusable inside the Docker VM, so the EE cannot reach the agent. Separately, the EE defaults to an SSH control socket directory under `/runner/.ansible/cp` that is non-writable for the runtime user UID 501, which causes OpenSSH multiplexing to fail unless the container runs as root or the control dir is moved.

##### ANSIBLE-NAVIGATOR VERSION

```shell
ansible-navigator 25.9.0
```

##### CONFIGURATION

Relevant working configuration that demonstrates the needed workaround on macOS:

**ansible-navigator.yaml**

```yaml
ansible-navigator:
  execution-environment:
    enabled: true
    container-engine: docker
    image: ghcr.io/ansible/community-ansible-dev-tools:latest
    # Use container-options, not volume-mounts, so macOS host validation does not reject the path
    container-options:
      - "--volume=/run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock:ro"
      - "--env=SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock"
    # Optional, belt and suspenders for control path dir
    environment-variables:
      set:
        ANSIBLE_SSH_CONTROL_PATH_DIR: "/tmp/ansible/cp"
```

**ansible.cfg**

```ini
[ssh_connection]
control_path_dir = /tmp/ansible/cp
```

##### LOG FILE

Key excerpts that show the failure and the successful probe afterward.

**Agent failure with default Navigator auto-mount on macOS**

```shell
$ ansible-navigator exec -- /bin/sh -lc 'echo $SSH_AUTH_SOCK; ssh-add -l'
/private/tmp/com.apple.launchd.<ID>/Listeners
Error connecting to agent: Operation not supported
```

**Multiplexing failure when not running the EE as root**

```shell
muxserver_listen: link mux listener /runner/.ansible/cp/<long> => /runner/.ansible/cp/<short>: Bad file descriptor
```

**Successful auth via agent after workarounds**

```shell
debug1: Offering public key: ... ED25519 ... agent
debug1: Offering public key: ... RSA ... agent
debug1: Server accepts key: ... RSA ... agent
[...]

PR #2076, opened 2025-12-19
fix: mount docker's proxy ssh agent socket when running on macOS (#2055)

joonashak:
Quote from #2055:

> On macOS with Docker Desktop, ansible-navigator auto-passes a launchd SSH agent socket path from the host that is unusable inside the Docker VM, so the EE cannot reach the agent.

This PR adds the relevant `docker run` options to the runner to automatically configure SSH agent on macOS.

I originally came across this issue when trying to set up Ansible Navigator at work for a team with both Linux and macOS machines. Configuring the SSH agent automatically would be a huge QOL upgrade  🚀 

Reflecting on #2055 which also mentions an issue with SSH control path, I don't find it immediately related to this fix as the control path can be configured with `ANSIBLE_SSH_CONTROL_PATH_DIR` as it stands and doing so does not break things on Linux. However, that is one extra step that has to be accounted for when setting up for use on macOS. Do you think that a writeable SSH control path should also be set in this fix?

I didn't manage to get the tests working yet on my machine but I'll work on that once I get your input on whether or not this is overall something you'd be willing to merge.

### DIFF
--- a/src/ansible_navigator/runner/base.py
+++ b/src/ansible_navigator/runner/base.py
@@ -103,12 +103,11 @@ class Base:
         self.status: str | None = None
         self._runner_args: dict[str, Any] = {}
 
+        container_options = container_options or []
+
         # when the ce is podman, set the container user to root
         if self._ce == "podman":
-            if container_options:
-                container_options.append("--user=root")
-            else:
-                container_options = ["--user=root"]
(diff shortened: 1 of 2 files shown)


joonashak:
@shatakshiiii tests added and all checks passing 👍

PR #2114, opened 2026-04-08
fix(runner): resolve SSH_AUTH_SOCK symlink on macOS Tahoe before container mount

### LINKED ISSUE #2113: OSX Upgrade to Tahoe Breaks SSH_AUTH_SOCK Management
##### ISSUE TYPE

- Bug Report

##### SUMMARY

In OSX Tahoe >26.3, changes into how temporary files/directories are managed cause the SSH Agent socket to no longer be in location that is mountable by Ansible Navigator.

Previous versions of OSX mounted temporary files in the directory `/private/tmp`. However recent versions of OSX now place temporary files in `/var/run` which applies different permissions resulting in the following error when running most `ansible-navigator` commands:

```
Error: statfs /var/run/com.apple.launchd.xxxxxxxxx: no such file or directory
```

ansible-navigator 24.6.1

##### CONFIGURATION

No specific configurations

##### STEPS TO REPRODUCE

On a OSX Tahoe machine. Run a `ansible-navigator exec` command

```
ansible-navigator exec
```

##### EXPECTED RESULTS
  
Command completes successfully and user is able to perform desired `ansible-navigator` actions

##### ACTUAL RESULTS

Receive:

```
Error: statfs /var/run/com.apple.launchd.xxxxxxxxx: no such file or directory
```

stevefulme1:
## Summary

Fixes #2113

On macOS Tahoe (>26.3), the launchd SSH agent socket moved from `/private/tmp` to `/var/run`. Since `/var/run` is a symlink to `/private/var/run`, container runtimes (Podman, Docker) fail to mount the socket when given the unresolved symlink path, producing:

```
Error: statfs /var/run/com.apple.launchd.xxxxxxxxx: no such file or directory
```

### Root Cause

`ansible-runner`'s `_handle_automounts()` reads `SSH_AUTH_SOCK` from `os.environ` and passes it to `_update_volume_mount_paths()`, which normalises paths using `os.path.abspath()`. `os.path.abspath()` does **not** resolve symlinks, so the container engine receives the unresolvable `/var/run/...` path unchanged.

### Fix

In `Base.__init__()`, call `os.path.realpath()` on `SSH_AUTH_SOCK` before `ansible-runner` reads it from `os.environ`. This resolves `/var/run/com.apple.launchd.xxx/Listeners` → `/private/var/run/com.apple.launchd.xxx/Listeners`, which the container engine can mount.

The fix:
- Applies to all container engines on macOS (Docker already has its own hardcoded proxy socket override; this covers Podman and any future engines)
- Is a no-op on Linux (guard is `sys.platform == "darwin"`)
- Is a no-op on pre-Tahoe macOS where `SSH_AUTH_SOCK` already points to a real path (guard checks `realpath != original`)

## Test plan

- [x] `test_ssh_auth_sock_symlink_resolved_on_macos` — symlink path is rewritten to real path on darwin
- [x] `test_ssh_auth_sock_already_real_path_unchanged_on_macos` — non-symlink path is not modified
- [x] `test_ssh_auth_sock_not_modified_on_linux` — fix does not run on Linux
- [x] `test_ssh_auth_sock_missing_env_var_no_error` — no crash when `SSH_AUTH_SOCK` is unset
- [x] All 20 unit tests pass: `python3 -m pytest tests/unit/runner/test_base.py -v`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -4,6 +4,7 @@ argnames
 argparser
 argspec
 argvalues
+automounts
 autorefs
 basesystem
 caplog
@@ -54,6 +55,7 @@ netcommon           # Ansible network collection, seen in tests and README
 netconf
 nofile
 nonblocking
+normalises
 oldmask
 oneline
 onigurumacffi
@@ -70,6 +72,7 @@ rocannon's
 scap
 scrollback
 somevalue
+statfs
 stripspaces
 subaction
 subschema
--- a/src/ansible_navigator/runner/base.py
+++ b/src/ansible_navigator/runner/base.py
@@ -124,6 +124,25 @@ class Base:
             # Prepend to existing container options to allow overriding this fix.
             container_options = ssh_agent_socket_opts + container_options
 
+        # On macOS Tahoe (>26.3), the launchd SSH agent socket moved from
+        # /private/tmp to /var/run. Since /var/run is a symlink to /private/var/run,
+        # container runtimes require the resolved real path when building volume mount
+        # arguments. ansible-runner normalises mount paths with os.path.abspath(), which
+        # does not resolve symlinks, causing the container engine to fail with:
+        #   Error: statfs /var/run/com.apple.launchd.xxxxxxxxx: no such file or directory
+        # Resolving SSH_AUTH_SOCK to its real path here — before ansible-runner reads it
+        # from os.environ in _handle_automounts() — ensures the correct path is passed to
+        # the container engine regardless of which CE is in use.
+        if sys.platform == "darwin" and "SSH_AUTH_SOCK" in os.environ:
+            _real_sock = os.path.realpath(os.environ["SSH_AUTH_SOCK"])
+            if _real_sock != os.environ["SSH_AUTH_SOCK"]:
+                self._logger.debug(
+                    "macOS SSH_AUTH_SOCK symlink resolved: %s -> %s",
+                    os.environ["SSH_AUTH_SOCK"],
+                    _real_sock,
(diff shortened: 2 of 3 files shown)


stevefulme1:
The \`ack\` check is failing due to two related issues:

1. The \`release-drafter\` autolabeler step failed with \`Unsupported file extension: .ansible-community/devtools\` — this appears to be a transient issue with the shared \`ack\` workflow from \`ansible/team-devtools\`.
2. Because the autolabeler did not run successfully, no label was applied, causing the required-labels check to fail.

Could a maintainer please add the \`fix\` and \`minor\` labels to allow the \`ack\` check to pass? All other CI checks (lint, py310–py313, macOS, docs, codecov) are green.

sabre1041 (approved):
From a functional perspective, tested and validated on OSX 26.4 (Tahoe)

sathyapramod (changes requested):
cspell fix needed

The pre-commit hook is failing because three technical terms need to be added to .config/dictionary.txt:

  1. automounts - refers to the _handle_automounts() method in ansible-runner
  2. normalises - British spelling of "normalizes" (path normalization)
  3. statfs - Unix/Linux system call for filesystem statistics

  Add these words in alphabetical order:

   argvalues
  +automounts
   autorefs
   basesystem

   nonblocking
  +normalises
   oldmask
   oneline

   somevalue
  +statfs
   stripspaces
   subaction
   
   Happy to help you with it.

sabre1041 (approved):
LGTM

PR #2124, opened 2026-06-08
fix: optimize image introspection with section filtering

sathyapramod:
## Summary

The existing `--details` / `-d` flag on `ansible-navigator images` only controlled which sections were **displayed** — all 6 introspection collectors always ran inside the container regardless of the user's selection. This meant requesting a single lightweight section (e.g. `ansible_version`) still paid the full cost of collecting system packages, python packages, etc., resulting in ~5 minute execution times.

This fix threads the user's section selection into the container script so only the requested collectors execute, reducing filtered introspection from ~5 minutes to seconds.

Fixes AAP-43508

## Changes

- Add `ALL_COLLECTORS` dictionary in `image_introspect.py` mapping section IDs to their collector classes
- Add `--sections` CLI argument to `image_introspect.py` (internal interface between navigator and the container script — not user-facing)
- Update `main()` to only instantiate requested collectors when sections are provided; `"everything"` and `None` still run all collectors
- Pass `images_details` selection from `images.py` → `_run_runner()` → container cmdline so collection matches the user's intent
- `python_version` and `environment_variables` remain always-collected (hardcoded, not part of the collector mechanism)
- Update `codecov/codecov-action` to v6.0.2 (pinned to SHA) for GPG key migration

## Test plan

- [x] `tox -e lint` passes
- [x] `tox -e py` passes (CI)
- Unit tests added for `ALL_COLLECTORS` keys, section filtering, `main()` with specific/everything/None/invalid sections, and `_parse_args()` argument parsing

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -166,7 +166,7 @@ jobs:
           path: .
 
       - name: Upload coverage data
-        uses: codecov/codecov-action@v6.0.0
+        uses: codecov/codecov-action@8cad3ba95e5920c42f44492e54bc9639cba47959 # v6.0.2
         with:
           name: ${{ matrix.name }}
           fail_ci_if_error: true
--- a/src/ansible_navigator/actions/images.py
+++ b/src/ansible_navigator/actions/images.py
@@ -182,8 +182,12 @@ class Action(ActionBase):
             A message and return code
         """
         image_name = self._args.execution_environment_image
+        sections: list[str] = self._args.entry("images_details").value.current
 
-        output, error, return_code = self._run_runner(image_name=image_name)
+        output, error, return_code = self._run_runner(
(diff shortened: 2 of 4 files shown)


shatakshiiii (approved):

cidrblock (approved):
Nice work on this one — the fix is clean, CI is green, and the performance improvement is significant.

One suggestion for future PRs: a structured summary in the PR body (what the bug was, why it happened, what the fix does) really helps reviewers understand the change quickly. For example, this PR's original description listed *what* changed but didn't mention that `--details` was only filtering the display output while all collectors still ran — that's the key insight that makes the fix make sense. I've updated the PR body here as an example of the format we use (Summary, Changes, Test plan).

Not a blocker at all, just something that will make your future PRs even easier to review. Great contribution!

PR #2125, opened 2026-06-09
fix(command-runner): prevent indefinite deadlock in run_multi_process

sathyapramod:
## Description 
When ansible-navigator runs multiple commands in parallel, it launches worker processes and waits for all of them to report back. The main process expects exactly one result per command and calls `queue.get()` with no timeout. If any worker crashes (due to an unhandled exception in `post_process`, OOM, or signal), it dies without placing a result on the queue. The main process then blocks forever — no error message, no way to recover, and no indication of what went wrong.

## Changes
- **`command_runner.py` — `worker()`**: Wrapped `post_process()` in `try/except` so exceptions are captured on `command.errors` and the result is still placed on the queue
- **`command_runner.py` — `start_workers()`**: Now returns `list[multiprocessing.Process]` instead of joining internally
- **`command_runner.py` — `run_multi_process()`**: Uses `queue.get(timeout=30)` with a worker health check — raises `RuntimeError` with a descriptive message if all workers are dead and results are incomplete
- **`image_introspect.py` — `run_multi_thread()`**: Same timeout + thread health check pattern applied
- **`image_introspect.py` — `start_workers()`**: Now returns `list[threading.Thread]` instead of joining internally
- **`tests/unit/test_command_runner.py`** (new): 5 unit tests covering happy path, crashing `post_process`, single command, all callbacks crashing, and process reference return

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -166,7 +166,7 @@ jobs:
           path: .
 
       - name: Upload coverage data
-        uses: codecov/codecov-action@v6.0.0
+        uses: codecov/codecov-action@8cad3ba95e5920c42f44492e54bc9639cba47959 # v6.0.2
         with:
           name: ${{ matrix.name }}
           fail_ci_if_error: true
--- a/src/ansible_navigator/actions/images.py
+++ b/src/ansible_navigator/actions/images.py
@@ -182,8 +182,12 @@ class Action(ActionBase):
             A message and return code
         """
         image_name = self._args.execution_environment_image
+        sections: list[str] = self._args.entry("images_details").value.current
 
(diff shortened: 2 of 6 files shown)


.github/workflows/tox.yml
```diff
@@ -166,7 +166,7 @@ jobs:
           path: .
 
       - name: Upload coverage data
-        uses: codecov/codecov-action@v6.0.0
+        uses: codecov/codecov-action@8cad3ba95e5920c42f44492e54bc9639cba47959 # v6.0.2
```
sudhirverma, line 169: try to use codecov 7.0.0 version

sathyapramod: Thanks will update the PR

ssbarnea (approved):

PR #2147, opened 2026-07-20
fix: reduce cognitive complexity in ui_framework/ modules (S3776)

rockygeekz:
## Summary

- Extracted helper functions in `colorize.py` to reduce CC of `rgb_to_ansi`, `columns_and_colors`, `ansi_to_curses`, `strip_markdown`
- Extracted `_render_line_part` in `curses_window.py` to reduce CC of `_add_line`
- Extracted `_calculate_field_widths`, `_calculate_form_height`, `_generate_body_field_lines`, `_handle_form_input` in `form.py` to reduce CC of `_dimensions`, `_generate_form`, `present`
- Extracted `_handle_key_input` in `form_handler_options.py` to reduce CC of `handle`
- Extracted `_handle_arrow_down`, `_handle_arrow_up` in `form_handler_text.py` to reduce CC of `_do_command`
- Extracted `_build_text_field`, `_build_checks_field`, `_build_radio_field`, `_rekey_fields_by_name` in `form_utils.py` to reduce CC of `dict_to_form`, `form_to_dict`
- Extracted `_handle_display_key`, `_handle_obj_navigation`, `_handle_menu_cursor_keys` in `ui.py` to reduce CC of `_display`, `_show_obj_from_list`, `_show_menu`
- Removed stale `pylint: disable=too-many-locals` suppressions

## Test plan

- [ ] Verify all 15 S3776 SonarCloud issues are resolved
- [ ] Confirm no behavioral changes via existing test suite

## Summary by CodeRabbit

* **Refactor**
  * Improved terminal color/markup rendering by reorganizing ANSI conversion, region application, and markdown stripping into clearer reusable helpers.
  * Streamlined curses line rendering to draw segments more consistently within the screen bounds.
  * Refined form layout, body line generation, input handling, and option/text key behavior through extracted helper logic.
  * Centralized UI key handling for display, list navigation, and menu cursor/scroll movement.
* **Tests**
  * Added unit tests for RGB-to-ANSI conversion helpers.
  * Added unit tests for form utility helper behaviors and field/option rekeying.

### DIFF
--- a/src/ansible_navigator/ui_framework/colorize.py
+++ b/src/ansible_navigator/ui_framework/colorize.py
@@ -401,17 +401,21 @@ def columns_and_colors(
     return results
 
 
-def _parse_ansi_match(cap: dict[str, str | None]) -> tuple[int, int]:
+def _parse_ansi_match(
+    cap: dict[str, str | None],
+    color: int = 0,
+    style: int = 0,
+) -> tuple[int, int]:
     """Parse an ANSI color match groupdict into a color and style.
 
     Args:
         cap: The groupdict from a color_regex match
+        color: The current color state to preserve across calls
+        style: The current style state to preserve across calls
 
     Returns:
         A (color, style) tuple
     """
-    color = 0
-    style = 0
     one = cap["one"]
     two = cap["two"]
     if cap["fg_action"] == "39;" or (one == "0" and two is None):
@@ -469,7 +473,7 @@ def ansi_to_curses(line: str) -> CursesLine:
         if part:
             match = color_regex.match(part)
             if match:
-                color, style = _parse_ansi_match(match.groupdict())
+                color, style = _parse_ansi_match(match.groupdict(), color, style)
             else:
                 curses_line = CursesLinePart(
                     column=colno,
--- a/src/ansible_navigator/ui_framework/colorize.py
+++ b/src/ansible_navigator/ui_framework/colorize.py
@@ -244,6 +244,47 @@ def hex_to_rgb_curses(value: str) -> RgbTuple:
     return (scale_for_curses(red), scale_for_curses(green), scale_for_curses(blue))
 
 
+def _rgb_to_ansi_256(red: int, green: int, blue: int) -> int:
+    """Convert an RGB color to a 256-color ansi value.
+
+    Args:
+        red: The red component
+        green: The green component
+        blue: The blue component
+
+    Returns:
+        A 256-color ansi value
+    """
+    if red == green and green == blue:
+        if red < 8:
+            return 16
+        if red > 248:
+            return 231
+        return round(((red - 8) / 247) * 24) + 232
+    return 16 + (36 * round(red / 255 * 5)) + (6 * round(green / 255 * 5)) + round(blue / 255 * 5)
+
+
+def _rgb_to_ansi_16(red: int, green: int, blue: int) -> int:
+    """Convert an RGB color to a 16-color ansi value.
+
+    Args:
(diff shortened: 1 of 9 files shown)


rockygeekz:
codecov/patch is failing because this PR touches files under `src/ansible_navigator/ui_framework/`. These modules are curses/TUI code that can only be exercised through tmux integration tests, not unit tests. Any PR that modifies lines in these files will show low patch coverage since the unit test suite cannot import or drive curses based code.

codecov/project passes, confirming overall project coverage is unchanged. The changes in this PR are pure refactoring with no new logic, so the same integration tests that covered the original code still cover it at runtime.

sathyapramod (dismissed):
LGTM! - We can Ignore CodeCov and ready to merge!

src/ansible_navigator/ui_framework/colorize.py
```diff
@@ -308,38 +390,46 @@ def columns_and_colors(
...
-        column = 0
-        for line_part in result:
-            line_part.column = column
-            column += len(line_part.chars)
+    _update_line_part_columns(results)
 
     return results
 
 
+def _parse_ansi_match(cap: dict[str, str | None]) -> tuple[int, int]:
```
Hrithik-Gavankar: ### ANSI Style Persistence Regression

During the refactoring of `ansi_to_curses`, `_parse_ansi_match()` resets `color` and `style` to `0` on every call. This changes the original behavior where style/color state was preserved across consecutive ANSI escape sequences.

**Impact**

* Consecutive ANSI codes (e.g. `\x1b[1;31m\x1b[38;5;46m`) lose the previously applied style (e.g. **bold**) when only the foreground color changes.

**Suggested Fix**
Pass the current parser state into `_parse_ansi_match()` instead of reinitializing it:

```python
color, style = _parse_ansi_match(match.groupdict(), color, style)
```

and initialize from the passed values inside the function instead of `0`.

**Note:** The handling of the reset sequence (`\x1b[0m`) is correct and should remain unchanged, as it fixes an issue in the previous implementation.

Hrithik-Gavankar (changes requested):

src/ansible_navigator/ui_framework/ui.py
```diff
@@ -812,46 +905,20 @@ def _show_obj_from_list(
...
-                continue
-
-            if entry == "+":
-                more = [i for i in self._menu_indices if i - index > 0]
-                less = [i for i in self._menu_indices if i - index < 0]
-
-                ordered_indices = more + less
-                if ordered_indices:
-                    index = ordered_indices[0]
+            should_continue, new_index, new_entry = self._handle_obj_navigation(
```
Hrithik-Gavankar, line 908: ### Behavioral Regression: Scroll Reset on Navigation

The refactoring introduces a subtle behavioral change in `_handle_obj_navigation()`.

Previously, pressing `-`, `+`, or a numeric key always reset the scroll position (`self.scroll(0)`), even if the navigation resolved to the currently selected item. After the extraction, `self.scroll(0)` is only called when `new_index != index`, so navigating to the same item no longer scrolls back to the top.

This changes the existing behavior despite the refactor intending to be functionally equivalent.

**Suggested fix:** Decouple the scroll reset from the index update. Preserve the original behavior by always calling `self.scroll(0)` for `-`, `+`, and numeric navigation, regardless of whether the selected index changes.

Hrithik-Gavankar (changes requested):

Hrithik-Gavankar (approved):
Changes LGTM!

PR #2148, opened 2026-07-20
fix: reduce cognitive complexity in config, utils, and tests (S3776)

rockygeekz:
## Summary
- Refactored 28 functions across config/, utils/, data/, and tests/ directories to reduce cognitive complexity below SonarCloud's S3776 threshold of 15
- Extracted nested logic into focused helper functions without changing any behavior or method signatures

### Files refactored:
- `configurator.py`: `_apply_settings_file`, `_check_choices`, `_apply_previous_cli_to_current`
- `navigator_post_processor.py`: `execution_environment`, `execution_environment_volume_mounts`, `inventory`, `mode`
- `transform.py`: `to_effective`, `to_schema`
- `catalog_collections.py`: `_catalog_roles`, `_one_path`
- `image_introspect.py`: `splitter`
- `dict_merge.py`: `in_place_list_replace`
- `dot_paths.py`: `place_at_path`
- `print.py`: `color_lines`
- `migrate.py`: `run_migrations`
- `rules.py`: `Rule.make`, `_captures`
- `conftest.py`: `_cmd_in_tty`, `pytest_configure`
- `defaults.py`: `id_func`
- `_tmux_session.py`: `interaction`
- 6 integration test base files: `test` methods in config, doc, run, run_unicode, settings, templar

## Test plan
- [x] `ruff format` passes
- [x] `ruff check` passes
- [x] `tox -e lint` passes

## Summary by CodeRabbit

* **Refactor**
  * Streamlined configuration and settings-file processing, including validation, choice handling, and reuse of eligible prior CLI values.
  * Improved collection metadata and role cataloging, plus schema generation and volume-mount transformation behavior.
  * Refined mode post-processing and configuration/path placement logic.
  * Improved terminal color rendering and more consistent grammar capture handling.
  * Clarified version-migration prompts, execution flow, and completion messaging.
* **Chores**
  * Updated documentation-linting baseline and dictionary terminology.


rockygeekz:
codecov/patch is failing because this PR touches files under `src/ansible_navigator/ui_framework/`. These modules are curses/TUI code that can only be exercised through tmux integration tests, not unit tests. Any PR that modifies lines in these files will show low patch coverage since the unit test suite cannot import or drive curses based code.

codecov/project passes, confirming overall project coverage is unchanged. The changes in this PR are pure refactoring with no new logic, so the same integration tests that covered the original code still cover it at runtime.

sathyapramod (dismissed):
LGTM! We can ignore CodeCov and Reday to merge.

tests/integration/_tmux_session.py
```diff
@@ -306,136 +306,249 @@ def _capture_pane(self) -> list[str]:
...
+
+    def _wait_for_response(
+        self,
+        search_within_response: list[str] | str | None,
+        ignore_within_response: str | None,
+        start_time: float,
+        timeout: int,
+    ) -> tuple[list[str], bool]:
+        """Wait for the expected response to appear on screen.
+
```
Hrithik-Gavankar, line 474: I think this introduces a behavior change compared to the original implementation.

When `_wait_for_stable_screen()` times out, the new code calls `_handle_response_timeout()`, which sets `self._fail_remaining`. That effectively marks the current timeout as a test failure and causes all subsequent tests to fail as well.

In the original implementation, a stable-screen timeout **did not** trigger the timeout handler. Although `err_message = "5 LIKE SCREENS"` was assigned, the outer `break` exited before reaching the error-handling path, so execution simply continued with the best available screen.

Current code:

```python
final_showing, stable_timeout = self._wait_for_stable_screen(
    showing, start_time, timeout,
)
if stable_timeout:
    stable_elapsed = timer() - start_time
    return self._handle_response_timeout(
        final_showing, "5 LIKE SCREENS", stable_elapsed,
    ), True
```

Original behavior (`main`):

```python
if elapsed > timeout:
    err_message = "5 LIKE SCREENS"
    break  # exits inner loop

...
break  # exits outer loop before timeout handling
```

To preserve the existing behavior, I think we should avoid invoking `_handle_response_timeout()` here and instead return the last stable screen:

```python
if stable_timeout:
    return final_showing, False
```

This keeps the behavior consistent with `main` and avoids introducing cascading test failures in CI due to `self._fail_remaining` being set.

Hrithik-Gavankar (changes requested):

Hrithik-Gavankar (approved):
Changes LGTM!

PR #2158, opened 2026-08-11
fix(docs): remove unused mkdocstrings plugin causing 429 failures

rockygeekz:
## Summary

- Remove the unused `mkdocstrings` plugin from `mkdocs.yml` and its dictionary entry

The mkdocstrings plugin is configured to fetch `objects.inv` from docs.ansible.com, which intermittently returns HTTP 429 (Too Many Requests). Since mkdocs runs in strict mode, this causes the docs CI job to fail.

The plugin is not actually used anywhere in the docs (no `:::` autodoc directives exist), so removing it is safe and has no effect on the generated documentation.

## Test plan

- [ ] Verify docs CI job passes without the mkdocstrings plugin
- [ ] Confirm no `:::` autodoc directives exist in the docs that would break

## Summary by CodeRabbit

* **Documentation**
  * Removed the `mkdocstrings` documentation plugin and related Python and Ansible reference configuration.
  * Removed the associated documentation dictionary entry.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -49,7 +49,6 @@ magiclink
 maskables
 maxsplit
 mergeable
-mkdocstrings
 mqueue              # https://github.com/ansible/ansible-runner/issues/984
 myuser
 myproject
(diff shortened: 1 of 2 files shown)


shvenkat-rh (approved):

rockygeekz:
This PR only touches mkdocs config, not actual docs content, so the ReadTheDocs check won't report. Bypassing and merging.

PR #2160, opened 2026-08-23
fix: handle invalid artifact filename formats

shvenkat-rh:
Handle malformed braces in playbook_artifact_save_as as a configuration error instead of allowing Formatter.parse() to raise ValueError and crash the application. The error now identifies the invalid filename and provides a hint listing the supported variables.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for malformed artifact filename format strings.
  * Displays a clear error and supported-variable hint when unmatched braces are detected.
  * Prevents further validation until the invalid format is corrected.

* **Tests**
  * Added coverage for malformed format strings and supported-variable hints.

### DIFF
--- a/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
+++ b/src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
@@ -1205,12 +1205,26 @@ class NavigatorPostProcessor:
         """
         messages: list[LogMessage] = []
         exit_messages: list[ExitMessage] = []
-        # literal_text, fname, format_spec, conversion
-        found = {f for _, f, _, _ in Formatter().parse(entry.value.current) if f}
         available = {f for _, f, _, _ in Formatter().parse(entry.value.default) if f}
         non_defaults_also_available = {"playbook_status"}
 
         available.update(non_defaults_also_available)
+        # literal_text, fname, format_spec, conversion
+        try:
+            found = {f for _, f, _, _ in Formatter().parse(entry.value.current) if f}
+        except ValueError:
+            # Handle invalid artifact filename formats (malformed braces)
+            exit_msg = (
+                f"The playbook artifact file name '{entry.value.current}', set by"
+                f" {entry.value.source.value.lower()}, is not a valid format string."
+            )
+            exit_messages.append(ExitMessage(message=exit_msg))
+            exit_msg = (
+                "Invalid format string. Ensure braces are properly matched and only"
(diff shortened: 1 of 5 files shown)


src/ansible_navigator/configuration_subsystem/navigator_post_processor.py
```diff
@@ -1206,7 +1206,25 @@ def playbook_artifact_save_as(
...
+            found = {f for _, f, _, _ in Formatter().parse(entry.value.current) if f}
+        except ValueError:
+            # Handle invalid artifact filename formats (malformed braces)
+            exit_msg = (
+                f"The playbook artifact file name '{entry.value.current}', set by"
+                f" {entry.value.source.value.lower()}, is not a valid format string."
+            )
+            exit_messages.append(ExitMessage(message=exit_msg))
+            supported = oxfordcomma(
+                {"playbook_name", "playbook_status", "inventory", "playbook_artifact_save_as"},
```
rockygeekz: This hardcoded set is wrong. The actual placeholders are `playbook_dir`, `playbook_name`, `playbook_status`, and `time_stamp` (from the default format string, `write_artifact()` in `run.py`, and the setting description). This list is missing `playbook_dir` and `time_stamp`, and includes `inventory` and `playbook_artifact_save_as`, which are setting names, not format variables.

Compute `available` from `entry.value.default` plus `playbook_status` (same as the unknown-variables path below) and reuse:

```python
exit_msg = f"Try again with only {oxfordcomma(available, \"and/or\")}"
```

tests/unit/configuration_subsystem/post_processors/test_playbook_artifact_save_as.py
```diff
@@ -66,6 +66,15 @@ def __str__(self) -> str:
             " has unrecognized variables: 'name'"
         ),
     ),
+    Scenario(
+        name="4",
+        current="{playbook_name}.json{",
+        exit_message_substr=(
```
rockygeekz, line 72: `exit_message_substr` only covers the primary error (`exit_messages[0]`). The malformed-brace path also emits a HINT as `exit_messages[1]` with the supported-variables list. Add a field here (or a separate assertion) to verify the hint contains `playbook_dir`/`time_stamp` and does not contain `inventory`.

rockygeekz (approved):
Please resolve the above comments

rockygeekz (changes requested):
Please resolve the above comments

rockygeekz (approved):

PR #2164, opened 2026-08-25
fix: handle OSError when checking a playbook path

shvenkat-rh:
Handle OSError (e.g. ENAMETOOLONG) that can be raised when checking if a playbook path exists. When the path cannot name a file on this system, immediately return "missing" instead of attempting the FQCN fallback logic, which could incorrectly classify a too-long path as an FQCN playbook if it contains dots.

Changes:
- Wrap Path.exists() in try-except to catch OSError
- Return "missing" immediately on OSError to prevent FQCN fallback
- Restructure the FQCN logic to only apply when the file genuinely doesn't exist (not when the existence check fails)
- Add test case to verify OSError handling with too-long path names

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved playbook detection when a path cannot be checked by the system.
  * Such paths are now correctly reported as missing instead of causing an error.
  * Added safer handling for paths that exceed system filename or path-length limits.

* **Tests**
  * Added coverage confirming expected results when path checks fail due to system limits.
  * Expanded validation of missing-path detection in these edge cases.

### DIFF
--- a/.config/dictionary.txt
+++ b/.config/dictionary.txt
@@ -19,6 +19,7 @@ colval
 devel
 dumpable
 dunder
+enametoolong
 escdelay
 facelessuser
 fedoraproject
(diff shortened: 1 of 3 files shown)


sathyapramod:
@coderabbitai review

sathyapramod (approved):
No critical or major findings. LGTM - Ready to merge

PR #2172, opened 2026-09-14
fix: request explicit columns when listing images

### LINKED ISSUE #2098: KeyError Tag: docker 29 incompatibility
##### ISSUE TYPE

- Bug Report

##### SUMMARY

ansible-navigator images fails with KeyError: 'tag' when using Docker 29.x due to JSON output schema changes in docker images --format json.

Docker 29 returns capitalized JSON keys (e.g. "Tag"), while ansible-navigator 26.1.2 expects lowercase keys (e.g. "tag"), causing a crash during image inspection.

```
"ansible-navigator 26.1.2"
```

##### CONFIGURATION

No custom ansible-navigator.yaml present.

Container engine: docker (default)

##### LOG FILE
```
CRITICAL 'ansible_navigator.actions._actions.run_interactive' Subcommand 'images' encountered a fatal error.
ERROR 'ansible_navigator.actions._actions.run_interactive' Logging an uncaught exception
  File ".../ansible_navigator/actions/_actions.py", line 202, in run_interactive
    return run_action(app=app, interaction=interaction)
  File ".../ansible_navigator/actions/images.py", line 240, in run
    self._collect_image_list()
  File ".../ansible_navigator/actions/images.py", line 457, in _collect_image_list
    images, error = inspect_all(container_engine=self._args.container_engine)
  File ".../ansible_navigator/image_manager/inspector.py", line 115, in inspect_all
    result = cmd_runner.run_single_process(commands=images_list_class.commands)
  File ".../ansible_navigator/command_runner/command_runner.py", line 124, in run_single_process
    command.post_process(command)
  File ".../ansible_navigator/image_manager/inspector.py", line 97, in parse
    valid_images = [image for image in local_images if image["tag"] != "<none>"]
  File ".../ansible_navigator/image_manager/inspector.py", line 97, in <listcomp>
    valid_images = [image for image in local_images if image["tag"] != "<none>"]
KeyError: 'tag'
```

##### STEPS TO REPRODUCE

1. Install Docker 29.x
2. Install ansible-navigator 26.1.2 via pip or pipx.
3. Run `ansible-navigator images`
##### EXPECTED RESULTS

ansible-navigator images should list local container images without crashing, regardless of Docker JSON field capitalization changes.

##### ACTUAL RESULTS

The command crashes with:

KeyError: 'tag'

##### ADDITIONAL INFORMATION

Docker version 29.0.4, build 3247a5a
```
docker images --format json | head -n 1
{"Containers":"1","CreatedAt":"2026-02-16 08:07:02 +0100 CET","CreatedSince":"31 hours ago","Digest":"<none>","ID":"1ef93f0d9e75","Repository":"ddev/ddev-webserver","SharedSize":"N/A","Size":"1.27GB","Tag":"v1.24.10-efficare-built","UniqueSize":"N/A"}

```
Suggested fix: normalize keys (case-insensitive lookup) or explicitly handle both "tag" and "Tag" when parsing Docker image output.

### LINKED ISSUE #2126: docker 29 incompatibility. images command fails with docker 29
##### ISSUE TYPE

- Bug Report

##### SUMMARY

images command doesn't work with docker 29.x

```
ansible-navigator 26.4.0
```

##### CONFIGURATION

default

##### LOG FILE
```
2026-06-12T14:07:49.037580+00:00 CRITICAL 'ansible_navigator.actions._actions.run_interactive' Subcommand 'images' encountered a fatal error.
2026-06-12T14:07:49.037943+00:00 ERROR 'ansible_navigator.actions._actions.run_interactive' Logging an uncaught exception
  File "/home/scicore-adm/.venv_ansible_ee/lib/python3.10/site-packages/ansible_navigator/actions/_actions.py", line 202, in run_interactive
    return run_action(app=app, interaction=interaction)
  File "/home/scicore-adm/.venv_ansible_ee/lib/python3.10/site-packages/ansible_navigator/actions/images.py", line 240, in run
    self._collect_image_list()
  File "/home/scicore-adm/.venv_ansible_ee/lib/python3.10/site-packages/ansible_navigator/actions/images.py", line 457, in _collect_image_list
    images, error = inspect_all(container_engine=self._args.container_engine)
  File "/home/scicore-adm/.venv_ansible_ee/lib/python3.10/site-packages/ansible_navigator/image_manager/inspector.py", line 123, in inspect_all
    images = {image["image_id"]: image for image in images_list.details}
  File "/home/scicore-adm/.venv_ansible_ee/lib/python3.10/site-packages/ansible_navigator/image_manager/inspector.py", line 123, in <dictcomp>
    images = {image["image_id"]: image for image in images_list.details}
KeyError: 'image_id'
```

##### STEPS TO REPRODUCE

Execute command `images` with docker 29.x

##### ADDITIONAL INFORMATION

The problem seem to be that the output of `docker images` has changed from docker 28.x to 29.x 

This is the output for docker 29.x
```
docker images | head -1
IMAGE                                                ID             DISK USAGE   CONTENT SIZE   EXTRA
```

and this is the output for docker 28.x
```
 docker images | head -1
REPOSITORY                                    TAG       IMAGE ID       CREATED          SIZE
```

as a workaround you can downgrade package `docker-ce-cli` to 28.x

anandghegde:
## Summary

`ansible-navigator images` fails with `KeyError: 'image_id'` on Docker 29. This follows up on the discussion in #2098.

## Cause

`ImagesList.parse` builds the dict keys from the header row of the default `docker images` / `podman images` table. Docker 29 changed that default table to `IMAGE  ID  DISK USAGE  CONTENT SIZE  EXTRA`, so the parsed rows have no `repository`, `tag` or `image_id` keys. #2096 made the `tag` lookup safe, which moved the crash to `inspect_all` (`image["image_id"]`). The same problem happens on any Docker version when the user sets `imagesFormat` in `~/.docker/config.json`.

## Fix

For docker and podman, ask for the fields explicitly:

```
images --format '{{.Repository}}\t{{.Tag}}\t{{.ID}}\t{{.CreatedSince}}\t{{.Size}}'
```

The parser then maps each tab separated line to `repository`, `tag`, `image_id`, `created` and `size` by position, so it no longer depends on the header text. These are the same keys and values the old table produced. Both engines support these template fields and the `\t` escape. Apple Container (`container image list`) is not changed.

## Testing

- Updated the unit tests in `tests/unit/image_manager/test_image_inspector.py` for the new command and the tab separated parsing (docker and podman, including a size with a space and a `<none>` tag that gets dropped). The updated tests fail without the change and pass with it.
- Docker 29.5.2 engine with a 29.x CLI: before the change, `ansible-navigator images --mode stdout` failed with `KeyError: 'image_id'`. After the change it exits 0 and lists the execution environment image.
- With a temporary `DOCKER_CONFIG` that sets `imagesFormat` to `table {{.ID}}\t{{.Size}}`, `inspect_all("docker")` still returns every tagged image with all five keys and no inspect errors.
- Docker 28.5.2 CLI (`docker:28-cli` container against the same daemon): the template output has the same shape.
- Podman 5.8.4 (inside `quay.io/podman/stable`): `inspect_all("podman")` from this branch returns the expected `repository` / `tag` / `image_id` / `created` / `size` values and inspect details.
- `tox -e lint` passes.

Fixes #2126
Fixes #2098

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image listing consistency across Docker and Podman.
  * Blank lines in image-list output are now ignored.
  * Empty image-list results are handled cleanly without errors.
  * Malformed image-list entries are detected instead of being parsed incorrectly.
  * Image-list fields are now interpreted consistently across supported environments.


shvenkat-rh:
@anandghegde , please ensure commits must have verified signatures.

anandghegde:
@shvenkat-rh Done. I re-signed the commit (f2af2a8) and GitHub now shows it as verified. The content is unchanged.

src/ansible_navigator/image_manager/inspector.py
```diff
@@ -98,13 +108,17 @@ def parse(cls, command: Command) -> None:
...
+                headers = [key.lower().replace(" ", "_") for key in re_2omo.split(images.pop(0))]
                 headers = [cls.APPLE_HEADER_MAP.get(h, h) for h in headers]
-            local_images = [
-                dict(zip(headers, re_2omo.split(line), strict=False)) for line in images
-            ]
+                local_images = [
+                    dict(zip(headers, re_2omo.split(line), strict=False)) for line in images
+                ]
+            else:
+                local_images = [
```
rockygeekz, line 119: Can we skip empty lines before building rows? A trailing blank line from engine output becomes a row with no image_id and inspect_all still hits KeyError on image_id. Something like filtering with line.strip() on the docker/podman branch would avoid that.

src/ansible_navigator/image_manager/inspector.py
```diff
@@ -98,13 +108,17 @@ def parse(cls, command: Command) -> None:
...
+                    dict(zip(cls.FORMAT_KEYS, line.split("\t"), strict=False)) for line in images
```
rockygeekz, line 120: For the explicit --format path, strict=True on zip(FORMAT_KEYS, line.split("\t")) would fail parse on a malformed line instead of a partial dict and a later KeyError in inspect_all. Happy path on Docker 29.8.0 looks good from my side; this is mostly defensive.

tests/unit/image_manager/test_image_inspector.py
```diff
@@ -34,8 +34,18 @@ def test_images_inspect_command(
 @pytest.mark.parametrize(
     ("container_engine", "expected_list_cmd"),
     (
-        pytest.param("podman", "podman images", id="podman-images"),
-        pytest.param("docker", "docker images", id="docker-images"),
+        pytest.param(
```
rockygeekz, line 37: nit: expected_list_cmd duplicates ImagesList.FORMAT. Building the string from ImagesList(container_engine).commands[0].command (or FORMAT) would keep the test tied to one source if the template changes.

rockygeekz (changes requested):

src/ansible_navigator/image_manager/inspector.py
anandghegde: Fixed in 930d72ca. `parse` now drops blank lines before building rows:

```python
images = [line for line in command.stdout.splitlines() if line.strip()]
```

I put the filter before the branch rather than inside the docker/podman one, so the Apple Container path gets it too and its `images.pop(0)` still finds the header. Added `test_images_list_parse_skips_blank_lines`, which feeds a trailing newline and asserts a single row.

src/ansible_navigator/image_manager/inspector.py
anandghegde: Done in 930d72ca, the docker/podman zip is now `strict=True`. Added `test_images_list_parse_rejects_malformed_line`.

One thing worth flagging: `inspect_all` uses `CommandRunner.run_single_process`, which calls `post_process` without the `try/except` that the multi-process worker has, so this surfaces as a `ValueError` out of `inspect_all` rather than landing in `command.errors`. With the blank-line filter above I don't expect it to trigger in practice: every field of the requested template is tab separated even when a value is empty, so a short line means the engine ignored `--format`.

I left the Apple Container zip non-strict, since that path maps headers it reads from the output rather than a template we control.

Happy to skip malformed lines instead of raising if you'd rather `images` degrade to the rows it can read.

tests/unit/image_manager/test_image_inspector.py
anandghegde: Done in 930d72ca. Both params now build the expected string from `ImagesList.FORMAT`:

```python
pytest.param("docker", f"docker images --format '{ImagesList.FORMAT}'", id="docker-images"),
```

I kept the `--format` wrapper spelled out in the test so it still pins the command shape, and only the template itself comes from the one source.

src/ansible_navigator/image_manager/inspector.py
```diff
@@ -97,14 +107,18 @@ def parse(cls, command: Command) -> None:
         """
         if command.stdout:
             is_apple = command.command.startswith("container ")
-            images = command.stdout.splitlines()
-            re_2omo = re.compile(r"\s{2,}")
-            headers = [key.lower().replace(" ", "_") for key in re_2omo.split(images.pop(0))]
+            images = [line for line in command.stdout.splitlines() if line.strip()]
```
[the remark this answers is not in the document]
anandghegde: Good catch, this was real. `container list` on a host with no images can come back as blank lines only, and the blank-line filter I added in 930d72ca then left `images` empty, so `images.pop(0)` raised `IndexError` instead of reporting no images. `command.stdout` is truthy for `"\n"`, so the outer guard did not cover it.

Fixed in 34513cf0:

```python
images = [line for line in command.stdout.splitlines() if line.strip()]
if not images:
    command.details = []
    return
```

I put the guard before the engine branch rather than only in the Apple one, so every engine reports no images the same way, and `[]` matches the `Command.details` default.

Added `test_images_list_parse_blank_output`, parametrized over `docker` and `container image list`, which feeds `"\n\n"` and asserts `details == []`. With the guard reverted the `container` case fails with `IndexError`. Unit suite is 13/13 on `tests/unit/image_manager/`.

rockygeekz (approved):
Thanks @anandghegde . LGTM!

